PR #334, opened 2014-12-30
add collisionreport debug message

ompugao:


rdiankov:
there's a much easier way to do that

```
if( _pmanip->CheckEndEffectorCollision(t*_ttogripper, report) ) {
    RAVELOG_VERBOSE("endeffector is in collision, %s\n",report->__str__());
}
```

Also, I would create the report once in the class and re-use it. Calling "new" everytime kills performance.

ompugao:
i fixed it

rdiankov:
awesome! thanks

PR #341, opened 2015-01-27
pnodeparent is sometimes null

ompugao:


rdiankov:
oops, i had already made the change in production ;0)

Issue #167, opened 2012-09-15
add 'make altinstall' for installing with suffixes


rdiankov:
possible solution from Michael Hertling

```
Look at the following CMakeLists.txt:

CMAKE_MINIMUM_REQUIRED(VERSION 2.8.4 FATAL_ERROR)
PROJECT(ALTINSTALL C)
SET(CMAKE_VERBOSE_MAKEFILE ON)
FILE(WRITE ${CMAKE_BINARY_DIR}/f.c "void f(void){}\n")
ADD_LIBRARY(f SHARED f.c)
SET_TARGET_PROPERTIES(f PROPERTIES SOVERSION 1.0)
INSTALL(TARGETS f RUNTIME DESTINATION lib LIBRARY DESTINATION lib)
INSTALL(TARGETS f RUNTIME DESTINATION lib LIBRARY DESTINATION lib
   NAMELINK_SKIP COMPONENT altinstall)
FILE(WRITE ${CMAKE_BINARY_DIR}/main.c "int main(void){return 0;}\n")
ADD_EXECUTABLE(main main.c)
SET_TARGET_PROPERTIES(main PROPERTIES OUTPUT_NAME "main1.0")
ADD_CUSTOM_COMMAND(TARGET main POST_BUILD
   COMMAND ${CMAKE_COMMAND} -E create_symlink
   $<TARGET_FILE_NAME:main> ${CMAKE_BINARY_DIR}/main)
INSTALL(TARGETS main RUNTIME DESTINATION bin)
INSTALL(TARGETS main RUNTIME DESTINATION bin COMPONENT altinstall)
INSTALL(FILES ${CMAKE_BINARY_DIR}/main DESTINATION bin)
ADD_CUSTOM_TARGET(altinstall
   ${CMAKE_COMMAND}
   -DCOMPONENT=altinstall
   -P ${CMAKE_BINARY_DIR}/cmake_install.cmake)

The basic idea is to define an installation component "altinstall"
along with an equally named custom target without ALL clause which
installs the altinstall component in the known way, i.e. by running
cmake_install.cmake with COMPONENT. For the library target, add two
INSTALL() commands, one as usual and the other with a NAMELINK_SKIP
and a COMPONENT clause. For the executable targets, I'd recommend to
create the version symlink in the build tree via a POST_BUILD custom
command. Use two INSTALL() commands here, too, and an additional one
without COMPONENT to install the executables' symlinks when you are
not performing the altinstall.
```

PR #342, opened 2015-02-13
Install symlinks using relative path to make openrave installation relocatable.

ziyan:
Tested compiling and installing on debian wheezy.  Could be related to issue #167.


rdiankov:
FYI, the conversation was: due to openrave libraries not being relocatable, it is not worth making the symlinks relative

PR #345, opened 2015-05-04
Added releasegil option to PlanningUtils retimers

mkoval:
This patch adds a boolean argument to release the GIL in the PlanningUtils RetimeTrajectory and SmoothTrajectory calls. This includes #5.


rdiankov:
cool! great commit!

PR #351, opened 2015-06-16
fix for sympy0.7.9. seems zeros argument has changed

emijah:
Not quite sure if I'm breaking PR etiquette, but here goes.
It seems that sympy 0.7.9 has a different interface to sympy 0.7.1 which is bundled with openrave. I split the fix into another file so that the bundled version takes precedence.


emijah:
Retracting this because OpenRAVE requires sympy version 0.7.1

PR #352, opened 2015-07-01
use pkg-config to find fparser if cmake fails otherwise

morxa:
fparser does not always come with a cmake module. As an example, both the Debian and the Fedora package of fparser do not contain a cmake module, but instead provide pkgconfig files. Fall back to pkg-config if fparser was not found by cmake.


rdiankov:
thanks.
you still need the following line though:

```
set(FPARSER_CXX_FLAGS "${FPARSER_CXX_FLAGS} -DOPENRAVE_FPARSER_SETEPSILON") 
```

morxa:
Thanks for spotting the error! I've just pushed the fix.

rdiankov:
cool, thanks

PR #353, opened 2015-07-01
cmake: make all installed python scripts executable

morxa:


rdiankov:
hmm...... ikfast_generator_cpp.py ikfast_generator_cpp_sympy0_6.py are not "executable". they can only be used by importing them from python. making them executable will give wrong impression to users.

morxa:
Thanks for clearing this up and sorry for the wrong pull request!

morxa:
Follow-up question: Why is there a shebang in those files if they are not meant to be executed?

PR #357, opened 2015-07-10
added missing else in openravepy_collisionchecker.cpp and package.xml for catkin workspace

dseredyn:
There was a bug in openravepy_collisionchecker.cpp that caused CheckCollision(link1, link2, report) to crash (exception thrown).

I've also added package.xml so the OpenRAVE can be compiled as a catkin package (isolated).


rdiankov:
cool! great spot!
i cannot maintain package.xml at all, so if catkin changes, can you promise to update it?

PR #359, opened 2015-08-10
Add localization to libopenrave, libopenrave-core, and openrave pytho…

ziyan:
…ng binding.

TODO:

- ikfast is not localized
- in python/bindings, some files are not localized because they do not import any openrave code, e.g. pyANN_int, convexdecompositionpy, bindings.cpp


rdiankov:
- remove TABs from CMakeLists.txt
- rename RaveGetText to RaveGetLocalizedTextForDomain
- updatetranslations.bash should use the script path rather than looking for .git

ziyan:
Fixed:

- remove TABs from CMakeLists.txt
- rename RaveGetText to RaveGetLocalizedTextForDomain
- updatetranslations.bash should use the script path rather than looking for .git

Also:

- moved updatetranslations.bash into locale folder
- implemented multi domain support, each localization domain will only need to list its files once in locale/doamins folder
- implemented localization for some plugins: configurationcache, ikfastsolvers, oderave, rplanners

rdiankov:
good job!

PR #360, opened 2015-08-10
Couple of fixes regarding localization pull request #359

ziyan:
Some leftover fixes for pull request #359 
- Build mo files in CAMKE_BINARY_DIR
- Only build mo file when msgfmt is available
- Put mapping of locale domain and source directory into updatetranslations.bash
- Add gettext as a build dep


ziyan:
@rdiankov Please review. This should be ready to merge. Thanks!

rdiankov:
awesome!

PR #364, opened 2015-09-04
Fix a segfault at "mux.c:237: s->internal->nb_interleaved_streams++;"…

ziyan:
… due to self allocated avformat context not initializing the internal pointer.


rdiankov:
awesome! good job ;0)

PR #367, opened 2015-09-21
Whenever got_packet is 0, avcodec_encode_video2 will free the caller'…

ziyan:
…s buffer. So we cannot pass openrave's buffer in avcodec_encode_video2. Let avcodec_encode_video2 allocate the buffer itself.


rdiankov:
awesome!

PR #372, opened 2015-10-21
Tutorial for install in Ubuntu in RST format

Cescuder:
Hi,
I have removed the previous .txt file, and make the new one using .rst format. I dunno if it'll be ok for you. Thanks anyway.

Regards,

Carlos Rubert
Researcher at RobInLab
Universitat Jaume I(Castellon, Spain)


rdiankov:
thanks! much better

PR #374, opened 2015-11-04
alternative workaround for QTBUG-22829

CaeruleusAqua:
alternative workaround for QTBUG-22829
The old one no longer works with boost 1.57


rdiankov:
cool! thanks!

PR #384, opened 2016-01-24
Add log4cpp support to openrave.

ziyan:
Right now a problem remains with logging output being split into two lines because of the function name, filename and line number:

```
2016-01-24 11:41:01,959 openrave [DEBUG] [environment-core.h:117 Environment] 
2016-01-24 11:41:01,959 openrave [DEBUG] setting openrave home directory to /private/cache/openrave
2016-01-24 11:41:01,959 openrave [DEBUG] [environment-core.h:797 SetPhysicsEngine] 
2016-01-24 11:41:01,959 openrave [DEBUG] disabling physics
```

This is because RAVELOG_XXX actually makes two separate logging calls. And we need to discuss on how to put these debugging information into a single logging message.


ziyan:
Ok, the problem with the separate lines for filename and line number has been solved.

rdiankov:
why is this closed?

ziyan:
Because, I am implementing one using log4cxx which has the TRACE level.

PR #386, opened 2016-01-26
Add log4cxx support to openrave.

ziyan:
Adds colorized log4cxx support to OpenRAVE. This optional feature can be turned on by `-DOPT_LOG4CXX`.

Added `RaveGetLogger()` that returns a `log4cxx::LoggerPtr` that points to the root logger.

Tested both multi byte and wide char logging.


rdiankov:
awesome!~

PR #393, opened 2016-04-11
Problem with drawarrow -- wrong color / segfault

orthez:
I ran into some problems with the drawarrow method when called from the openravepy interface.

(1) choosing an alpha value different from 1.0 lead to a segfault
(2) color of cylinder of arrow and cone of arrow did not match up. in particular, the color of the cylinder remained gray, instead of having the specified color, only the cone did change the color

I fixed both issues inside of the qtcoin interface by computing a color material (SoMaterial), and then setting it to the parent node. Color of the cylinder and cone are now consistent, and i can choose different alpha values without segfault.


rdiankov:
thanks!

PR #396, opened 2016-05-10
add header and namespace for std shared pointers

mikaelarguedas:
I couldnt compile the sources on ubuntu 14.04 without these changes. Hopefully this will allow prevent other users to run into the same issue.


rdiankov:
already merged, thanks!

PR #401, opened 2016-05-17
Fix GCC6 errors

morxa:
With the 6.1 release, GCC switched to compile with `-std=c++11` as default. This caused several errors:
* functions with a non-void return type must return a value
* implicit conversion from pointer to bool is not allowed anymore
* ambiguous type declarations (e.g. `tuple` instead `std::tuple` or `boost::python::tuple`) are not allowed anymore
* template argument deduction is stricter (see [this StackOverflow answer](http://stackoverflow.com/a/9642232/2785041) for more details)

This pull request fixes all issues with GCC6.1. Additionally, it removes an unused function from the OSGViewerWidget, which did not compile with recent versions of OSG.


rdiankov:
thanks!

PR #403, opened 2016-05-25
Fix undefined symbol errors

morxa:
This fixes three undefined symbol errors that occurred during startup of openrave:
* In rplanners, fix the signature of `CheckRamp` to match the declaration in the header file
* In `ConfigurationCache`, add a missing definition for `UpdateFreeConfigurations`. This function was declared but never defined. I implemented the function similar to the other update functions, please check if this makes sense.
* Remove the `SubParabolicSmoother`. It looks like it was never used, its implementation is not part of the library (it was removed in 524ecc83). Looking at the code, it seems like it was not possible to use it at all (the implementation was missing). Remove it to fix an undefined symbol error.

It looks like the undefined symbol errors only occurred sometimes; depending on the compiler flags. In fact, my local version has all those symbols defined, but when I build with Fedora's `%optflags%`, the symbols are not defined. I didn't check which flags caused the errors because those fixes work everywhere.

I also tried adding the `SubParabolicSmoother` back to the library, but it gave multiple compiler errors.

### DIFF
--- a/plugins/configurationcache/configurationcachetree.cpp
+++ b/plugins/configurationcache/configurationcachetree.cpp
@@ -1268,6 +1268,11 @@ int ConfigurationCache::UpdateCollisionConfigurations(KinBodyPtr pbody)
     return _cachetree.UpdateCollisionConfigurations(pbody);
 }
 
+int ConfigurationCache::UpdateFreeConfigurations(KinBodyPtr pbody)
+{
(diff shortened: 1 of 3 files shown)


rdiankov:
thanks. subparabolicsmoother is still under development, so np.

morxa:
Thanks for merging! I wasn't sure about the subparabolicsmoother, but it totally makes sense if it's still under development.

PR #404, opened 2016-05-25
Only create symlink for the openrave binary if a bin suffix is given

morxa:
If the bin suffix is empty, do not create the symlink, because then the
symlink points to itself (e.g. /usr/bin/openrave -> /usr/bin/openrave)
and the binary is deleted.


rdiankov:
thanks!

PR #406, opened 2016-06-01
Look for fparser version 4.5 instead of 4.4.3

kb0n:
Seems reasonable, as the provided fparser version is 4.5 as well.
I added `QUIET` because subsequent checks are made, so an error at an earlier phase seems redundant.


rdiankov:
sure, i see no problem with this atm.

rdiankov:
thanks

Issue #405, opened 2016-06-01
more recent sympy versions


kb0n:
Hi there,

I'm currently in the process of adapting ikfast to more recent sympy versions,  v0.7.6 to be more precise. Not only for using it with Fedora23, but also for the packaging of Openrave for Fedora.

Anyhow, I ended up with a version that works for us. I'm fairly certain though that many scenarios of ikfast were not triggered, so I was wondering if there was any way to thoroughly test ikfast.

I appreciate the help. I'll also create a pull request soonish with my current patches, so you can keep track of the progress, but also for Q&A on aspects of ikfast or other problems I might encounter. If that's ok with you.
I imaging there is a general desire to be compatible with a more recent version rather than being stuck with 0.7.1?

Cheers

rdiankov:
Awesome!
By the way, for testing stuff, you can try:

python run_tests.py --test test_ikfast_mandatory

it requires nose testing library.

in sandbox i also have:

python updateikfiles.py --forcegenerate --destdir=../plugins/ikfastsolvers --testnum=100

hope this helps

kb0n:
Thanks, I will try those.

Why do you run the second command in a sandbox? Should I be worried if I dont'?

rdiankov:
in plugins/ikfastsolvers you'll see there's ikfast generated solutions for some of the samples we have. it's not necessary, but it helps people run openrave without having to wait 5min to genreate the ikfast solution. the updateikfiles.py updates to files.

PR #407, opened 2016-06-01
Make ikfast compatible with sympy-0.7.6

kb0n:
As mentioned in #405, I'm trying to get ikfast to work with more recent versions of sympy.
This is my current progress and I think a pull request makes development easier, being able to ask questions, comment on commit and all.

### DIFF
--- a/python/ikfast.py
+++ b/python/ikfast.py
@@ -6268,7 +6268,7 @@ class IKFastSolver(AutoReloader):
                         linearlyindependent = True
                     break
                 else:
-                    log.info('not all eigenvalues are > 0. min is %e', min([Abs(f) > eps for f in eigenvals]))
+                    log.info('not all abs(eigenvalues) are > 0. min is %e', min([Abs(f) for f in eigenvals if Abs(f) > eps]))
             if not linearlyindependent:
                 raise self.CannotSolveError('equations are not linearly independent')
 
--- a/python/ikfast.py
+++ b/python/ikfast.py
@@ -3233,7 +3233,7 @@ class IKFastSolver(AutoReloader):
                         endbranchtree=[AST.SolverSequence([leftovervarstree])]
                         unusedsymbols = []
                         for solvejointvar in solvejointvars:
-                            usedinequs = any([var in rawpolyeqs[0][0].gens or var in rawpolyeqs[0][1] for var in self.Variable(solvejointvar).vars])
+                            usedinequs = any([var in rawpolyeqs[0][0].gens or var in rawpolyeqs[0][1].gens for var in self.Variable(solvejointvar).vars])
                             if not usedinequs:
                                 unusedsymbols += self.Variable(solvejointvar).vars
                         AllEquationsExtra = []
--- a/python/ikfast.py
+++ b/python/ikfast.py
@@ -2344,7 +2344,7 @@ class IKFastSolver(AutoReloader):
         Links = LinksRaw[:]
         LinksInv = [self.affineInverse(link) for link in Links]
         T = self.multiplyMatrix(Links)
-        self.Tfinal = zeros((4,4))
+        self.Tfinal = zeros(4,4)
         self.Tfinal[0,0:3] = (T[0:3,0:3]*basedir).transpose()
         self.testconsistentvalues = self.ComputeConsistentValues(jointvars,self.Tfinal,numsolutions=4)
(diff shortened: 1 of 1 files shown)


kb0n:
So here is my first question. In recent sympy versions, the degree of zero is `NegativeInfinity`, whereas for sympy-0.7.1 it was `-1` as far as I can tell. Easy to check with 
```python
from sympy import *
x = symbols('x')
a = Poly(0, x)
a.degree()
```
This leads to a problem in the `SolvePairVariablesHalfAngle` method, when iterating over `range(mindegree, maxdegree+1)`. My quick dirty hack for testing was to add
```python
if mindegree == S.NegativeInfinity:
    mindegree = -1
```
which got the job done, but I assume there is a better way to do this?

rdiankov:
thanks for the patches! i tried them on sympy 0.7.1 and  the zeros((4,4)) -> zeros(4,4) breaks. is there a more backwards compatible way of doing this? Ie we setup a function in ikfast.py that creates a zeros array, which then tests for the sympy version?

as for NegativeInfinity, i think that will work. can you also include it in your branch?

best

kb0n:
Yes, obviously the zeros() and ones() break for 0.7.1. For a backward compatible solution I would've suggested the same, it should work with some sort of wrapper functions.
My intention was to move on from 0.7.1 though. As far as I understand from sympy's release notes, they're trying to be more "correct" with each update (besides new functionality), so I assumed you'd prefer having a more recent sympy version as well. So how should I handle it, with backwards compatiblity or going for a more recent version (sympy 1.0 was released in March, maybe that would be a desirable goal)? :)

Patch for NegativeInfinity will be added shortly.

rdiankov:
ikfast is stress tested every day with a lot of robots only on sympy 0.7.1.

i'll try to convert our internal infrastructure to something more recent, but this takes time since incompatibilities pop up randomly.

i'm fine with shooting for sympy 1.0 compatibility. until then, we'll have to consider all patches compatible to 0.7.1 (this is easy to test with sympy.__version__

cielavenir:
I don't see the corresponding fix of this error

```
    if len(coeffs) == 1 and Abs(coeffs[0]) < 2*(10.0**-self.precision):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Abs(1.0*new_r22 + 1.0) < 2.0e-8

    def __bool__(self):
>       raise TypeError("cannot determine truth value of Relational")
E       TypeError: cannot determine truth value of Relational
```

cielavenir:
I merged production to ikfast_sympy076, pushed to https://github.com/rdiankov/openrave/commits/ikfast_sympy076_merge/ .

Relational error is reproduced, fixed by https://github.com/rdiankov/openrave/commit/fd519d02b634f2d2b7a073d410b24a8491c78059 .

Then I got this

```
/private/cache/openrave/kinematics.07d2f91176824501770bf9b1d22d00d3/ikfast0x1000004c.Transform6D.x86_64.0_1_3_4_5_6_f2.cpp:15063:22: error: ‘tj5’ was not declared in this scope; did you mean ‘j5’?
15063 | IkReal x719=(new_r02*tj5);
      |                      ^~~
      |                      j5
```

This is forcibly fixed by https://github.com/rdiankov/openrave/commit/2ae582f23048d58fbb6973b7a680b8c582181ffc (I don't know how the iksolver is different like this)

python/ikfast.py
```diff
@@ -7697,6 +7658,9 @@ def SolvePairVariablesHalfAngle(self,raweqns,var0,var1,othersolvedvars,subs=None
             newpolyeqs = [Poly(eq,varsyms[1-ileftvar].htvar) for eq in polyeqs]
             mindegree = __builtin__.min([max(peq.degree_list()) for peq in newpolyeqs])
             maxdegree = __builtin__.max([max(peq.degree_list()) for peq in newpolyeqs])
+            # limit mindegree to -1
+            mindegree = -1 if mindegree is S.NegativeInfinity else mindegree
```
cielavenir, line 7662: main difference between my branch is only this

PR #408, opened 2016-06-01
Increase working precision for polyroots (fix ikfast compatibility with mpmath-0.19)

kb0n:
Since the tolerance calculation has been fixed and become more precise in mpmath-v0.19, higher working precision is required to avoid a NoConvergence error.
See https://github.com/fredrik-johansson/mpmath/commit/a2e7c3aeff03c7d68ff9446691dd7a2e94c2b59f for the commit in mpmath.

I tried a few values for `extraprec` and 18-20 seemed to work (maybe there is a "proper" way to find the right precision?). Adjusting `maxsteps` didn't help.

Since those arguments seemed to be ineffective in prior verisions of mpmath, this commit should be "downward compatible".

### DIFF
--- a/python/ikfast.py
+++ b/python/ikfast.py
@@ -8151,7 +8151,7 @@ class IKFastSolver(AutoReloader):
                 break            
             realsolution = pfinal.gens[0].subs(subs).subs(self.globalsymbols).subs(testconsistentvalue).evalf()
             # need to convert to float64 first, X.evalf() is still a sympy object
-            roots = mpmath.polyroots(numpy.array(numpy.array(coeffs),numpy.float64))
+            roots = mpmath.polyroots(numpy.array(numpy.array(coeffs),numpy.float64),extraprec=20)
             for root in roots:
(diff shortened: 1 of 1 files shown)


rdiankov:
thanks!

kb0n:
Might need to readjust that value though, as test runs with
```
python run_tests.py --test test_ikfast_mandatory
```
raised a couple of `NoConvergence` errors.
I'll update if/when I find reliable values.

rdiankov:
how about extraprec=100?
it really doesn't matter if it is slow...

kb0n:
That alone didn't do it, but the combination of `extraprec=100, maxsteps=750` did.
I'm getting `AssertionError` now though, may be numpy related. I'll report as soon as I figure out more about this :)

rdiankov:
i'm using numpy v1.6.2 and mpmath v0.17

i'll revert your changes for now, let me know when you are ready for another pull request.

thanks

kb0n:
Thanks. I hope I can get those ancient versions running on Fedora23 :D
I'll just open a new pull request when then, as I assume this one can't be reopened?

Issue #400, opened 2016-05-15
make error when building qtosgrave plugin with Ubuntu 14.04.4


WHAAAT:
Hello,
normally i had no issue building openrave from source. 
However since the newest Ubuntu I'm getting the following error.
```

/openrave/plugins/qtosgrave/qtosgviewer.cpp: In member function ‘virtual void qtosgrave::QtOSGViewer::_SetTriangleMesh(const float*, int, const int*, int, osg::ref_ptr<osg::TemplateArray<osg::Vec3f, (osg::Array::Type)28u, 3, 5126> >, osg::ref_ptr<osg::DrawElementsUInt>)’:
/home/megatron/openrave/plugins/qtosgrave/qtosgviewer.cpp:1554:17: error: ‘class osg::DrawElementsUInt’ has no member named ‘resizeElements’
     osgindices->resizeElements(numTriangles*3);
                 ^
make[2]: *** [plugins/qtosgrave/CMakeFiles/qtosgrave.dir/qtosgviewer.cpp.o] Error 1
make[1]: *** [plugins/qtosgrave/CMakeFiles/qtosgrave.dir/all] Error 2
make: *** [all] Error 2
```
I've tried different older revisions with same error.

When I disable the plugin I'm getting errors when building plugins/fclrave/CMakeFiles/fclrave.dir/fclrave.cpp.o
this seems to be the same problem like [issues/399](https://github.com/rdiankov/openrave/issues/399)

[here is my full log of the error](https://github.com/rdiankov/openrave/files/264786/log.log.zip)

rdiankov:
i just committed some fixes to master, can you check them out?

thanks

WHAAAT:
hello thank you for fast reply.
The master branch fixed the error for qtosgrave plugin.
However fclrave is still the same. I've attached the log for the master branch 
[log.zip](https://github.com/rdiankov/openrave/files/264906/log.log.zip)

rdiankov:
it looks like we weren't forcing c++11, can you pull from master and try again?

thanks

WHAAAT:
allright! 
There was still a problem with the ros-indigo-fcl package, i had to purge it and install the fcl master branch. Now everything is working fine! Thank you :+1:

hustfzt2:
how to solve the problem about " fclrave" in ROS Indigo as:
openrave/plugins/fclrave/fclspace.h:15:9: error: ‘shared_ptr’ in namespace ‘std’ does not name a type
[...]

Issue #397, opened 2016-05-10
Failed to create name fcl_, interface collisionchecker


conversAlgoDS:
I managed to compil openrave  on my fedora machine but when I run the hanoi example., it fails to display the gui. There is only 1 warning about fcl. Messages to console are however getting displayed.

```
$openrave.py --example hanoi
[plugindatabase.h:577 Create] Failed to create name fcl_, interface collisionchecker
robot resolutions:  [ 0.00195998  0.00198425  0.00328496  0.01093304  0.0106936   0.0234112
  0.02297028]
robot weights:  [ 1.52329723  0.92804625  0.38978808  0.15792621  0.10813445  0.04291764
  0.02730175]
hanoimove disk2 from srcpeg to destpeg
opening hand
moving hand to location
succeeded so grab the disk
try to put the disk in the destination peg
move to position above source peg
move to position above dest peg
move to dest peg
wait for robot to complete all trajectories
done with one disk
hanoimove disk1 from srcpeg to peg
opening hand
moving hand to location
succeeded so grab the disk
try to put the disk in the destination peg
.
.
.
```

hustfzt2:
Have you solved for this problem？

rdiankov:
sounds like you failed to compile the openscenegraph plugin or the soqt plugin for the viewers....

Atanker:
Hello,have you solved it?I have the same problem,and more ,it shows "failed to creat name ode,interface collisionchecker"

Cescuder:
Hi @Atanker 
It sounds that the ode plugin is missing, have you installed it before compiling?

You can download and install it manually, then compile openrave.

yijiangh:
In my scenario, just making sure you have install the correct version of FCL will fix this problem.
Check FCL installation instruction from [Stéphane Caron](https://scaron.info/teaching/installing-openrave-on-ubuntu-16.04.html)

PR #409, opened 2016-06-06
Updated Tutorial for installing in Ubuntu and issue showing Hanoi Example

Cescuder:
Hi,

I have updated the tutorial for installing in Ubuntu 14.04, since last commits on OpenRave code, a new package is needed to be installed in ubuntu Mate.

Also, I have solved the problem with collada files since commit ffa18d9f6312ba8130225a0ff57749b5c4d6558c
which were not being shown in the environment, this solves also the problems with hanoi example explained in this issues:
https://github.com/rdiankov/openrave/issues/397
https://github.com/rdiankov/openrave/issues/400


rdiankov:
thanks!
so it looks like the real change was because of the locale?
that is so weird......  forcing the locale like this makes translation files not work for c++ applications linking to openrave, so do you think you can narrow down exactly what line in c++ (probably somewhere in collada-dom?) it is breaking due to the locale?

thanks!

Cescuder:
Hi, i will take a look.

PR #412, opened 2016-06-13
Re-fix the issue when setting camera transform the focal distance not being respected.

ziyan:
Turns out just need to setDistance before calling setByMatrix.


rdiankov:
thanks!

PR #414, opened 2016-06-27
Fclrave small caching

kyoDralliam:
Caching of the broadphase managers to get some speed up


rdiankov:
already done in fcl-caching2 branch

PR #416, opened 2016-06-27
Parabolicsmoother

Puttichai:
- Implemented a new parabolic interpolation algorithm.

  - For interpolation with a given duration, we try to interpolate a velocity profile with accelerations that minimize their sum squared instead of forcing them to have equal magnitudes.

  - Taking care of inoperative time interval: now the interpolation will always be successful (if the trajectory does not violate joint limits).

- Fixed existing bugs in `parabolicsmoother`


rdiankov:
great job! very nice catches.

PR #418, opened 2016-07-08
Parabolicsmoother

Puttichai:
Fixed handling of the case when endTime is (roughly) zero and updated the python version accordingly.


rdiankov:
thanks!

Issue #419, opened 2016-07-16
replaceObject not available in fcl


yn:
When compiling current master, I get this:

/openrave/plugins/fclrave/fclmanagercache.h:294:43: error: no member named 'replaceObject' in
      'fcl::BroadPhaseCollisionManager'
                                pmanager->replaceObject(itcache->second.vcolobjs.at(ilink).get(), pcol.get(), false);
                                ~~~~~~~~  ^

I am not sure where replaceObject is supposed to be coming from; it is not present in the current sources at https://github.com/flexible-collision-library/fcl

kyoDralliam:
Openrave is currently using a modified version of the fcl library that you can find at https://github.com/rdiankov/fcl/tree/kenjiSpeedUpAdditions

rdiankov:
oops, sorry about this!
we'll put in some compiler flags to detect this automatically

until then, please turn off compiling FCL with: -DOPT_FCL_COLLISION=OFF

rdiankov:
we just committed CMake fixes for fcl, can you check again?

thanks

yn:
The previous error related to replaceObject went away. I have now the following errors (actually, I had them previously, but didn't mention them because I was trying to make the report isolated:

```
[ 99%] Building CXX object plugins/fclrave/CMakeFiles/fclrave.dir/fclrave.cpp.o
In file included from /Users/yn/code/others/openrave/plugins/fclrave/fclrave.cpp:16:
In file included from /Users/yn/code/others/openrave/plugins/fclrave/plugindefs.h:102:
In file included from /usr/local/Cellar/fcl/0.4.0/include/fcl/collision.h:42:
In file included from /usr/local/Cellar/fcl/0.4.0/include/fcl/math/vec_3f.h:43:
/usr/local/Cellar/fcl/0.4.0/include/fcl/math/math_details.h:350:5: warning: 
      'register' storage class specifier is deprecated [-Wdeprecated-register]
    register T tmp = rs[0][1];
    ^~~~~~~~~
/usr/local/Cellar/fcl/0.4.0/include/fcl/math/math_details.h:367:5: warning: 
      'register' storage class specifier is deprecated [-Wdeprecated-register]
    register T inrsdet = 1 / det;
    ^~~~~~~~~
In file included from /Users/yn/code/others/openrave/plugins/fclrave/fclrave.cpp:20:
/Users/yn/code/others/openrave/include/openrave/plugin.h:59:48: warning: 
[...]

PR #420, opened 2016-07-16
Make fixes for OS X

yn:
The default homebrew openscenegraph installation doesn't package osgQt - that means that a compilation error happened prior to this PR, because the condition checks for OPENSCENEGRAPH only, not for OSGQT

On OS X, intl isn't part of the default C library, so some CMAKE fixes are in order to get that working.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,7 @@ include(CheckCXXSourceRuns)
 include(CheckCXXCompilerFlag)
 include(CheckTypeSize)
 include(InstallSymlink)
+include(FindIntl)
 
 find_package(PkgConfig) # pkg_check_modules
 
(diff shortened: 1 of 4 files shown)


rdiankov:
YN, all your changes are good except  `#  set(PLUGINS ${PLUGINS} fclrave)`

We'll be pushing in a fix later today, so can you test afterwards?

thanks

yn:
Will test as soon as I see you fix.

rdiankov:
just pushed a fix, can you double check? thanks

yn:
I reverted the part where I comment out fclrave. I think you can merge this PR now. As I mentioned in #419, FCL still errors out, but that seems like a separate issue, so let's address it separately.

rdiankov:
thanks!
we're resolving the fcl issue right now, but that is separate from your commits.

PR #421, opened 2016-07-21
Parabolicsmoother

Puttichai:
Bug fixes:

 - Corrected elapsed time checking in `Check2`
 - Correctly report solving status of `CalculateLeastBoundInoperativeInterval` when the solution is zero
 - Added checking of the interpolation of the last segment in shortcutting to make sure all successful shortcuts are good.
 - Other minor fixes


rdiankov:
awesomeness!~

PR #423, opened 2016-07-27
jacobian convergence improvement

kanbouchou:
changes made
- fixed computation of angle error between current and goal. 
- made gain from 1.

tested
 - tested on 6d robots and verified that fast convergence without causing unstability around goal.


rdiankov:
awesome! thanks

Issue #296, opened 2014-04-25
PCRE not found


quangounet:
Not sure whethis has been reported yet.
When compiling master on Ubuntu 12.04, I got this error
~/git/openrave/src/libopenrave-core/environment-core.h:26:21: fatal error:
pcrecpp.h: No such file or directory
This can be fixed by 
sudo apt-get -y install libpcre++-dev
Maybe a problem when building the dependencies?

PeterBowman:
I just run into this on Ubuntu 16.04 and the `latest_stable` branch.

rdiankov:
we fixed  abunch of stuff recently on master branch. have you tried it?

PeterBowman:
I'm getting these errors on current master (34c985c) with `-DOPENRAVE_PLUGIN_FCLRAVE=OFF`:

```shell
/home/asibot-msi/git/openrave/plugins/ikfastsolvers/ikfastsolver.cpp: In member function ‘void IkFastSolver<IkReal>::_SetJacobianRefine(OpenRAVE::dReal, int)’:
/home/asibot-msi/git/openrave/plugins/ikfastsolvers/ikfastsolver.cpp:118:9: error: ‘_jacobinvsolver’ was not declared in this scope
         _jacobinvsolver.SetErrorThresh(_fRefineWithJacobianInverseAllowedError)
         ^
/home/asibot-msi/git/openrave/plugins/ikfastsolvers/ikfastsolver.cpp: In instantiation of ‘void IkFastSolver<IkReal>::Clone(OpenRAVE::InterfaceBaseConstPtr, int) [with IkReal = double; OpenRAVE::InterfaceBaseConstPtr = boost::shared_ptr<const OpenRAVE::InterfaceBase>]’:
/home/asibot-msi/git/openrave/plugins/ikfastsolvers/ikfastsolver.cpp:2313:1:   required from here
/home/asibot-msi/git/openrave/plugins/ikfastsolvers/ikfastsolver.cpp:806:27: error: ‘const class IkFastSolver<double>’ has no member named ‘_jacobinvsolver’
         _SetJacobianRefine(r->_fRefineWithJacobianInverseAllowedError, r->_jaco
                           ^
[...]

PR #430, opened 2016-08-06
Set libpcrecpp_FOUND when using local pcre

kktk-KO:
This pull request is related to an issue [#296](https://github.com/rdiankov/openrave/issues/296).

A problem here is that when local libpcre is loaded in `openrave/CMakelits.txt`, a variable `libpcrecpp_FOUND` is still unset.
And if `libpcrecpp_FOUND` is unset, a include directory of `libpcre` is not passed to the compiler.
However, `openrave/src/libopenrave-core/environment-core.h` always includes `<pcrecpp.h>` ignoring `libpcrecpp_FOUND`.

In this pull request, I set the variable to `1` when local libpcre is used.


rdiankov:
great catch, thanks!

PR #433, opened 2016-08-15
Fclrave local info

kyoDralliam:
Moves the KinBodyInfo in FCLSpace from each KinBody UserData map to a map stored inside the FCLSpace. This change allows to access to KinBodyInfo without taking an expensive lock and results in a non-negligeable performance gain.


rdiankov:
yay~

PR #434, opened 2016-08-15
Undefined behaviour when creating a shared_ptr from a naked pointer inheriting from enable_shared_from_this

kyoDralliam:
This seems to be currently fine with boost::shared_ptr (but it is specified as an UB in the documentation), and it is definitively not fine with std::shared_ptr from c++11.


rdiankov:
bulk of code has already been merged to production

PR #436, opened 2016-08-22
Puttichai/newparabolicsmoother

Puttichai:
A new implementation of parabolicsmoother.


rdiankov:
woot!! finally~ after 4 months~
now your code will run on all of our production systems ;0)

PR #437, opened 2016-08-26
Test link link collision attached bodies

kyoDralliam:
The previous test was not correct python code


rdiankov:
looks like already merged

PR #440, opened 2016-09-01
Added different GetPublishedBody... routines and safely removed some vector at() bound checks

woodychow:


rdiankov:
great! good job! finally~

PR #441, opened 2016-09-05
Correctly set the libs when writing geometry but not visual

woodychow:
Code would crash if IsWrite("geometry") is true and IsWrite("visual") is false. I am not sure if this is the right fix though.


rdiankov:
yay!

PR #444, opened 2016-09-12
Prevent crashing due to log4cxx logger already gone during RaveGlobal destroy.

ziyan:
Just added some simple checks before using the logger.


rdiankov:
awesome!

PR #449, opened 2016-09-24
Puttichai/fixcheckmanipconstraints3

Puttichai:
- Fixed `CheckManipConstraints3` which previously skipped checking the end of the input segment
- Fixed Ramp1D acceleration evaluation


Puttichai:
fixed in commits 59aa359c3 and 4cdbcceb9c9.

PR #450, opened 2016-09-27
Fixcheckmanipconstraints

Puttichai:
In `CheckManipConstraints`, we should skip checking  in the later part only when there is only one  ramp and the ramp is too short.


rdiankov:
yay!

PR #451, opened 2016-09-29
 Remove PyUserObject, which stores python::boost::object in UserData

woodychow:
It is dangerous. Needs GIL.


rdiankov:
woody, why did you change GetUserData and remove PyUserObject? I thought we were only removing SetUserData overload until we can figure out a better alternative...

woodychow:
Without SetUserData there won't be any PyUserObject in the system. I thought it would be a better idea to just remove all the PyUserObject related code. Should I only change SetUserData?

PR #457, opened 2016-11-01
When writing to collada, write cylinder as cylinderz such that we don't have to rotate axis.

ziyan:
After this change, when reading dae file, we will respect both <cylinder> and <cylinderz>. When writing, we will only write <cylinderz>.


rdiankov:
thanks!

PR #460, opened 2016-11-16
when done searching extra tag, try to find out unit in asset tag 

braineo:
asset tag could appear in the root level


rdiankov:
awesome!

PR #461, opened 2016-11-28
Russ verification

rmislam:
Changes in openrave to deal with calibboards being links on robots instead of kinbodies that are imported into the scene


rdiankov:
awesome!

PR #464, opened 2016-12-15
Found circular dependency in robot controller. Rosen made weakptr fix.

ziyan:
IdealController was using RobotBasePtr instead of RobotBaseWeakPtr.


rdiankov:
awesome!

PR #468, opened 2017-01-12
Clean C API

sirop:
Cleaned C API to be compiled with gcc.  

Functions ( signatures) changed:  
OPENRAVE_C_API void ORCInitialize(bool bLoadAllPlugins=true, int level=Level_Info) -->
OPENRAVE_C_API void ORCInitialize(int bLoadAllPlugins, int level)

OPENRAVE_C_API bool ORCBodyInitFromTrimesh(void* body, void* trimesh, bool visible) -->
OPENRAVE_C_API int ORCBodyInitFromTrimesh(void* body, void* trimesh, int visible)

OPENRAVE_C_API bool ORCEnvironmentLoad(void* env, const char* filename) -->
OPENRAVE_C_API int ORCEnvironmentLoad(void* env, const char* filename)

OPENRAVE_C_API bool ORCEnvironmentSetViewer(void* env, const char* viewername) -->
OPENRAVE_C_API int ORCEnvironmentSetViewer(void* env, const char* viewername)   

The main  motivation for this cleanup was the desire to use OpenRAVE API within real time components upon https://github.com/machinekit/machinekit . 
The first tests look promising:  
calculating Angular Jacobian for a 5 DOF robot takes less than 15 microseconds,  inversion of a 3x3 jacobian matrix takes less than 35 microseconds in the worst case. Tests were run on Intel(R) Celeron(R) CPU  N2930  @ 1.83GHz  with the develop branch of OpenBLAS.  

*.cpp and *.c files have been uncrustify'ed.

### DIFF
--- a/include/openrave_c/openrave_c.h
+++ b/include/openrave_c/openrave_c.h
@@ -65,7 +65,7 @@ enum DebugLevel {
 OPENRAVE_C_API void ORCSetDebugLevel(int level);
 
 /// \brief Calls \ref RaveInitialize
-OPENRAVE_C_API void ORCInitialize(bool bLoadAllPlugins=true, int level=Level_Info);
+OPENRAVE_C_API void ORCInitialize(int bLoadAllPlugins, int level);
 
 /// \brief Calls \ref RaveDestroy
 OPENRAVE_C_API void ORCDestroy();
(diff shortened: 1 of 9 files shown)


rdiankov:
thanks! looks good

PR #472, opened 2017-02-21
Add input checks to _ComputeConvexHull and redirect QHull error messages to OpenRAVE log

woodychow:
Without the checks, the code will crash

### DIFF
--- a/plugins/grasper/graspermodule.cpp
+++ b/plugins/grasper/graspermodule.cpp
@@ -1612,6 +1612,16 @@ protected:
         boost::mutex::scoped_lock lock(s_QhullMutex);
         vconvexplanes.resize(0);
 #ifdef QHULL_FOUND
+        if ( vpoints.empty() ) {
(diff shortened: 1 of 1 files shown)


woodychow:
The following input checks are added

nan
points.size() > 0
dim >= 2

Also, QHull error messages are now redirected to the log.

rdiankov:
thanks!
can we remove the nan check? probably system won't segfault if it is nan anyway...

woodychow:
Done

rdiankov:
thanks!
by the way, if you are going to output errfile, you should probably clear it everytime before you make the qhull call or you'll be displaying old results (same goes for outfile if you do output it)

woodychow:
I was assuming that if exit code == 0, nothing will be written to errfile. That is probably true but I will change the code.

rdiankov:
awesome!

woodychow:
thanks!

PR #476, opened 2017-03-14
minor bugfix and speed ups

kanbouchou:
- minor fix of typological logic
- use empty for list's emptiness check
- stop converting string to char* (which then get converted to string again)


rdiankov:
thanks, but next time have a merge request to production.

kanbouchou:
ok, will send to production next time

PR #477, opened 2017-03-27
minor fixes

kanbouchou:
fixed issues reported by clang-tidy

- delete string comparison whose result was unused 
- delete an always-true check


rdiankov:
thanks

PR #480, opened 2017-04-05
Optimize PadMesh function

woodychow:
Code is 7~8x faster on my computer.

### DIFF
--- a/python/databases/convexdecomposition.py
+++ b/python/databases/convexdecomposition.py
@@ -69,7 +69,7 @@ __license__ = 'Apache License, Version 2.0'
 if not __openravepy_build_doc__:
     from numpy import *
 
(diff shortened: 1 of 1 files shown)


rdiankov:
cool, a minor suggestion:

- perhaps you can cache vertices[indices[:,INDEX]] and newvertices[newindices[:, INDEX]] so you don't look it up several times.

thanks

woodychow:
I am not sure if that would help. From what I understand, those are just views of the original matrix. I don't believe creating the view will take much time at all. Let me do some experiments.

woodychow:
Caching the view helps a tiny bit; perhaps the number of vertices is too small in my experiment. Anyhow, let's cache the views.

rdiankov:
looks good, thanks!

PR #481, opened 2017-04-07
Optimize PyTriMesh constructor and GetTriMesh function

woodychow:
by eliminating python calls / using memcpy


rdiankov:
awesome~

PR #482, opened 2017-04-07
Add merge duplicated vertices to PadMesh

woodychow:
PadMesh could break down when there were duplicated vertices in the vertices array. Each triangle would translate by padding * normal but no triangles were connected to each other


rdiankov:
yay~

PR #485, opened 2017-04-17
Take care of degenerate triangles in PadMesh

woodychow:
by setting the normals of all degenerate triangles to 0 (they would be nan otherwise).


rdiankov:
ok for now~~~

PR #489, opened 2017-05-16
Stop using numpy.tile in TransformPoints and TransformInversePoints

woodychow:
as they are inefficient


woodychow:
Also, this will work properly if one point is given as an 1D array. Before, these functions would return a 3x3 matrix.

rdiankov:
thanks

PR #496, opened 2017-06-19
Set window title in gui thread.

ziyan:
Kevin reported that sometimes window title does not change, causing his code to fail to find the correct window.


rdiankov:
thanks

PR #499, opened 2017-06-26
Update KinBody::Joint::GetTorqueLimits to use GetMaxTorque if available and _vmaxtorque is not set

cielavenir:
Currently I cannot use robot.GetDOFTorqueLimits() (all returned values are zero). `[e.GetMaxTorque() for e in robot.GetJoints()]` is a workaround, but it is not intuitive.

### DIFF
--- a/src/libopenrave/kinbodyjoint.cpp
+++ b/src/libopenrave/kinbodyjoint.cpp
@@ -877,7 +877,12 @@ void KinBody::Joint::GetTorqueLimits(std::vector<dReal>& vmax, bool bAppend) con
         vmax.resize(0);
     }
     for(int i = 0; i < GetDOF(); ++i) {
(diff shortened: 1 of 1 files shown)


rdiankov:
if you are going to change GetTorqueLimits, then you also need to update SetTorqueLimits

cielavenir:
Yes... One of the ideas can be `if( _info._vmaxtorque[i] == 0 && _info._infoElectricMotor ) {`

cielavenir:
(Committing this anyway)

cielavenir:
At the time I wrote this, I didn't know that nominal torque and max torque are different. Current design let us care how to plot the torque limits. Let me close this after a while.

cielavenir:
Closing this until I get serious issue in testing trajectory.

PR #502, opened 2017-06-30
Collada update

fuzzyTew:
Required to build on newer systems.


rdiankov:
sorry, but i cannot accept merges into latest_stable...

PR #506, opened 2017-07-18
Use with to correctly close python file handlers

woodychow:
I ran out of file descriptors on my machine. This seems to be the culprit.


rdiankov:
awesome!

PR #509, opened 2017-07-21
env was cloned twice, and the first clone was not destroyed

woodychow:
The environment is cloned again in _WorkerThread. Therefore, cloning beforehand is not necessary. In addition, the first clone was not being destroyed.


rdiankov:
code is correct, there's no memory leak here

PR #510, opened 2017-07-26
Fix collada memory leak

woodychow:
To reproduce the memory leak: Just open an environment and keep calling env.Save()

I can also change the code such that we only call SetGlobalDAE for libxml2 2.9+. However, the #ifdefs might make the code hard to read. Please let me know.


rdiankov:
please add comments and a #if for libxml2 version
thanks

woodychow:
Updated. Thanks.

rdiankov:
yay~~

PR #521, opened 2017-08-21
Code to support external QT loop

woodychow:


plugins/qtosgrave/qtosgrave.cpp
```diff
@@ -101,6 +101,15 @@ InterfaceBasePtr CreateInterfaceValidated(InterfaceType type, const std::string&
...
                 }
                 else {
+                    // The function below can be posted before qt app's exec()
+                    // is called, so the function below won't get executed
+                    // immediately and the GetViewer function will time out.
+                    //
+                    // QThread::eventDispatcher(), which can be used to tell if
+                    // an event loop is running in a thread, is only available
+                    // in QT5.
+                    return qtosgrave::CreateQtOSGViewer(penv, sinput);
```
rdiankov: please explain this more.

woodychow: Posted functions (in this case QtOSGViewerCreator) will not be executed before the app is started. Therefore, RaveCreateViewer will timeout and fail if it is called before the app is launched.

rdiankov:
ok, works for me~

PR #525, opened 2017-09-08
Fix projection mode switching. Add Zoom function for orthogonal projection

woodychow:
The projection mode fix is similar to what @ziyan had done in https://github.com/rdiankov/openrave/tree/qtosg-znear

Shall I change the code such that the Zoom function also affects perspective projection?


woodychow:
I have a branch in planningcommon that uses it. But since Zoom only affects orthogonal projection at the moment, I don't want to commit it yet.

rdiankov:
thanks

PR #530, opened 2017-09-27
Python3 support

EdsterG:
I'm working on improving python3 compatibility while retaining python2 backwards compatibility. I've gotten the code to compile and run with python3 using a very hacky approach however I don't know how to modify the cmake files to officially support python3.


EdsterG:
Compile with:

```
PYTHON_LIB_DIR=$(python3 -c "import sys; print(sys.prefix)")
PYTHON_VERSION=$(python3 -c "import sys; print('{}.{}'.format(*sys.version_info[:2]))")
cmake .. -DPYTHON_LIBRARY=${PYTHON_LIB_DIR}/lib/python${PYTHON_VERSION}/config-${PYTHON_VERSION}m-x86_64-linux-gnu/libpython${PYTHON_VERSION}m.so -DPYTHON_EXECUTABLE=$(which python3) -DPYTHON_INCLUDE_DIR=${PYTHON_LIB_DIR}/include/python${PYTHON_VERSION}m
```

EdsterG:
I thought int() was more explicit than //, but that's purely subjective.

rdiankov:
Cool~ I did run into some problems.
Can you fix the warnings with the following command:

```
openrave.py --listdatabases
```

I get a lot of warnings like:

```
/usr/lib/python3.5/importlib/_bootstrap.py:222: RuntimeWarning: to-Python converter for boost::multi_array<float, 1ul, std::allocator<float> > already registered; second conversion method ignored.
  return f(*args, **kwds)
/usr/lib/python3.5/importlib/_bootstrap.py:222: RuntimeWarning: to-Python converter for boost::shared_ptr<openravepy::PyViewerBase> already registered; second conversion method ignored.
  return f(*args, **kwds)
No module named 'openravepy.databases.DatabaseGenerator
```



Also "iterkeys" is used in the bindings, which you'll have to convert to `keys`

```
bindings/openravepy_kinbody.cpp:        boost::python::list iterkeys = (boost::python::list)odict.iterkeys();
bindings/openravepy_kinbody.cpp:        for (int i = 0; i < boost::python::len(iterkeys); i++) {
bindings/openravepy_kinbody.cpp:            int linkindex = boost::python::extract<int>(iterkeys[i]);
bindings/openravepy_kinbody.cpp:            object olinkaccelerations = odict[iterkeys[i]];
bindings/openravepy_kinbody.cpp:        boost::python::list iterkeys = (boost::python::list)odict.iterkeys();
bindings/openravepy_kinbody.cpp:        for (int i = 0; i < boost::python::len(iterkeys); i++) {
bindings/openravepy_kinbody.cpp:            int linkindex = boost::python::extract<int>(iterkeys[i]);
bindings/openravepy_kinbody.cpp:            object oforcetorque = odict[iterkeys[i]];
bindings/openravepy_int.cpp:        boost::python::list iterkeys = (boost::python::list)odict.iterkeys();
bindings/openravepy_int.cpp:        size_t num = boost::python::len(iterkeys);
bindings/openravepy_int.cpp:            std::string key = boost::python::extract<std::string>(iterkeys[i]);
bindings/openravepy_int.cpp:            std::string value = boost::python::extract<std::string>(odict[iterkeys[i]]);
```


There was also some cmake stuff to patch up since openrave wouldn't build with python3 since i had a python2 installation. I started merged your changes and mine into a python3_support branch on rdiankov, please take a look.

EdsterG:
I've updated iterkeys() to keys() and iteritems() to items() in the cpp file.

The warnings have to do with modules being initialized multiple times. For example, calling init_python_bindings() multiple times in the same run. I was able to fix some of the warnings by creating a variable 'bindings_initialized' inside of bindings.cpp and putting the content inside init_python_bindings() into an if-block.

In general, I think the initialization part of openravepy_int.cpp gets called multiple times, that's why the warnings show up.

undisputed-seraphim:
Was scrolling through old issues and came across this; we support compiling with Python3 now.

I suppose it is fine to close this.

PR #531, opened 2017-09-28
Ikfast bug fixes

EdsterG:
These changes fix small bugs in ikfast.py while improving support for future versions of sympy. Note that these changes do not affect the performance of ikfast.py with sympy version 0.7.1


rdiankov:
thanks dude~

PR #533, opened 2017-10-27
Adds support for SendJSONCommand. 

ziyan:


rdiankov:
thanks, i added code to check for rapidjson existence and have openrave compile without it.

PR #534, opened 2017-11-10
Merge outputsignals

kanbouchou:
better handling of outputSignals when merging two trajectories and configuration specifications


rdiankov:
thanks!

PR #535, opened 2017-11-21
Specify whether the dof value is greater than the upper limit

woodychow:
or smaller than the lower limit in the log


rdiankov:
thanks

PR #536, opened 2017-12-06
Reduce python wrapper overhead of CheckCollisionRays. Release GIL

woodychow:
Function runtime reduced by approximately 40%

The original code already assumed "rays" is a numpy array.

### DIFF
--- a/python/bindings/openravepy_global.cpp
+++ b/python/bindings/openravepy_global.cpp
@@ -191,19 +191,6 @@ public:
     }
 };
 
-class AutoPyArrayObjectDereferencer
-{
-public:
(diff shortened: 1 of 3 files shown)


rdiankov:
i like it. perhaps we can do a `memset` for ppos to do it only once rather than set to 0 every time?
also, is it safe to unlock GIL when we're still reading/writing memory owned by python?

woodychow:
Yes we can memset first. Let me do it.

Data pointer is extracted from the python array object before hand. It won't affect Python's reference count. It's the user's problem if he/she decides to modify the content of "rays" before the function returns.

rdiankov:
thanks!

PR #538, opened 2017-12-08
Add Environment#CopyKinBody/CopyRobot

cielavenir:
Backend support for UI (except link)

By the way, for openrave itself, do we need CopySensor()?


rdiankov:
I'm not sure if "Copy" is the correct name, since all you are doing is wrapping RaveClone. `CloneBodyByName` will be a more appropriate name, but I'm still not sure why you need this functionality in openrave?

cielavenir:
>functionality

well, itlplanner will also use this. Let me discuss with Woody first

PR #539, opened 2017-12-13
Use re-entrant qhull interface if available.

woodychow:
http://www.qhull.org/html/qh-code.htm
> New code should be written with libqhull_r. Existing users of libqhull should consider converting to libqhull_r. Although libqhull will be supported indefinitely, improvements may not be implemented. Reentrant qhull is 1-2% slower than non-reentrant qhull.


rdiankov:
thanks!

PR #540, opened 2017-12-16
Fix malformed if() check when collada is missing

PeterBowman:
The line

```cmake
if( ${COLLADA_DOM_VERSION} VERSION_LESS "2.5.0" )
```

would evaluate to

```cmake
if( VERSION_LESS "2.5.0" )
```

and throw an *Unknown arguments specified* CMake error (tested: CMake v3.9.2) if collada is not found. Also avoid showing a misleading status message.


rdiankov:
thanks!

PR #541, opened 2017-12-22
Manipulatorikgoalsampler freevalues

kanbouchou:
add optional argument ``freevalues`` so that hand can be moved to ik parameter with unique posture.


rdiankov:
thanks

PR #545, opened 2018-01-16
More planning

cielavenir:
(This is what you have done to my PC last week, and small tweak for nice torque graph)


rdiankov:
can you explain?

```
                    if (abs(rawvelocity) < 1.0/360) { 
```

thanks

cielavenir:
if abs(rawvelocity) is too low, there will be jitter about _back emf_ . So Kei and I agreed that we should not multiply 0.9 if the velocity is less than 1deg/sec.
\# the rawvelocity unit is rot/sec IIRC

rdiankov:
interesting, for now i'll accept, but we'll need better parameters for this.
thanks

PR #546, opened 2018-01-18
Fixed mp4 recording

cielavenir:
Fixing this:

```
[libx264 @ 0x561b4183b4e0] broken ffmpeg default settings detected
[libx264 @ 0x561b4183b4e0] use an encoding preset (e.g. -vpre medium)
[libx264 @ 0x561b4183b4e0] preset usage: -vpre <speed> -vpre <profile>
[libx264 @ 0x561b4183b4e0] speed presets are listed in x264 --help
[libx264 @ 0x561b4183b4e0] profile is optional; x264 defaults to high
```

(it does not change that we need qtcoin for video capture though)

### DIFF
--- a/plugins/logging/viewerrecorder.cpp
+++ b/plugins/logging/viewerrecorder.cpp
@@ -715,8 +715,12 @@ protected:
         AVCodecContext *codec_ctx;
         AVCodec *codec;
 
+        bool bFixH264 = false;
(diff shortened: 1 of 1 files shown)


cielavenir:
By the way, the output mp4 is not seekable (i.e. can be opened only with mplayer or ffplay). Did this apply also when we used debian wheezy?

rdiankov:
thanks!

Issue #519, opened 2017-08-11
cannot run openrave.py --example graspplanning


nag92:
I am trying to install openrave on 16.04. I followed the instructions here: https://scaron.info/teaching/installing-openrave-on-ubuntu-16.04.htm

when I try to run the example the window opens for a second then it crashes and I get the following error:

`osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/segwayrmp/segwayrmp.iv
osgDB::ReaderWriterIV::readNode() File /usr/local/share/openrave-0.9/models/segwayrmp/segwayrmp.iv loaded successfully.
osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/WAM/wam0.iv
osgDB::ReaderWriterIV::readNode() File /usr/local/share/openrave-0.9/models/WAM/wam0.iv loaded successfully.
osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/WAM/wam1.iv
osgDB::ReaderWriterIV::readNode() File /usr/local/share/openrave-0.9/models/WAM/wam1.iv loaded successfully.
osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/WAM/wam2.iv
osgDB::ReaderWriterIV::readNode() File /usr/local/share/openrave-0.9/models/WAM/wam2.iv loaded successfully.
osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/WAM/wam3.iv
osgDB::ReaderWriterIV::readNode() File /usr/local/share/openrave-0.9/models/WAM/wam3.iv loaded successfully.
osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/WAM/wam4.iv
osgDB::ReaderWriterIV::readNode() File /usr/local/share/openrave-0.9/models/WAM/wam4.iv loaded successfully.
osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/WAM/wam5.iv
osgDB::ReaderWriterIV::readNode() File /usr/local/share/openrave-0.9/models/WAM/wam5.iv loaded successfully.
osgDB::ReaderWriterIV::readNode() Reading file /usr/local/share/openrave-0.9/models/WAM/wam6.iv
[...]

PR #548, opened 2018-01-19
fix graspplanning.py for numpy 1.12

cielavenir:
#519


cielavenir:
Since we have upgraded numpy, we are affected by this (very few people will try this example though)

rdiankov:
thanks

PR #549, opened 2018-01-26
not set fpu state back if RaveGlobal is not even initialized

akovski:


rdiankov:
closing in favor of individually tracking intialization level of crlibm

PR #557, opened 2018-02-16
Verification

rmislam:
Adds iktype argument to allow specification of parameterization type. Needed for verification pattern pose planning when the pattern is not on the end effector but instead on another link before the flange.

Tested and has already been in use for a while for calibration verification.


rdiankov:
thanks!

PR #561, opened 2018-02-22
Add check of nEndTimeDiscretization in parabolicsmoother

cielavenir:
Or the square will overflow and vVisitedDiscretization.resize() will go wrong way.


rdiankov:
thanks

PR #566, opened 2018-03-12
Fix planning

Puttichai:
* In `DynamicsCollisionConstraint::Check` : added collision checking for configurations along the linear segment connecting configurations before and after calling `_neighstatefn`

* In RRT, now also adding intermediate configurations (returned from `CheckPathAllConstraints`) to the tree

* In linear shortcutting, now stopping subsampling a segment when the configuration returned by `_neighstatefn` deviates from the expected straight line segment.

* `_neighstatefn` now returns one of `NSS_X` defined in `planner.h` instead of a `bool`.


rdiankov:
yay~

PR #569, opened 2018-03-19
Added PyCollisionReport::Reset() and some utilities

cielavenir:


cielavenir:
teach 3877 might be related

cielavenir:
Antoine told me I should use `env.GetCollisionChecker().SetCollisionOptions(CollisionOptions.Contacts)`. Maybe Reset() is not required... But I'd like your opinion. Or I can make smaller MR containing 86c0a41 if asked.

rdiankov:
why do we need `coloptions` in constructor of CollisionReport?

rdiankov:
ping - why do we need `coloptions` in constructor?

PR #570, opened 2018-04-06
Fix memory leak in Python converter

woodychow:
For some reason PyObject* returned by PyNumber_Int and so won't get deleted even when Py_DECREF is called.

https://stackoverflow.com/questions/26595350/extracting-unsigned-char-from-array-of-numpy-uint8


rdiankov:
yay!

PR #571, opened 2018-04-06
PyObjects* returned by PyErr_Fetch have to be decref'd

woodychow:
From the official documentation:
Retrieve the error indicator into three variables whose addresses are passed. If the error indicator is not set, set all three variables to NULL. If it is set, it will be cleared and you own a reference to each object retrieved. The value and traceback object may be NULL even when the type object is not.


rdiankov:
thanks

PR #573, opened 2018-04-09
Convert sympy type to numpy type for axisAngleFromRotationMatrix

woodychow:
axisAngleFromRotationMatrix no longer accepts sympy type after the memory leak fix.

### DIFF
--- a/python/ikfast.py
+++ b/python/ikfast.py
@@ -1467,7 +1467,7 @@ class IKFastSolver(AutoReloader):
         """
         if axisAngleFromRotationMatrix is not None:
             Teval = T.evalf()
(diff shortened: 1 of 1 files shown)


cielavenir:
@rdiankov Please decide which is better:

```
axisangle = axisAngleFromRotationMatrix(numpy.array(T.evalf()[:3, :3].tolist(), dtype=numpy.float64))

axisangle = axisAngleFromRotationMatrix(numpy.array(T.evalf()[:3, :3]).astype(numpy.float64))
```

rdiankov:
both ways are sub-optimal. @woodychow for now, can you just allow casting of sympy types until we can narrow down the memory leak issue? i don't think the memory leaks due to sympy objects will be blowing up the system.
thanks

woodychow:
https://github.com/rdiankov/openrave/pull/574

PR #574, opened 2018-04-09
Python converter that works with sympy types

woodychow:
The memory leak is caused by PyNumber_Int, which is deprecated in Python 3. PyNumber_Float and PyNumber_Long are good. They are just like `float(x)` or `long(x)` in Python.


rdiankov:
thanks~

PR #576, opened 2018-04-18
Updated how to holf checkpreemptfn in ikfast, eliminating weakref

cielavenir:
In this way we can remove weakref.proxy(self).func (and we don't have to put weakrefmethod as an openrave dependency)

Already reviewed by @woodychow .


rdiankov:
please don't have checkpreemptfn initialized in the parameter list like this, it looks weird and bad..

rdiankov:
ping

PR #582, opened 2018-05-08
Add jerk limit for collada files

snozawa:
This is update to support jerk limitation in collada files.
I have several questions and discussions:

1. I update colladareader/writer and drop openrave xml. Is this OK?
2. For jerk values in `kinbodyjoint.cpp`, i set negligibly large jerk by default, which can change acceleration between min and max within a typical time step.
I set `1e5`, obtained from `(50-(-50))/0.001` where 50, -50, and 0.001 are max accel, min accel, and typical time step, respectively.
3. I found several naming depending on acceleration and velocity: `Prop_JointAccelerationVelocityTorqueLimits` and `Save_JointMaxVelocityAndAcceleration`.
Do i need to keep these naming for backward compatibility?
Currently, I do not change these like `Prop_JointAccelerationVelocityTorqueJerkLimits`.
4. I'm not sure that the intended calculation in the following `RobotBase::GetActiveDOFAccelerationLimits`: https://github.com/rdiankov/openrave/blob/production/src/libopenrave/robot.cpp#L959-L974.
Currently, I added `RobotBase::GetActiveDOFJerkLimits` without these computations.


rdiankov:
thanks. i'd like to change the minor version and some comments, but overall, no problem~

PR #583, opened 2018-05-10
Migrate to Qt5 and remove osgQt

braineo:
- Migrated to Qt5 library
- Add high DPI support
- Remove osgQt


sirop:
Does the camera plugin work as it should with the new QtViewer?

rdiankov:
yay!

PR #591, opened 2018-06-22
Error visualization translation

cielavenir:
Well, we could have choose two workflows.
One is using this modification. The other is updating PO files on production branch and merging it to errorVisualization **first**.
Actually I love the latter, but Rosen is very busy recently.....


cielavenir:
@kanbouchou I checked your planningcommon update. Perhaps you might want to update some of these translations.....

rdiankov:
thanks!

PR #592, opened 2018-06-23
compile with Clang

strm10:
- compliance to the more strict type checking
- proper compiler test for Clang on Ubuntu/Debian (`CMAKE_CXX_COMPILER` can be just a symbolic link like `/usr/bin/c++`)


rdiankov:
thanks~

PR #593, opened 2018-07-02
fixed boost_assert_failed new build

cielavenir:
already-installed-environment were not affected.

----

already-installed-envioronment basically mean "dev machines".
we have seen that almost all teamcity builds are failing.


rdiankov:
thanks!

PR #595, opened 2018-07-09
Make cdpy_exception and pyann_exception visible

woodychow:
Also, add exception handling to the python wrapper


rdiankov:
thanks!

PR #600, opened 2018-08-09
take rotor_inertia into consideration when computing torque

kanbouchou:
rotor_inertia in ``ElectricMotorActuatorInfo`` was not used to compute joint torque. 

In reality, accelerating motor rotor itself requires motor torque.  Previously, ``ComputeInverseDynamics`` would return smaller torque which doesn't take this acceleration component into account and thus planned trajectory sometimes could not be followed due to motor torque saturation.


rdiankov:
thanks

PR #604, opened 2018-09-04
Fix memory leak of DAE read

woodychow:
To reproduce the leak without this fix

```py
import openravepy
env = openravepy.Environment(0)
for i in range(10000):
    env.ReadKinBodyURI('test.mujin.dae')
```

### DIFF
--- a/src/libopenrave-core/colladaparser/colladareader.cpp
+++ b/src/libopenrave-core/colladaparser/colladareader.cpp
@@ -17,6 +17,7 @@
 #include "colladacommon.h"
 #include <boost/algorithm/string.hpp>
 #include <openrave/xmlreaders.h>
+#include <libxml/xmlversion.h>
 
 namespace OpenRAVE
 {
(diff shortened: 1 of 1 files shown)


rdiankov:
hmm... how much does this affect performance re-initializing collada when trying to read/write?

rdiankov:
By the way, I wasn't releasing collada resources on purposes because of this.

woodychow:
Let me do a quick test

woodychow:
```
In [4]: %timeit env.ReadKinBodyURI('./test.mujin.dae')
10 loops, best of 3: 25.8 ms per loop

In [3]: %timeit env.ReadKinBodyURI('./test.mujin.dae')
10 loops, best of 3: 28.1 ms per loop
```

woodychow:
With https://github.com/rdiankov/collada-dom/pull/31/, loading the same DAE file takes 10-12ms on my computer.

woodychow:
Thanks!

PR #605, opened 2018-09-08
fix update linkinfo as always update geometryInfos

yoshikikanemoto:
change Link::UpdateInfo always update GeometryInfo in case of that on…ly content of geometries is changed


rdiankov:
thanks!

PR #608, opened 2018-10-01
hot fix to prevent NoneType error when trying to evaluate atan(0, eva…

kanunikov-denys:
this is a hotfix to prvent NoneType error raised. sympy version 0.7.1 throws an error when:

atan(0, evaluate=False).evalf()

called. The sympy bug is fixed in 0.7.2 version


rdiankov:
great catch

PR #611, opened 2018-10-18
 FCL_LDFLAGS is a list. A list doesn't need double quotes 

woodychow:
Otherwise, ninja build will fail.


rdiankov:
thanks

rsinnet:
This is not the correct usage of `set_target_properties`. These properties are scalars as shown in the docs: https://cmake.org/cmake/help/latest/command/set_target_properties.html

woodychow:
Perhaps this is the proper fix?
```cmake
    string(REPLACE ";" " " FCL_LDFLAGS_STR "${FCL_LDFLAGS}")
    set_target_properties(fclrave PROPERTIES COMPILE_FLAGS "${PLUGIN_COMPILE_FLAGS} ${FCL_CFLAGS_OTHER} -std=c++11" LINK_FLAGS "${PLUGIN_LINK_FLAGS} ${FCL_LDFLAGS_STR}")
```
The reason why I made the change was

My ${FCL_LDFLAGS} was `-L/home/mypath;-lfcl`

which came from fcl.pc.in (`Libs: -L${libdir} -lfcl`)

The generated command from this was

```
/usr/bin/c++ ....... -L/home/mypath;-lfcl
```

And c++ gives `c++: fatal error: no input files`

https://stackoverflow.com/questions/11594905/how-do-i-correctly-pass-cmake-list-semicolon-sep-of-flags-to-set-target-proper

rsinnet:
That solution should work fine. I don't really know of a better solution. The issue is that `PLUGIN_LINK_FLAGS` seems to be a space separated string and `FCL_LDFLAGS` is a list. Unfortunately what you've written is the best way to convert AFAIK :sweat_smile:

woodychow:
Thanks.

I wish there is CI to catch this kind of issues...

rsinnet:
Ahhh no CI, that explains it! Lots of platforms offer free CI services for open source repos. Would be fairly easy to set up a Docker image to build this. Even if no tests are run, would still be a good sanity check.

PR #612, opened 2018-10-24
sort roots for stability of solution indices

kanbouchou:
sorting roots so that hopefully solution indices branching is more robust to small changes in coefficients.

btw, std::less didn't have specialization for complex so I had to implement it myself.


rdiankov:
thanks!

PR #616, opened 2018-11-02
 set real joint values to free joints when testing an IK solver

tgn3000:
- put the class definition of IkFastSolver<T> into a separate header
- include the header in ikfastsolver.cpp, ikfastmodule.cpp
- add IkFastSolver<T>::GetIkFunctions() to allow access _ikfunctions
- add an extra test that sets real joint values to free joints before testing with DebugIK; keep the original test
- simple refactoring in DebugIK
- can pass "--freeindices" on command line to "openrave.py --database inversekinematics"
- print out five percentages from DebugIK test in testik from inversekinematics.py


plugins/ikfastsolvers/ikfastmodule.cpp
```diff
@@ -995,21 +996,31 @@ class IkFastModule : public ModuleBase
         RAVELOG_DEBUG("Starting DebugIK... max iterations=%d\n", num_itrs);
         RobotBase::RobotStateSaver saver(robot);
         robot->Enable(bTestSelfCollision);
+
+        /* ========== Beginning of requried modifications in DebugIK ========== */
         RobotBase::ManipulatorPtr pmanip = robot->GetActiveManipulator();
-        if( !pmanip->GetIkSolver() ) {
+        IkSolverBasePtr piksolverbase = pmanip->GetIkSolver();
+        IkFastSolverPtr<dReal> piksolver = boost::dynamic_pointer_cast<IkFastSolver<dReal>>(piksolverbase);
+        boost::shared_ptr<ikfast::IkFastFunctions<dReal> > ikfunctions = piksolver->GetIkFunctions();
```
woodychow: piksolver can be null. You probably want to move the nullptr check up.

plugins/ikfastsolvers/ikfastmodule.cpp
```diff
@@ -1145,29 +1158,40 @@ class IkFastModule : public ModuleBase
...
+                }
+                vfreeparameters_out.resize(vfreeparameters.size());
+                
                 // have to start at a random config
                 while(1) {
-                    for(int j = 0; j < (int)vrand.size(); j++) {
+                    for(int j = 0; j < (int) narminds; j++) {
                         vrand[j] = vlowerlimit[j] + (vupperlimit[j]-vlowerlimit[j])*RaveRandomFloat();
                     }
+                    for(int k = 0; k < (int) nfreeparams; k++) {
```
rdiankov: whole point of the random test is to not rely on the randomly chosen freevalue...

tgn3000: add an extra test that computes with free joint values known; keep the original test that computes with free joint values unknown

tgn3000: can pass "--freeindices" on command line to "openrave.py --database inversekinematics"; print out five percentages from DebugIK test in testik from inversekinematics.py

PR #617, opened 2018-11-13
Add libopenrave-core to the list of libraries to link against in pkg config info

woodychow:


woodychow:
Nevermind. Didn't realize there is openrave-core.pc.in

PR #620, opened 2018-11-23
Fix CMake link warning

woodychow:
Fixes this warning
```
CMake Warning (dev) at plugins/rplanners/ParabolicPathSmooth/CMakeLists.txt:6 (target_link_libraries):
  Policy CMP0023 is not set: Plain and keyword target_link_libraries
  signatures cannot be mixed.  Run "cmake --help-policy CMP0023" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  The plain signature for target_link_libraries has already been used with
  the target "ParabolicPathSmooth".  All uses of target_link_libraries with a
  target should be either all-keyword or all-plain.

  The uses of the plain signature are here:

   * plugins/rplanners/ParabolicPathSmooth/CMakeLists.txt:5 (target_link_libraries)
```


woodychow:
The warning is caused by making 2 calls of `target_link_libraries` with the same target. It has nothing to do with `PUBLIC`

I am not sure if it is necessary to add `PUBLIC` to all the existing `target_link_libraries`

rsinnet:
@woodychow, which linker warnings are you thinking of? Maybe we're talking about something different. The problem I was seeing and what seems to be described in this ticket is `CMP0023`: https://cmake.org/cmake/help/v3.0/policy/CMP0023.html

I'm fairly certain this warning is caused by mixing keyword signatures and non‒keyword signatures. I've validated that it's fine to have multiple `target_link_libraries` calls to the same target. It just appends to the list of libraries. You can also see this in the source code for CMake. During the processing of the `target_link_libraries` command, the parse checks what type of signature the `target_link_libraries` command has, either`cmTarget::PlainTLLSignature` or new-style `cmTarget::KeywordTLLSignature`; see here: https://gitlab.kitware.com/cmake/cmake/blob/master/Source/cmTargetLinkLibrariesCommand.cxx#L317.

Then the command is pushed to an `std::vector` on the `cmTarget` which validates that the signature is the same as the previous command pushed (https://gitlab.kitware.com/cmake/cmake/blob/master/Source/cmTarget.cxx#L730). If not, it prints `CMP023` depending on warning settings. 

Based on my reading of the above code, it should be perfectly fine to have multiple `target_link_library` calls with the same target, otherwise they wouldn't need an `std::vector` to store all the commands.

Should be easy enough to test this.

PR #622, opened 2018-12-05
Return PS_Failed on inconsistent constraints in waypoints

strm10:
Currently `TrajectoryRetimer2::PlanPath()` throws an exception on a constraint violation in waypoints. But it's more straightforward to return an error code like for other kinds of unexpected inputs.


rdiankov:
thanks

PR #623, opened 2018-12-06
fix invalid variable access

yoshikikanemoto:
this caused exception for some case.


rdiankov:
woot! congratulations! after 8 months you found one.. good job~

PR #624, opened 2018-12-10
Add hard limit saving functionality for state saver

snozawa:
Add hard limit saving functionality for state saver.


rdiankov:
why?

snozawa:
The hard limits already have setter functions (`SetDOFHardVelocity`..etc).
I just thought that it would be better to have state saver for them if we have setter functions.

Of course, usually hard limits should not be changed so much in the code.
but for now, i change hard limits for testing and i'd like to have state sever for them.

PR #626, opened 2018-12-17
Add subject for env

braineo:


rdiankov:
not going to work. you don't save subject anywhere in openrave, so once openrave re-saves the environment, you lose the data.

rdiankov:
for now will merge since same as keywords, but should have SetDescription/GetDescription in environment property.

PR #627, opened 2018-12-21
fixed how configs are added to tree backward

Puttichai:
- In RRT, for tree backward, need to add configs from `filterreturn` into the tree in reverse order.
- In dynamicscollisionconstraints's check function, if `!bCheckEnd`, then not add `q1` to `filterreturn`


rdiankov:
woot!

PR #628, opened 2018-12-24
Add missing header for fixed-size 64-bit int type

PeterBowman:
Alternatives: `<cstdint>` (since C++11), `<boost/cstdint.hpp>`.

* https://stackoverflow.com/q/44574682
* https://stackoverflow.com/a/2709127

Addresses a *‘uint64_t’ was not declared in this scope* error ([Travis](https://travis-ci.org/roboticslab-uc3m/asibot-main/jobs/471729244#L4570)).


rdiankov:
thanks!

PR #629, opened 2018-12-24
Add if-guard for newly added JSON command

PeterBowman:
Link step crashes if rapidjson is not found. This is a follow-up of https://github.com/rdiankov/openrave/commit/2bcc051254e2a76340db39bad56839fd77139a3f.


rdiankov:
thanks!

PR #630, opened 2018-12-24
Add missing header, fix unresolved 'slice'

PeterBowman:
Getting a *‘slice’ has not been declared* error since https://github.com/rdiankov/openrave/commit/849abeace83aa91d450c3d51f9c2bbe53b238c5b (Ubuntu Xenial). Docs: https://www.boost.org/doc/libs/1_34_0/libs/python/doc/v2/slice.html.


rdiankov:
thanks!

Issue #631, opened 2018-12-24
Unable to compile rplanners plugin (OR v0.15, Ubuntu Xenial)


PeterBowman:
```
openrave/plugins/rplanners/linearsmoother.cpp:1211:60: error: no matching function for call to ‘std::__cxx11::list<std::vector<double> >::erase(std::__cxx11::list<std::vector<double> >::const_iterator&, std::__cxx11::list<std::vector<double> >::iterator)’
             listpath.erase(startInfo.itnode, listpath.end());
                                                            ^
```

(committed at https://github.com/rdiankov/openrave/commit/dc1b5600fb58ac29dda1eb5bed87e5d3a150ce87)

I followed [these install steps](https://github.com/roboticslab-uc3m/installation-guides/blob/dd2938dbe0b4b05240945062afcb04fe208fead5/install-openrave.md#install-openrave-most-modern-ubuntu-such-as-1404-or-1604).

Compiled with `gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.11)`. Looks like C++11 might solve this issue as non-const `iterator`s (such as `listpath.end()`) can be implicitly cast to const ones, plus the signature of `::erase` was conveniently changed ([cppreference](https://en.cppreference.com/w/cpp/container/list/erase)).

This was originally reported by @crtassos at https://github.com/rdiankov/openrave/issues/458#issuecomment-425746513.

PeterBowman:
Done at https://github.com/rdiankov/openrave/pull/632.

PR #632, opened 2018-12-28
Consume SampleInfo::itnode as a non-const iterator

PeterBowman:
Solves #631. All this boils down to avoiding a compile error in:

https://github.com/rdiankov/openrave/blob/737c4b1bbeb4d32f8aef1bf3a027eedb93854c26/plugins/rplanners/linearsmoother.cpp#L1211

Here, [`std::list::erase`](https://en.cppreference.com/w/cpp/container/list/erase) accepts a `const_iterator` and an `iterator` as first and second arguments, respectively. In C++11, this code would be OK due to an implicit cast from `iterator` to `const_iterator`. In this patch, `SampleInfo::itnode` is turned into a non-const `iterator`; some function internals and signatures have been updated accordingly.

### DIFF
--- a/plugins/rplanners/linearsmoother.cpp
+++ b/plugins/rplanners/linearsmoother.cpp
@@ -769,21 +769,21 @@ protected:
     {
         SampleInfo() : fabsnodedist(0), fdeltadist(0), inode(0) {
         }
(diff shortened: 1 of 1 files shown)


PeterBowman:
See also: [Scott Meyers, *Three Guidelines for Effective Iterator Usage*](http://www.aristeia.com/Papers/CUJ_June_2001.pdf), Guideline 1: "Prefer `iterator` to `const_iterator`, `reverse_iterator`, and `const_reverse_iterator`".

rdiankov:
sure, thanks!

PR #634, opened 2019-01-06
Fix: explicit cast from dReal to IkReal is needed in list initialisation in C++11

strm10:
With the default CMake parameter values (`OPT_IKFAST_FLOAT32=ON`), the actual type of `dReal` is `double` and that of `IkReal` is `float`. Since implicit narrowing conversion for initialiser list is prohibited in C++11 (https://en.cppreference.com/w/cpp/language/list_initialization), the following code can cause an error in some compilers (GCC 7.3.0 for example).

```
TransformMatrix t = param.GetTransform6D();
IkReal eetrans[3] = {t.trans.x, t.trans.y, t.trans.z};  // type of t.trans.X is dReal
```

```
error: non-constant-expression cannot be narrowed from type 'double' to 'float' in initializer list [-Wc++11-narrowing]
```

This patch adds explicit casts for initialiser lists.

### DIFF
--- a/plugins/ikfastsolvers/ikfastsolver.cpp
+++ b/plugins/ikfastsolvers/ikfastsolver.cpp
@@ -925,8 +925,8 @@ protected:
                 if( _bEmptyTransform6D ) {
                     t = t * tLocalTool.inverse();
                 }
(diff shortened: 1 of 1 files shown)


rdiankov:
thanks!

PR #642, opened 2019-01-24
Fix smoother ramp initialization

Puttichai:
- added checking so that we don't continue after an interpolation failure during ramp initialization
- added a vel mult cut off so that we stop if the current vel mult goes below this cut off during ramp initialization


rdiankov:
thanks!

PR #643, opened 2019-01-24
Parabolicsmoother2

Puttichai:
- removed forcing parabolicsmoother2
- changed the default smoother name to `parabolicsmoother2`


rdiankov:
let's keep parabolicsmoother as default

PR #647, opened 2019-02-07
Fix acceleration checking

Puttichai:
- Assigned `_vtempaccelconfig` to `(dq1 - dq0)/timeelapsed` when having the option `IT_AllLinear`
- Modified `VerifyTrajectory` to support the new all-linear interpolation


rdiankov:
thanks!

PR #648, opened 2019-02-13
added ik translational/rotational distance computation

Puttichai:
added functions for computing translational and rotational distances between IKs


rdiankov:
closing since added

PR #651, opened 2019-02-26
Test for lib ends with .so, because debug symbols are at .so.debug.

ziyan:
Currently, openrave finds any file that contains `.so` anywhere in the filename as loadable plugin, so things like:

- `libxxx.so~`
- `libxxx.so.debug`
- `libxxx.something`

will all be treated as plugin library.

This change makes it check for file that ends with `.so`. This change makes the logic same as that on windows.


rdiankov:
thanks~

PR #652, opened 2019-03-05
disabled validation & used correct interpolation equation

Puttichai:
When derivgroup's interpolation is not the expected one compared to group's interpolation (for example, when positions and velocities are both linear), set `derivoffset < 0` to disable validate and use correct interpolation.


rdiankov:
thanks!

PR #666, opened 2019-04-01
preload readline

cielavenir:
On rundocker.py envioronment, "ipython with viewer" will crash without this patch.

### DIFF
--- a/python/openrave.py.in
+++ b/python/openrave.py.in
@@ -18,6 +18,7 @@ __copyright__ = 'Copyright (C) 2009-2010 Rosen Diankov (rosen.diankov@gmail.com)
 __license__ = 'Apache License, Version 2.0'
 
 import sys,time
+import IPython.utils.rlineimpl # preload readline
 from types import ModuleType
 from numpy import *
 from OpenGL import GL


cielavenir:
(I'm expecting Kou to join also as reviewer)

cielavenir:
workaround: `function rave () { DISPLAY=:0 python -m IPython -c "import os,sys;sys.argv=sys.argv[2:];rave=open('/usr/local/bin/openrave.py').read();exec(rave);exit()" "$@"; }`

cielavenir:
I just found that this IPython matter is affected IPython<=4.2.1. If we move to IPython 5.x, this does not matter anymore.

rdiankov:
great~ i can accept this patch if you comment out the line and document that it is necessary for ipython <= 4.2.1. 
thanks~

cielavenir:
maybe I did not understand correctly - "commenting out" will require uncommenting to enable the feature. namely modifying code. It the same as adding the code after every satrtup.

----

Now I change the method. I removed preloading and moved importing IPython before creating viewer. Is this ok?

PR #671, opened 2019-04-16
Remove RAPIDJSON_HAS_STDSTRING

ziyan:
Because some projects import rapidjson before openrave and do not have RAPIDJSON_HAS_STDSTRING set.


rdiankov:
sure... thanks

PR #672, opened 2019-04-17
Hotfix compute inner empty volume container

kanunikov-denys:
hot fix for GT_Container copmutation of inner empty volume.
Also be noted that `KinBody::GeometryInfo::ComputeInnerEmptyVolume` return transformation in the parent link coordinate.


rdiankov:
thanks!

PR #673, opened 2019-04-19
Added a function for setting _neighstatefn for jitterer

Puttichai:
- Added a function for setting `_neighstatefn` for a jitterer
- When there are constraints, 
  - need to project the sampled configuration onto the constraint manifold using `_neighstatefn` before checking with perturbation
  - when checking with zero perturbation, should check if `_neighstatefn` returns `NSS_Reached`


rdiankov:
looks good, thanks!

PR #675, opened 2019-04-23
Update hard limit usage

snozawa:
- Set 0 as default hard limits
- In case of reading from collada file : if `hardMaxXX` tag is not written for a joint, do not set the hard limit value, and thus, skip soft limit hard limit consistency check.
- In case of writing to collada file : if the hard limit is not set (e.g. 0), do not write the `hardMaxXX` tags into collada file.

This PR is to clarify that the user has not set the hard limit yet.


rdiankov:
looks good, i think this is correct approach too. thanks!

PR #677, opened 2019-05-06
Increment startindex in GetTrajectorySegment function.

snozawa:
Increment startindex.
later on, we use startindex. but, if we 'Remove' the first point, corresponding 'startindex' should be incremented.

### DIFF
--- a/src/libopenrave/planningutils.cpp
+++ b/src/libopenrave/planningutils.cpp
@@ -1873,6 +1873,7 @@ TrajectoryBasePtr GetTrajectorySegment(TrajectoryBaseConstPtr traj, dReal startt
         else {
             // time between first and second point is non-existent, so remove first point
             outtraj->Remove(0, 1);
+            startindex++; // later on, we use startindex. but, if we 'Remove' the first point, corresponding 'startindex' should be incremented.
             outtraj->GetWaypoint(0, values);
             spec.InsertDeltaTime(values.begin(), 0);
             outtraj->Insert(0,values, true);
--- a/src/libopenrave/planningutils.cpp
+++ b/src/libopenrave/planningutils.cpp
@@ -1883,7 +1883,7 @@ TrajectoryBasePtr GetTrajectorySegment(TrajectoryBaseConstPtr traj, dReal startt
     if( endtime < traj->GetDuration() ) {
         // have to change the last endpoint, should sample from outtraj instead since both start and end can be within same waypoint range
         outtraj->Sample(values, endtime-starttime);
-        outtraj->Insert(endindex-startindex,values,true);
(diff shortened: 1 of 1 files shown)


snozawa:
- Problem
   if we use good-cut value for `starttime` such as multiple of deltatime, it  does not work.
    ```
   openrave_exception: openrave (Assert): [OpenRAVE::TrajectoryBasePtr OpenRAVE::planningutils::GetTrajectorySegment(OpenRAVE::TrajectoryBaseConstPtr, OpenRAVE::dReal, OpenRAVE::dReal):1890] RaveFabs(outtraj->GetDuration() - (endtime - starttime)) <= g_fEpsilonLinear*traj->GetNumWaypoints(), (eval 0.1 <= 1.58114e-13) 
    ```

- How to solve
    in the if sentence if we use good-cut value, we use `Remove` and it seems inconsistent with later computation.
    Just increment the `startindex`.


- test code (without this PR, this test code fails)
    ```
    import openravepy
    newtraj = openravepy.RaveCreateTrajectory(robot.GetEnv(),'')
    posspec = robot.GetActiveConfigurationSpecification('linear')
    posspec.AddDeltaTimeGroup()
    newtraj.Init(posspec)
    dt = 0.1
    newtraj.Insert(0, numpy.hstack([[0]*robot.GetDOF() + [dt]]))
    newtraj.Insert(0, numpy.hstack([[0]*robot.GetDOF() + [dt]]))
    newtraj.Insert(0, numpy.hstack([[0]*robot.GetDOF() + [dt]]))
    newtraj.Insert(0, numpy.hstack([[0]*robot.GetDOF() + [dt]]))
    newtraj.Insert(0, numpy.hstack([[0]*robot.GetDOF() + [0]]))
    duration = newtraj.GetDuration()

    # Call GetTrajectorySegment for different start/end times
    eps = 1e-9
    startTimeList = [0, eps, dt*0.5, dt-eps, dt+eps, dt*0.5+dt, 2*dt-eps, 2*dt, 2*dt+eps]
    endTimeList = [duration, duration-eps, duration-dt]
    for startTime in startTimeList:
        for endTime in endTimeList:
            log.warn('startTime=%r, endTime=%r', startTime, endTime)
            openravepy.planningutils.GetTrajectorySegment(newtraj, startTime, endTime)
    ```

rdiankov:
wow, yes, that was pretty silly mistake.
although your fix is correct, i'm wondering whether we should do the following instead:
```c++
        outtraj->Sample(values, endtime-starttime);
        outtraj->Insert(outtraj->GetNumWaypoints()-1,values,true); // assume last end point
```

what do you think?

snozawa:
oh, i see.
it is much more understandable!
i reflected your comment to this PR.
i also confirmed that the above test code passed with the new PR.

rdiankov:
thanks!

PR #679, opened 2019-05-20
Parabolicsmoother2

Puttichai:
`parabolicsmoother.cpp`
- added more informative logging on timing
- better logic for creating `vVisitedDiscretization`
- other small fixes

`parabolicsmoother2.cpp`
- removed separate dof velocity/acceleration scaling factor as it could deteriorate smoother performance.
- removed early manip constraints checking in `SegmentFeasible2`
- better logic for creating `vVisitedDiscretization`
- improved logic in `_RecomputeNDTrajectoryFixedDuration` so that we always try doing interpolation with the given duration first before computing a new duration considering inoperative time interval.
- normalized parabolicsmoother2 behavior so that it is identical to parabolicsmoother (when not turning on the merging function; except some differences in interpolation method which parabolicsmoother does not return the time-optimal velocity profile)
- other small fixes


rdiankov:
thanks!

PR #680, opened 2019-05-24
handle case that all spheres are zero radius better

kanbouchou:
there was a robot model whose totalweight was 0 and caused division by zero.

I initially thought of checking small ``jointspheres[ijoint][1]`` individually and  set to some non-zero value, but later found existing check to set to ``0.02`` so, I added code to avoid zero division and let the existing code handle small value of ``weights``


rdiankov:
thanks!

PR #683, opened 2019-05-31
cahnge inner region computation according to the wall and vGeomData2

kanunikov-denys:
make inner region computation respect froced inner region


rdiankov:
thanks!

PR #686, opened 2019-07-02
Deep copy member variables of JointInfo and LinkInfo

strmio:
Copy operators for `JointInfo` and `LinkInfo` make a copy of shared pointers to a same instance for some member variables like `JointInfo::_vmimic` or `LinkiInfo::_vgeometryinfos` instead of making a copy of the contents. As a result a modification to an instance of these classes affects a copied instance also.

For example `RobotBase::_ComputeConnectedBodiesInformation()` makes a copy of `JointInfo` to add a body name to joint names as a prefix, and modifies the joint names in the original mimic equations unintentionally. Which results in the following error.

```
In [1]: env.Remove(robot)
Out[1]: True

In [2]: env.Add(robot)
2019-07-01 10:48:45,176 openrave [ERROR] [kinbody.cpp:3493 KinBody::_ComputeInternalInformation] failed to set mimic equations on kinematics body Robot: openrave (InvalidArguments): [virtual void OpenRAVE::KinBody::Joint::SetMimicEquations(int, const string&, const string&, const string&):1441] (itvar->find("joint") == 0) failed equation '(0+gripper2_gripper2_slider)' uses unknown variable                                                                                                                                             

In [3]: env.Remove(robot)
Out[3]: True

In [4]: env.Add(robot)                               
2019-07-01 10:50:27,346 openrave [ERROR] [kinbody.cpp:3493 KinBody::_ComputeInternalInformation] failed to set mimic equations on kinematics body Robot: openrave (InvalidArguments): [virtual void OpenRAVE::KinBody::Joint::SetMimicEquations(int, const string&, const string&, const string&):1441] (itvar->find("joint") == 0) failed equation '(0+gripper2_gripper2_gripper2_slider)' uses unknown variable                                                                                                                                    
```


rdiankov:
does this change the ABI?

strmio:
According to https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html , adding an explicit constructor breaks binary compatibility. Let me bump the minor version.

rdiankov:
thanks!

PR #687, opened 2019-07-03
RaveAddCallbackForDestroy was getting called once per dae read/write.

ziyan:
Might be accumulating function pointers in _listDestroyCallbacks.

### DIFF
--- a/src/libopenrave-core/colladaparser/colladacommon.cpp
+++ b/src/libopenrave-core/colladaparser/colladacommon.cpp
@@ -23,11 +23,27 @@ namespace OpenRAVE
 {
 static boost::shared_ptr<DAE> s_dae;
 static boost::mutex s_daemutex;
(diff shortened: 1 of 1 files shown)


rdiankov:
Hmm... whole point of having a RaveGlobal runtime is to remove dependency on static variables like this. You should have a way of confirming whether the current RaveGlobal supports this, perhaps have an id for the add/remove callback

rdiankov:
i wonder if inside `AddCallbackForDestroy`, we can check if the function pointers are equal before adding it to `_listDestroyCallbacks`

ziyan:
Unfortunately, comparing boost functions is not supported:

https://www.boost.org/doc/libs/1_70_0/doc/html/function/faq.html

rdiankov:
ok, let's go with your idea of static bool. i added a different reset fn with locking to make it even more safe.
thanks~

ziyan:
Oops, forgot to initialize `s_daedestroycallback` to `false`

PR #688, opened 2019-07-10
avoid error on unpickling LinkInfo

yoshikikanemoto:
When we try to unpickle data picked with older openrave, we will have `tuple: index out of range` error. vision's H5 file is using pickle to save kinbody information.


rdiankov:
thanks!

PR #691, opened 2019-08-14
avoid bad_weak_ptr exception while accessing destroyed grabbed kinbody

yoshikikanemoto:
makes it keep continuing and gives warning message


rdiankov:
thanks~

PR #693, opened 2019-08-22
rosen bug fix non-unique external reference id

kanbouchou:
fix for not saving externally referenced object correctly


rdiankov:
thanks!

PR #694, opened 2019-08-24
use aabb checks for self collision checking

yoshikikanemoto:
It looks we don't check AABB intersections only on self-collsion checking. Is it on purpose?


rdiankov:
woot! great catch~

PR #695, opened 2019-09-07
bugfix computing left offset

kanbouchou:
bug fix for cloning kinbody when original kinbody's dof value is non-zero


rdiankov:
thanks!

PR #698, opened 2019-09-30
replace push_back(make_pair(A, B)) by emplace_back(A, B)

tgn3000:
- Save implicit `std::move` and help readability


rdiankov:
thanks! i'm not very big on minor changes like this ;0)

PR #700, opened 2019-10-11
ConnectedBody related APIs

ziyan:
Add connected body related APIs for planning.

- fix cloning robot that has connected bodies
- RobotStateSaver to save connected body active state
- publish connected body active state
- preserve active manip of robot when it is removed and re-added to env
- fix configuration jitter caching of link pointers

cc @strmio


rdiankov:
thanks

PR #701, opened 2019-10-16
Update _pDummyJointCache properly on cloning of a robot

strmio:
Without this patch a cloned robot's `_pDummyJointCache` points to a link of the original robot instance, which leads to taking over the link from the original robot on `_ComputeConnectedBodiesInformation()` call.

cc @ziyan


ziyan:
LGTM.

rdiankov:
thanks!

Issue #327, opened 2014-11-14
Build error: Undefined symbols for architecture x86_64


topskychen:
I am building openrave of latest_stable version on Mac OS Yosemite and met the following build error:

```
Scanning dependencies of target check_libm_accuracy
[ 14%] Building CXX object src/CMakeFiles/check_libm_accuracy.dir/check_libm_accuracy_main.cpp.o
Linking CXX executable check_libm_accuracy
Undefined symbols for architecture x86_64:
  "__FPU_SETCW", referenced from:
      _crlibm_exit in libcrlibm.a(crlibm_private.c.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[3]: *** [src/check_libm_accuracy] Error 1
make[2]: *** [src/CMakeFiles/check_libm_accuracy.dir/all] Error 2
make[1]: *** [all] Error 2
make: *** [all] Error 2
```

asadaku:
I'm stuck at this exact same point. Is there some advice on how to get past this?

cielavenir:
well, could you test https://github.com/cielavenir/crlibm/commit/974e2ead89604d8f72d8ef51dca9a4766374a436 ?

asadaku:
Hello! I assume it must be compiled using cmake? So I created a build directory and used cmake and got syntax warning and errors about missing exp_accurate.h?

```
>> cd build/
>> cmake ../
CMake Warning (dev) at CMakeLists.txt:2:
  Syntax Warning in cmake code at column 113

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.

Trying to build crlibm on a x86_64 under Darwin, compiler is /Library/Developer/CommandLineTools/usr/bin/cc
CMake Warning (dev) in CMakeLists.txt:
  No cmake_minimum_required command is present.  A line of code such as

    cmake_minimum_required(VERSION 3.13)

  should be added at the top of the file.  The version specified may be lower
  if you wish to support older CMake versions for this project.  For more
  information run "cmake --help-policy CMP0000".
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring done
CMake Error at CMakeLists.txt:4 (ADD_LIBRARY):
  Cannot find source file:

    exp_accurate.h

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
[...]

PR #702, opened 2019-10-19
fixed crlibm_exit compilation without fpu

cielavenir:
We should use exactly the same condition in both crlibm_init and crlibm_exit. macOS does not have fpu_control.h.


cielavenir:
#327 

I don't know why I did not make MR at that time, sorry

rdiankov:
thanks!

PR #705, opened 2019-10-20
partially reverted 57a832ba44fe20f2d8824b6a7207bfd98d077545

cielavenir:
daeMetaAttribute::getName returns daeStringRef, which needs to be explicitly converted to daeString(`const char*`).
(I reverted this time, but If I should cast to daeString rather than std::string, please tell me)

I checked the compilation changed from red from green using Debian jessie and Ubuntu xenial.


rdiankov:
sure.. thanks

PR #640, opened 2019-01-24
Replace semicollons in FCL_LDFLAGS with spaces

woodychow:
This fixes https://github.com/rdiankov/openrave/issues/633

Regression was introduced by
https://github.com/rdiankov/openrave/pull/611

Issue #633, opened 2019-01-01
CMake fails on fclrave (Ubuntu 16.04)


haudren:
I'm running Ubuntu 16.04 and updated OpenRAVE, it now fails to compile with this error:
```
CMake Error at plugins/fclrave/CMakeLists.txt:77 (set_target_properties):
  set_target_properties called with incorrect number of arguments.
-- Configuring incomplete, errors occurred!
```

After a little bit of bisection, it seems that 28117646f6e28d7c02dc2148b2ce39b32b71dbe1 is the cause of the issue, and those quotes *were* necessary: my `FCL_LDFLAGS` only contains `-lfcl` (unquoted) which triggers the error as it's interpreted as an extra argument instead of being appended to `PLUGIN_LINK_FLAGS`.

piraka9011:
I also had this issue. Reverted the changes of the commit and it works fine for me. Maybe revert #611 in master?

PR #708, opened 2019-10-23
Parse FCL_LDFLAGS

cielavenir:
Since we already have used the technique in bulletrave, why don't we use it also in fclrave? And I updated bulletrave comment as well.

As current origin/master and origin/production are the same, it is up to you which to merge.

Note: if this is merged, please close #640


rdiankov:
does `"${PLUGIN_LINK_FLAGS}" ${FCL_LDFLAGS}`work?

cielavenir:
@rdiankov I confirmed that would cause #633 if /usr/local/lib/pkgconfig/fcl.pc's Libs is only `-lfcl`. Usual fcl.pc.in has Libs `-L${libdir} -lfcl`, so we were not aware the corner case.

cielavenir:
#795 appeared...

cielavenir:
closing this one

PR #711, opened 2019-11-06
serialize _vGeomData2~4 when geometry type is container or cage

yoshikikanemoto:
for correct KinematicsGeometryHash for container and cage.  i don't have any real problem with this yet but i've noticed this while investigating the other KinematicsGeometryHash related issue.


rdiankov:
thanks!

PR #714, opened 2019-11-11
First MR for preparing 2to3 of OpenRAVE

tgn3000:
This MR and subsequent ones prepare OpenRAVE from migrating from python2 to python3.

I have a working version of OpenRAVE running on macOS with python3 (and the latest boost), so am slowly trying to close up the gaps between this production code and the future version.

- Make `toPyArrayN` as a template function.
- Use `nullptr`, `empty()`, `data()`, instead of comparing `size()` with 0.
- Do not do `using namespace boost::python;` in the header `bindings.h`. At the very least use alias `bp` for this namespace, or explicitly do `using bp::foo();`.


rdiankov:
thanks!

PR #715, opened 2019-11-22
[WIP] Add flag for force check collision

snozawa:


rdiankov:
@snozawa can you explain when you would use `CO_ForceCheckCollision`?

rdiankov:
@snozawa ping, can you explain the use case? thanks

snozawa:
I thought the explanation might include internal information, so i already reply in the internal service.
sorry for confusing.

briefly, i tried to implement the collision checking between object `A` and `B`, where `A` and `B` are not collision-checked for other objects and `A` is checked only with `B`.

PR #718, opened 2019-12-03
Set permission

AntoineRyu:
Adds support to change the default permissions set when generating and saving an iksolver.


python/databases/inversekinematics.py
```diff
@@ -295,15 +295,33 @@ def clone(self,envother):
...
+                log.info('changing filepermissions of path %s to \'%s\' recursively.', basepath, filepermissions)
+                try:
+                    os.chmod(basepath, filepermissions)
+                    for root, dirs, files in os.walk(basepath):
+                        for d in dirs:
+                            os.chmod(os.path.join(root, d), filepermissions)
+                        for f in files:
+                            os.chmod(os.path.join(root, f), filepermissions)
+                except OSError:
+                    pass
```
rdiankov: @AntoineRyu can you print the error so that we know when permissions fail? thanks!

AntoineRyu: Done

rdiankov:
thanks!

PR #723, opened 2020-01-08
Implemented binary trajectory feature

ShaysRebellion:


src/libopenrave/trajectory.cpp
```diff
@@ -21,54 +21,194 @@
...
+            
+        /* Read metadata */
+
+        // Read number of groups
+        uint16_t numGroups = 0;
+        ReadBinaryUInt16(I, numGroups);
+
+        ConfigurationSpecification spec;
+        spec._vgroups.resize(numGroups);
+        for (unsigned int groupIndex = 0; groupIndex < numGroups; groupIndex++)
```
ziyan:
```c++
FOREACH(itGroup, spec._vgroups) {
ReadBinaryString(I, itGroup->name);
...
}
```

src/libopenrave/trajectory.cpp
```diff
@@ -21,54 +21,194 @@
...
+        this->Init(spec);
+
+        /* Read trajectory data */
+        std::vector<dReal> trajectoryData;
+        ReadBinaryVector(I, trajectoryData);
+        this->Insert(this->GetNumWaypoints(), trajectoryData);
     }
-    else {
-        throw OPENRAVE_EXCEPTION_FORMAT(_("error, failed to find </trajectory> in %s"),buf.str(),ORE_InvalidArguments);
+    // Backwards compatible with old XML trajectory files
```
ziyan: I suggest you put a `return` before the `else` and keep the legacy part of code untouched with no indentation.

src/libopenrave/trajectory.cpp
```diff
@@ -21,54 +21,194 @@
...
+    WriteBinaryUInt16(O, VERSION_NUMBER);
+
+    /* Store meta-data */
+
+    // Indicate size of meta data
+    const ConfigurationSpecification& spec = this->GetConfigurationSpecification();
+    const uint16_t numGroups = spec._vgroups.size();
+    WriteBinaryUInt16(O, numGroups);
+
+    for (uint16_t groupIndex = 0; groupIndex < numGroups; groupIndex++)
```
ziyan:
```c++
FOREACHC(itGroup, spec._vgroups) {
WriteBinaryString(O, itGroup->name);
...
}
```

ziyan:
Also @ShaysRebellion please change the target branch of this pull request to `production` instead of `master`

src/libopenrave/trajectory.cpp
```diff
@@ -21,38 +21,175 @@
...
+        std::vector<dReal> trajectoryData;
+        ReadBinaryVector(I, trajectoryData);
+        this->Insert(this->GetNumWaypoints(), trajectoryData);
+        return shared_from_this();
+    }
+    // Backwards compatible with old XML trajectory files
+    
     stringbuf buf;
-    stringstream::streampos pos = I.tellg();
+    I.seekg((size_t) beginningPosition);    // Move back to old position
```
ziyan: You renamed `pos` to `beginningPosition`, are you sure it is not used later?

src/libopenrave/trajectory.cpp
```diff
@@ -21,38 +21,175 @@
...
+
+    // Read binary trajectory files
+    if (binaryFileHeader == MAGIC_NUMBER)
+    {   
+        uint16_t versionNumber = 0;
+        ReadBinaryUInt16(I, versionNumber);
+
+        if (versionNumber != VERSION_NUMBER)
+        {
+            // Throw assertion (for future versions)
```
ziyan: Please raise exception here.

ziyan:
```c++
throw OPENRAVE_EXCEPTION_FORMAT(_("unsupported trajectory format version %d "),versionNumber,ORE_InvalidArguments);
```

src/libopenrave/trajectory.cpp
```diff
@@ -21,38 +21,175 @@
...
+    FOREACHC(itgroup, spec._vgroups)
+    {
+        WriteBinaryString(O, itgroup->name);   // Writes group name
+        WriteBinaryInt(O, itgroup->offset);    // Writes offset
+        WriteBinaryInt(O, itgroup->dof);       // Writes dof
+        WriteBinaryString(O, itgroup->interpolation);  // Writes interpolation
+    }
+
+    /* Store data waypoints */
+    std::vector<dReal> trajectoryData;
```
rdiankov: @ShaysRebellion please do this without creating a new vector instance (since it allocates memory)

src/libopenrave/trajectory.cpp
```diff
@@ -21,38 +21,175 @@
...
+        {
+            ReadBinaryString(I, itgroup->name);             // Read group name
+            ReadBinaryInt(I, itgroup->offset);              // Read offset
+            ReadBinaryInt(I, itgroup->dof);                 // Read dof
+            ReadBinaryString(I, itgroup->interpolation);    // Read interpolation
+        }
+        this->Init(spec);
+
+        /* Read trajectory data */
+        std::vector<dReal> trajectoryData;
```
rdiankov: @ShaysRebellion please do this without creating a temporary `vector`

src/libopenrave/trajectory.cpp
ShaysRebellion: @rdiankov i noticed that there is no `vector<dReal>` member for waypoints in `TrajectoryBase` defined in `trajectory.h` that would allow us to read or write directly to this member (allowing us to remove the temporarily created vector). However, `GenericTrajectory` does contain such a member (called `_vtrajdata`)

i could:
1) move the entire serialization and deserialization to `GenericTrajectory` (because it seems like `GenericTrajectory` is the only class inheriting from `TrajectoryBase`, correct me if i'm wrong)
2) create a function that returns a pointer reference to the member function (declared in `trajectory.h` but implemented in `generictrajectory.cpp`
3) additional options?

src/libopenrave/trajectory.cpp
ShaysRebellion: @rdiankov please see above discussion

src/libopenrave/trajectory.cpp
rdiankov: @ShaysRebellion yes, in this case, we have to override the serialize function in GenericTrajectory and do the fast thing. You can still leave the slow function in TrajectoryBase, just please put comments saying that this is a default implementation and faster implementations should be done in the specific Trajectory classes.

ShaysRebellion:
@rdiankov made the requested changes. please let me know if there's anything that should be fixed

src/libopenrave-core/generictrajectory.cpp
```diff
@@ -319,26 +409,76 @@ class GenericTrajectory : public TrajectoryBase
...
+            this->Init(spec);
+
+            /* Read trajectory data */
+            ReadBinaryVector(I, this->_vtrajdata);
+            return shared_from_this();
         }
-        O << "</trajectory>" << endl;
+        
+        // Call on parent class deserialize method for backwards compatible with old XML trajectory files
+        return TrajectoryBase::deserialize(I);
```
rdiankov: @ShaysRebellion did you try deserializing with XML to see if it works. don't you need to call `seekg` to reset `I`?

rdiankov:
@ShaysRebellion please talk to testing team on how to insert this into our openrave tests. thanks

rdiankov:
@ShaysRebellion please do perf test with fopen/fwrite/fread to see if perf changes compared to current istream implementation (it is important we get the fastest serialization). thanks (cc @ziyan )

ShaysRebellion:
@rdiankov made changes suggested by @ziyan. i also locally tested the binary serialization and deserialization functions, and they seem to be working

> @ShaysRebellion please do perf test with fopen/fwrite/fread to see if perf changes compared to current istream implementation (it is important we get the fastest serialization). thanks (cc @ziyan )
i'll look into this

PR #725, opened 2020-01-15
Add geometry details

Puttichai:
- now displaying `mouse on <kinbodyname>:<linkname>:<geomname> ...` on the viewer


plugins/qtosgrave/osgrenderitem.cpp
```diff
@@ -278,8 +279,10 @@ void KinBodyItem::Load()
 //        }
 
         _veclinks.push_back(LinkNodes(posglinkroot, posglinktrans));
+        _vecgeoms.at(_veclinks.size() - 1).resize(0);
 
         for(size_t igeom = 0; igeom < porlink->GetGeometries().size(); ++igeom) {
+            _vecgeoms[_veclinks.size() - 1].push_back( GeomNodes(new osg::Group(), new osg::MatrixTransform()) );
```
rdiankov: @Puttichai the way you are indexing `_vecgoems` looks dangerous, perhaps there is a better way?!

rdiankov:
thanks!

PR #728, opened 2020-01-20
Support distance queries with the FCL collision checker

marlinstrub:
This PR implements distance queries for the FCL collision checker.

I took the logic from the collision detection callbacks and applied it to the new distance callbacks with the necessary tweaks as I recognized them. As is done in the PQP plugin, the collision checking methods throw if the `CO_Distance` option is set but no collision report is passed in. Based on the (limited) testing I've done so far it seems to work, but I'm relatively new to using OpenRAVE and would appreciate an experienced set of eyes glancing over this.

This PR also changes the default BVH used in FCL from `OBB` to `OBBRSS`, because FCL (0.5.0) does not support distance queries between `BV_OBB` and `GEOM_*` type objects. OBBRSS uses OBB for collision detection and RSS for distance queries (which seem to be the best representations for the respective tasks). If this is a bad idea, please let me know and I'll revert the change.

### DIFF
--- a/plugins/fclrave/fclcollision.h
+++ b/plugins/fclrave/fclcollision.h
@@ -79,6 +79,7 @@ public:
 
             // set the gjk solver (collision checking between convex bodies) so that we can use hints
             _request.gjk_solver_type = fcl::GST_INDEP;
+            _distanceRequest.gjk_solver_type = fcl::GST_INDEP;
 
             if( !!_report ) {
                 _report->Reset(_pchecker->GetCollisionOptions());
@@ -95,6 +96,8 @@ public:
         boost::shared_ptr<FCLCollisionChecker> _pchecker;
         fcl::CollisionRequest _request;
         fcl::CollisionResult _result;
+        fcl::DistanceRequest _distanceRequest;
+        fcl::DistanceResult _distanceResult;
         CollisionReportPtr _report;
         std::vector<KinBodyConstPtr> const& _vbodyexcluded;
         std::vector<LinkConstPtr> const& _vlinkexcluded;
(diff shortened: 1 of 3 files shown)


marlinstrub:
Some more testing revealed that very infrequently the distance computation with the GST_INDEP GJK solver returns different results for the same collision objects. I haven't seen this happen when GST_LIBCCD is used as GJK solver.

rdiankov:
@marlinstrub Marlin, first, congrats in compiling openrave with FCL ;0)

Our team tested these changes, and it looks like there's a slight perf decrease with regular checks using "OBBRSS"  vs "OBB" (perhaps due to internal structures getting updated?).
I think it would be great to keep "OBB" as the default, and for those users that want to use distance queries, during creation of the FCL checker, to specify the `bvhrepresentation` instead. This is as simple as

```
RaveCreateCollisionChecker(env,'fcl_ DynamicAABBTree2 OBBRSS')
```

marlinstrub:
@rdiankov Haha, thanks :)

Sure thing, it seems reasonable then to keep OBB as the default. I think OBBRSS creates the RSS representation regardless of whether distance queries are used, so your suggestion regarding internal structures makes sense to me. I've reverted the change. Thanks for looking into this, Rosen & team.

rdiankov:
@marlinstrub Marlin, congratulations on your first merge!

Actually, our team works on the "production" branch of openrave, and I hadn't merged the latest and greatest fixed into master for some time, which I just did.

Best,

marlinstrub:
Thanks! And thank you for all the work you put into this project and for keeping it open source.

Great, that's good to know.

PR #731, opened 2020-01-25
Increase test ray length for calibration visibility checking

rmislam:
In openrave, the visibility checking done for calibration uses test rays with a hardcoded length of 2 meters. This is why the calibration board used to get occluded when we calibrated using large robots. Although we fixed this in planning common, in openrave this problem still exists. This PR changes the hardcoded length from 2 meters to 200 meters. It's just a quick fix that reduces the likelihood of running into this problem.

This isn't for Mujin, it's for people outside of Mujin who are using openrave for calibration or visibility checking in general. I think this will save them a headache.


rdiankov:
sure ;0)

PR #735, opened 2020-02-01
Add include boost/python/slice when not using pybind11

Puttichai:
Fixed compilation error
```
openrave/python/bindings/include/openravepy/openravepy_trajectorybase.h:68:28: error: ‘openravepy::py::slice’ has not been declared
     object __getitem__(py::slice indices) const;
```


rdiankov:
interesting, thanks!

PR #740, opened 2020-02-17
Modify grasper

Puttichai:
- added `vordereddofindices` to allow control of the order of gripper dofs when doing grasp simulations


rdiankov:
sure, thanks!

PR #741, opened 2020-02-17
Gripper control

AntoineRyu:
Added gripperControlID to tool definition.


rdiankov:
renamed to `gripperid`, added `GripperInfo` class.

PR #744, opened 2020-02-26
Add readable interface support to binary trajectory format

strmio:
This patch adds serialisation/deserialisation code for readable interfaces to the binary trajectory (de)serialiser. It adds `<trajectory_addon>...</trajectory_addon>` block at the end of the current format, and elements inside will be read to a trajectory after the binary part is processed. The format can be debatable though.

### DIFF
--- a/src/libopenrave-core/generictrajectory.cpp
+++ b/src/libopenrave-core/generictrajectory.cpp
@@ -23,7 +23,7 @@ namespace OpenRAVE {
 
 // To distinguish between binary and XML trajectory files
 static const uint16_t MAGIC_NUMBER = 0x62ff;
-static const uint16_t VERSION_NUMBER = 0x0001;  // Version number for serialization
+static const uint16_t VERSION_NUMBER = 0x0002;  // Version number for serialization
 
 static const dReal g_fEpsilonLinear = RavePow(g_fEpsilon,0.9);
 static const dReal g_fEpsilonQuadratic = RavePow(g_fEpsilon,0.45); // should be 0.6...perhaps this is related to parabolic smoother epsilons?
@@ -46,6 +46,7 @@ inline void WriteBinaryInt(std::ostream& f, int value)
 
 inline void WriteBinaryString(std::ostream& f, const std::string& s)
 {
+    BOOST_ASSERT(s.length() <= std::numeric_limits<uint16_t>::max());
     const uint16_t length = (uint16_t) s.length();
     WriteBinaryUInt16(f, length);
     if (length > 0)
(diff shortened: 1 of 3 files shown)


src/libopenrave-core/generictrajectory.cpp
```diff
@@ -441,6 +441,17 @@ class GenericTrajectory : public TrajectoryBase
             WriteBinaryVector(O, this->_vtrajdata);
 
             WriteBinaryString(O, GetDescription());
+
+            if( !GetReadableInterfaces().empty() ) {
```
rdiankov: @strmio this is not a good way to maintain a binary format by looking for any left over data and trying to match that to a string. You should instead:
1. update traj version and preserve back compatibility
2. write the number of readable interfaces, and then write the string contents of each as a binary blob

When reading, you should read the number of interfaces, and for every blob of data, you should create a `StringXMLReadable` interface for every readable interface.

rdiankov: @strmio ping

strmio: You mean it is okay to put aside XML formatted readable interfaces?

src/libopenrave-core/generictrajectory.cpp
```diff
@@ -441,6 +442,18 @@ class GenericTrajectory : public TrajectoryBase
             WriteBinaryVector(O, this->_vtrajdata);
 
             WriteBinaryString(O, GetDescription());
+
+            // Readable interfaces, added on VERSION_NUMBER=0x0002
+            const uint16_t numReadableInterfaces = GetReadableInterfaces().size();
+            WriteBinaryUInt16(O, numReadableInterfaces);
+            FOREACHC(itReadableInterface, GetReadableInterfaces()) {
+                WriteBinaryString(O, itReadableInterface->first);  // xmlid
+                std::stringstream ss;
```
rdiankov: @strmio will be possibly faster if you take `std::stringstream` out of the scope and do `ss.clear(); ss.str("");` in the loop

src/libopenrave-core/generictrajectory.cpp
```diff
@@ -484,6 +498,24 @@ class GenericTrajectory : public TrajectoryBase
             /* Read trajectory data */
             ReadBinaryVector(I, this->_vtrajdata);
             ReadBinaryString(I, __description);
+
+            // versions >= 0x0002 have readable interfaces
+            if (versionNumber >= 0x0002)
```
rdiankov: @strmio you have to reset any current readable interfaces in the trajectory...

src/libopenrave-core/generictrajectory.cpp
strmio: Sure.

src/libopenrave-core/generictrajectory.cpp
strmio: All right. By the way behaviours of `deserialize()` for already initialised `Interface` fields are different between the text version and the binary version. While the binary one always resets these fields regardless whether those are given in a trajectory file or not, the text one does not. For example, the text version appends readable interfaces in a file to existing ones without wiping them, and leave the previously set `Description` if not specified in the file. Probably `deserialize()` for the text format also should reset these fields.

rdiankov: @strmio i agree, let's fix the text deserialization also

rdiankov:
thanks~

PR #747, opened 2020-02-27
WIP: JSONSerialization2 readable test

bucketzxm:


python/bindings/openravepy_global.cpp
```diff
@@ -102,6 +102,17 @@ class PyReadable
         return ConvertStringToUnicode(ss.str());
     }
 
+    object SerializeJSON(dReal fUnitScale=1.0, int options=0) const
```
ziyan: If you are going to add `SerializeJSON`, please also add `DeserializeJSON`

python/bindings/openravepy_global.cpp
```diff
@@ -102,6 +102,17 @@ class PyReadable
...
+    {
+        JSONReadablePtr pjsonreadable = OPENRAVE_DYNAMIC_POINTER_CAST<JSONReadable>(_readable);
+        rapidjson::Document doc;
```
ziyan: Move this after the if test.

PR #751, opened 2020-03-03
Reload DOF limits

tgn3000:
- When we call `SetDOFLimits`, openrave does set new DOF limits in `JointInfo::_vlowerlimit` and `_vupperlimit`. But when constructing an`IkFastSolver` object in `ikfastsolver.cpp`, it calls `IkFastSolver::SetJointLimits` and copies the joint limits into `_qlower`, `_qupper`, and never changes them. So the new DOF limits do not take effect.
- Add `IkSolverBase::ReloadData` so that we update data from OpenRAVE to `IkSolverBase`. Add python binding function.
- Add `override` identifiers. Make `IkSolverBase::_SetJointLimits` private.


rdiankov:
@tgn3000 i think using the `KinBody::Prop_JointLimits` change callback should be enough, why do you also need `ReloadData`?

rdiankov:
@tgn3000 ping

tgn3000:
@rdiankov `KinBody::Prop_JointLimits` is only used in `IkFastSolver::Init`, which calls `IkFastSolver::SetJointLimits` that copies `KinBody::JointInfo::_vlowerlimit` to `IkFastSolver::_qlower`, and `KinBody::JointInfo::_vupperlimit` to `IkFastSolver::_qupper`.

Imagine the user has already constructed an `IkFastSolver` object. Then he calls `KinBody::SetDOFLimits` to update `KinBody::JointInfo::_vlowerlimit` and `KinBody::JointInfo::_vupperlimit`. However, the `IkFastSolver` object still keeps the old joint limits in its `_qlower` and `_qupper`. Now, if the user calls `manip.FindIKSolution(s)`, then the solver does not return solutions within the new DOF limits.

Am I missing something here, or I am not aware of some particular use of updating data from openrave to a plugin?

tgn3000:
Right now `class ChangeCallbackData : public UserData` is in `kinbody.cpp`, and `KinBody::RegisterChangeCallback` upcasts `ChangeCallbackDataPtr` to `UserDataPtr` at return. If we are to use something similar to `_cblimit->_callback()`, then we have to expose the class definition of `ChangeCallbackData` to `kinbody.h`.

rdiankov:
@tgn3000 not sure what the problem with using `RegisterChangeCallback` is here. You should be able to register the callback in the ik solver then it will get called when the limits change. Are you saying something is not working with the design?

tgn3000:
@rdiankov The class `ChangeCallBackData` is defined `kinbody.cpp`, not `kinbody.h`.

```
class ChangeCallbackData : public UserData {
public: ...
    boost::function<void()> _callback;
...
};
```
Inside
```
UserDataPtr KinBody::RegisterChangeCallback(uint32_t, const boost::function<void()>&) const
```
we construct `ChangeCallbackDataPtr pdata` but upcast it to `UserDataPtr` at return. To do `_callback()`, inside `kinbody.cpp` it does a lot of
```
ChangeCallbackDataPtr pdata = boost::dynamic_pointer_cast<ChangeCallbackData>(pbase);
if( !!pdata ) { pdata->_callback(); }
```
so obviously other cpp files cannot benefit from the design of `ChangeCallbackData` and hence cannot execute the callback function.

Hence yes, I think the way it's currently designed does not allow me to do `_cblimt->_callback()` to reset DOF values. If you agree, then I can move the class definition of `ChangeCallbackData` to `kinbody.h` and do other appropriate changes.

rdiankov:
A "callback" means that the class (in this case kinbody) will call back into iksolver's method.
iksolver *should not* be explicitly calling `_callback`, that defeats the purpose of it.

All you have to do is call `RegisterChangeCallback` in the `Init` function where `_pmanip` is set, and that's it. It's one line.
What is the problem?

rdiankov:
and the line already exists in `Init`: `_cblimits = probot->RegisterChangeCallback`, so what is the problem?

tgn3000:
I see, thanks for explaining. Sorry I misunderstood something else. There might be other scenarios when dof limits don't get updated but unrelated to callback functions. Have seen one or two before, but couldn't find a reproducible example yet. Close for now. Thanks!

PR #757, opened 2020-03-09
Post command to GUIThread to avoid segfault

braineo:
Fix calling SetItemVisualization from other thread causes segmentation fault

How to reproduce
```py
# open any dae with openrave.py -i
viewer = env.GetViewer()
for i in range(10000):
    viewer.SendCommand('SetItemVisualization %s selected' % 'body1')
    viewer.SendCommand('SetItemVisualization %s' % 'body1')
```


rdiankov:
woot, thanks!

PR #758, opened 2020-03-09
OpenRAVE JSON Scene

ziyan:
JSON provides 16X write speed up, see #754


rdiankov:
@ziyan you have to deprecate `XMLReadable` correctly: add XMLReadable deprecated typename, in python, have to continue to add `XMLReadable` class, with potential warning every time it is used.

Also, for rapidjson serialization, did you check if we can save multiple environments in different threads at the same time without any corruption and any mutex locks?

ziyan:
Right now we lock the environment when saving the environment. Both JSON and Collada sets dof values to all zeros before taking the transform in joint info.

Going forward, the plan is to save to rapidjson document using the env lock, then save the rapidjson on a different thread.

python/bindings/openravepy_robot.cpp
```diff
@@ -1878,11 +1886,12 @@ void init_openravepy_robot()
                         "jointinfos"_a,
                         "manipinfos"_a,
                         "attachedsensorinfos"_a,
+                        "connectedbodyinfos"_a,
                         "uri"_a = "",
                         DOXY_FN(RobotBase, Init)
                         )
 #else
-                      .def("Init", initrobot, Init_overloads(PY_ARGS("linkinfos", "jointinfos", "manipinfos", "attachedsensorinfos", "uri") DOXY_FN(RobotBase, Init)))
+                      .def("Init", initrobot, Init_overloads(PY_ARGS("linkinfos", "jointinfos", "manipinfos", "attachedsensorinfos", "connectedbodyinfos", "uri") DOXY_FN(RobotBase, Init)))
```
rdiankov: @ziyan @bucketzxm Won't this break back compatibility by adding `connectedbodyinfos` in the middle? Do you have a set of merge requests to update the dependency codes on our side?

src/libopenrave/robotconnectedbody.cpp
```diff
@@ -108,52 +108,53 @@ void RobotBase::ConnectedBodyInfo::DeserializeJSON(const rapidjson::Value &value
 {
     openravejson::LoadJsonValueByKey(value, "name", _name);
     openravejson::LoadJsonValueByKey(value, "linkName", _linkname);
-    openravejson::LoadJsonValueByKey(value, "url", _url);
+    openravejson::LoadJsonValueByKey(value, "uri", _uri);
     openravejson::LoadJsonValueByKey(value, "transform", _trelative);
 
-    if(value.HasMember("links"))
```
rdiankov: @ziyan @bucketzxm shouldn't you always clear `_vLinkInfos` and other structs when deserializing since they can contain old data? Why preserve them?

src/libopenrave/trajectory.cpp
```diff
@@ -41,8 +41,12 @@ void TrajectoryBase::serialize(std::ostream& O, int options) const
     if( GetReadableInterfaces().size() > 0 ) {
         xmlreaders::StreamXMLWriterPtr writer(new xmlreaders::StreamXMLWriter("readable"));
         FOREACHC(it, GetReadableInterfaces()) {
-            BaseXMLWriterPtr newwriter = writer->AddChild(it->first);
-            it->second->Serialize(newwriter,options);
+            // some readable are not xml readable and does not get serialized here
```
rdiankov: @ziyan @bucketzxm the point here is that we serialize any Readlabe and save it as part of the trajectory. Doesn't have to be XML, why not serialize the other readables?

rdiankov:
@ziyan @bucketzxm We still have to support correct Readable conversion to JSON when the scene is loaded from collada/xml, have you tested this? (ie BodyParameters/RobotMotionParameters, etc)

python/bindings/openravepy_robot.cpp
ziyan: We have some choices:

1. Move `connectedbodyinfos` to the end and make it optional
2. Make `JSONReader` a friend class of `RobotBase` and add connected bodies after `Init` is done
3. Maybe introduce a `RoboBasetInfo` and `KinBodyInfo` that has the other infos as members, I can see that there are additional things that might need to be added in the future.

ziyan:
> @ziyan @bucketzxm We still have to support correct Readable conversion to JSON when the scene is loaded from collada/xml, have you tested this? (ie BodyParameters/RobotMotionParameters, etc)

Yes we do that already, `BodyParameters` is both a `JSONReadable` and a `XMLReadable`.

src/libopenrave/trajectory.cpp
ziyan: Because we don't plan to read `JSONReadable` out of XML format file. For example, we don't deserialize some JSONReadable out of a Collada file.

I think our focus should be trying to improve the binary format of the trajectory. This particular change here is just to provide back-compatibility, i.e. things that were in XML will still work, but no new JSON gets added to XML file.

ziyan: If we plan to support serializing `A` to both XML and JSON, we should make it implement both `XMLReadable` and `JSONReadable`, and it can serialize/deserialize freely between XML and JSON.

ziyan: At the time of writing, this https://github.com/rdiankov/openrave/pull/744 was not merged yet. We will work on top of those changes to support `JSONReadable` in binary format.

src/libopenrave/robotconnectedbody.cpp
ziyan: This one is fixed. @bucketzxm please check other info structure and see if similar problem exists.

python/bindings/openravepy_robot.cpp
rdiankov: @ziyan i was thinking it's about time we make a proper Info structure (your #3 suggestion). How about you leave the current `Init` for back compat purposes and create a new `InitFromInfo` function that takes in a `RobotInfo` structure?

src/libopenrave/trajectory.cpp
ziyan: @bucketzxm has added support for json format in the binary trajectory format.

python/bindings/openravepy_robot.cpp
ziyan: Added `KinBodyInfo` and `RobotBaseInfo` and `InitFromInfo`.

rdiankov:
@bucketzxm @ziyan production now has new GripperInfo structure. Please make sure to support it in the new scene saving. Thanks~

rdiankov:
@ziyan @bucketzxm is this merge still valid?

bucketzxm:
> @ziyan @bucketzxm is this merge still valid?

@rdiankov this is the one ready to review

rdiankov:
@bucketzxm @ziyan 

1. can we have: `exception.h` -> `openraveexception.h`?

2. can we put all the Info structs outside of the openrave classes?

include/openrave/robot.h
```diff
@@ -450,19 +510,44 @@ class OPENRAVE_API RobotBase : public KinBody
...
+        AttachedSensorInfo(const AttachedSensorInfo& other) {
+            *this = other;
+        };
+        AttachedSensorInfo& operator=(const AttachedSensorInfo& other) {
+            _id = other._id;
+            _name = other._name;
+            _linkname = other._linkname;
+            _trelative = other._trelative;
+            _sensorname = other._sensorname;
+            rapidjson::Document docSensorGeometry;
```
rdiankov, line 523: @bucketzxm @ziyan is there a way to re-use the same allocator in `_docSensorGeometry` without having memory grow?

rdiankov:
@bucketzxm  i'm not sure if openrave users are all using compilers that can support the below.

```
+    uint64_t _revision = 0;  ///< environment current revision
```

rdiankov:
> @bucketzxm i'm not sure if openrave users are all using compilers that can support the below.
> 
> ```
> +    uint64_t _revision = 0;  ///< environment current revision
> ```

@bucketzxm maybe it's ok...... we'll just have to tell those users to upgrade their compilers...

PR #759, opened 2020-03-10
add string readable serialize and deserialize for json

bucketzxm:


bucketzxm:
TODO: add python binding

PR #760, opened 2020-03-11
fix pertubation type to be same as _pertubation

kanbouchou:
make ``perturbation`` type same as ``_perturbation``


rdiankov:
oops, nice catch~

PR #761, opened 2020-03-12
assign a new variable when load json value

bucketzxm:
Things like Ikparams' transform can be deserialized with partially setting translation or rotation.  Previous way may keep some old value.


ziyan:
Is this also in mujinjson.h?

bucketzxm:
although for the ikparam case, when we read value, it's safe to 

> Is this also in mujinjson.h?

yes

PR #762, opened 2020-03-13
Let cmake search for LAPACK implementation

cielavenir:


rdiankov:
@cielavenir sure, thanks. i'll trust you verified all the lapack cmake variables are filled correctly.

PR #763, opened 2020-03-16
Make sure RAPIDJSON_ASSERT is using BOOST_ASSERT

ziyan:
To throw exception and avoid cassert crashing the process.


rdiankov:
thanks!

PR #764, opened 2020-03-16
add jsonreadable to binary trajectory and serialize ikparam by using openravejson.h

bucketzxm:


ziyan:
I assume you've tested the planner ikparam change, because if someone is relying on the old format, they will break.

PR #766, opened 2020-03-16
python binding ikparams deserialize format

bucketzxm:


ziyan:
Please make a separate `DeserialieJSON` instead. This changes the meaning of the constructor.

PR #767, opened 2020-03-17
Fix joint spheres using the biggest radius, centered around the last …

AntoineRyu:
…iterated link, even if not properly centered


rdiankov:
@AntoineRyu there might be a bug in this implementation too. let's talk in person

rdiankov:
thanks~

PR #771, opened 2020-03-18
make ikfastpy link to @LAPACK_LIBRARIES@

cielavenir:
caveats: we will not able to share iksolver if CMAKE_INSTALL_PREFIX is different (you know what I mean actually).


rdiankov:
@cielavenir why don't you put `find_package(LAPACK)` in the root CMakeLists.txt?

rdiankov:
@cielavenir can you move `find_package(LAPACK)` to the root CMakeLists.txt?

cielavenir:
WIP until our liblapack policy is firmly settled. We are aware that building dynamic-linked so fails when find_package(LAPACK) selects MKL.

cielavenir:
[to track the issue properly] related comment https://gitlab.kitware.com/cmake/cmake/-/issues/20491#note_719552

PR #772, opened 2020-03-19
bindings for InitFromInfo

bucketzxm:


src/libopenrave/kinbody.cpp
```diff
@@ -125,6 +125,58 @@ void ElectricMotorActuatorInfo::DeserializeJSON(const rapidjson::Value& value, d
...
+    if (_vLinkInfos.size() > 0) {
+        rapidjson::Value rLinkInfoValues;
+        rLinkInfoValues.SetArray();
+        rLinkInfoValues.Reserve(_vLinkInfos.size(), allocator);
+        FOREACHC(it, _vLinkInfos) {
+            rapidjson::Value linkInfoValue;
+            (*it)->SerializeJSON(linkInfoValue, allocator, options);
+            rLinkInfoValues.PushBack(linkInfoValue, allocator);
+        }
+        value.AddMember("linkInfos", rLinkInfoValues, allocator);
```
ziyan: I think you should do `links` directly. Remove the `Infos`

ziyan: Same for other places, like `joints`, etc.

python/bindings/openravepy_kinbody.cpp
```diff
@@ -1977,6 +2038,86 @@ py::object PyKinBody::PyGrabbedInfo::__unicode__() {
...
+    FOREACHC(itJointInfo, info._vJointInfos) {
+        vJointInfos.append(*itJointInfo);
+    }
+    _vJointInfos = vJointInfos;
+#endif
+}
+
+std::string PyKinBody::PyKinBodyInfo::__str__() {
+#ifdef USE_PYBIND11_PYTHON_BINDINGS
+    return boost::str(boost::format("<kinbodyinfo: %s>")%_uri);
```
ziyan: This is supposed to be lower cased? `kinbodyinfo`

src/libopenrave/kinbody.cpp
ziyan: Please don't forget `attachedSensors` etc.

PR #773, opened 2020-03-20
remove duplicate code in overloading GetIkParameterization

tgn3000:
- Remove duplicate code in the overloading function `RobotBase::Manipulator::GetIkParameterization` that takes `const IkParameterization&`, and let it simply call the one with `const IkParameterizationType`.
- To ensure we copy the custom data map, add `IkParameterization::SetCustomMap`.
- Note that only `IKP_Lookat3D` and `IKP_TranslationLocalGlobal6D` have different formulas in the overloaded function. Not sure which one is correct. The code reviewer can make them consistent.


tgn3000:
@rdiankov  ping

tgn3000:
@rdiankov please take a look when free

yoshikikanemoto:
rosen asked me to review this pull request.
you cleaned up `GetIkParameterization(ikparam, inworld)` which overwrites given ikparam. because there were a lot of duplicated lines in it. the change looks fine in terms of identity of expected results between before and after.
but because you cleaned up the one that overwrites ikparam, you needed to add the new code to copy members of `IkParametrazation` to newly created instance other than _transform and _type (for now only _mapCustomData). i think this new demand for copying can be vulnerable to future changes like adding new members. how about cleaning up reversely?

tgn3000:
@yoshikikanemoto 
`_type` and `_transform` are set in these `Set` functions of  `IkParameterization` already, so they should be good.
```
    inline void SetTransform6D(const Transform& t)
    inline void SetTransform6DVelocity(const Transform& t)
    inline void SetRotation3D(const Vector& quaternion)
    inline void SetTranslation3D
    inline void SetDirection3D(const Vector& dir)
    inline void SetRay4D(const RAY& ray)
    inline void SetLookat3D(const Vector& trans)
    inline void SetLookat3D(const RAY& ray)
    inline void SetTranslationDirection5D(const RAY& ray)
    inline void SetTranslationXY2D(const Vector& trans)
    inline void SetTranslationXYOrientation3D(const Vector& trans)
    inline void SetTranslationLocalGlobal6D(const Vector& localtrans, const Vector& trans)
    inline void SetTranslationXAxisAngle4D(const Vector& trans, dReal angle) 
    inline void SetTranslationYAxisAngle4D(const Vector& trans, dReal angle)
    inline void SetTranslationZAxisAngle4D(const Vector& trans, dReal angle)
    inline void SetTranslationXAxisAngleZNorm4D(const Vector& trans, dReal angle)
    inline void SetTranslationYAxisAngleXNorm4D(const Vector& trans, dReal angle)
    inline void SetTranslationZAxisAngleYNorm4D(const Vector& trans, dReal angle) {
```

Puttichai:
@tgn3000 What @yoshikikanemoto is concerned about is that in the future you will need to keep adding more such `Set` functions, one for each new member of `IkParameterization`. With the old implementation, such `Set` functions are never needed as everything will be copied over with 
```
IkParameterization ikp = ikparam; // copies the custom data
```
So having this copying stuff only in `operator=` function seems easier to maintain in that regard.

PR #775, opened 2020-03-26
Let GetEndEffectorTransform do exactly what it's meant to do

tgn3000:
- Originally `Manipulator::GetEndEffectorTransform` is just an alias of `Manipulator::GetTransform`, which takes into account the manipulator's local tool transform. Hence `GetEndEffectorTransform` does not do what it is meant to do.

To get EE's true transform, currently the user has to do `pmanip->GetEndEffector()->GetTransform()`. Not neat.

- To ensure backward compatibility, we need to replace `GetEndEffectorTransform` by `GetTransform` in all python code across all repos. Then we replace the python binding line.


rdiankov:
@tgn3000 I agree with this change, but I'm afraid that old code might be relying on the old behavior. Have you done a thorough sweep of the code bases (both openrave and m*) to make sure this is not the case?

tgn3000:
@rdiankov Glad that you agree. I just did a sweep of all repos, and replaced those occurrences in m___np______gc____n. Should be good to go.

tgn3000:
@rdiankov ping, good to merge

yoshikikanemoto:
@rdiankov  this looks good for me also

tgn3000:
@rdiankov  ping

rdiankov:
@tgn3000 you are giving up on this pull request?

PR #776, opened 2020-03-27
Call SetStateValues before _neighstatefn

Puttichai:
Make sure to call `parameters->SetStateValues` before `parameters->_neighstatefn` since `_neighstatefn` expects the configuration to already be set.


rdiankov:
thanks!

PR #777, opened 2020-03-30
wip: Binary readable interfaces

kanbouchou:
fix readable interfaces was not serialized / deserialized correctly with binary trajectory format.

@strmio


rdiankov:
thanks!

PR #779, opened 2020-04-01
Ikparam jacobian

kanbouchou:
with this mr, jacobian solver accepts ikparam as goal in addition to transform.

This enables us to use jacobian refinement on less-than-6dof robots.

Only TranslationZAxisAngle4D is implemented for now.


rdiankov:
thanks~

PR #780, opened 2020-04-06
move IkReturnAction into openrave

tgn3000:
It's best to have them here instead of doing static_cast<IkReturnAction> from elsewhere.


rdiankov:
is this branch necessary?

PR #781, opened 2020-04-06
Make each OPENRAVE_PLUGINS recognized only once

cielavenir:
In some cases (mostly due to bad Python code), plugindatabase.h Destroy's `_listplugins.clear()` might not cause destroying plugins. However this does not mean those plugins can be destroyed safely.
Now, what happens when the same entry is written multiple times in OPENRAVE_PLUGINS?
AddDirectory and LoadPlugin is called for multiple times for the same entry.
In LoadPlugin second call, _listplugins are updated and Plugin destructor will be called. **DestroyPlugin() will be called for existing plugin.** Now accessing remaining plugin instances is undefined behavior.

Although `_listplugins.clear()` should always clear plugin instances, we should limit loading each plugins only once.

### DIFF
--- a/src/libopenrave/plugindatabase.h
+++ b/src/libopenrave/plugindatabase.h
@@ -441,12 +441,16 @@ protected:
         if( pOPENRAVE_PLUGINS != NULL ) {
             utils::TokenizeString(pOPENRAVE_PLUGINS, delim, vplugindirs);
         }
-        for(int i=vplugindirs.size()-1;i>0;i--){
-            int j=0;
-            for(;j<i;j++){
-                if(vplugindirs[i]==vplugindirs[j])break;
+        for(int iplugindir=vplugindirs.size()-1;iplugindir>0;iplugindir--){
+            int jplugindir=0;
+            for(;jplugindir<iplugindir;jplugindir++){
+                if(vplugindirs[iplugindir]==vplugindirs[jplugindir]){
+                    break;
+                }
             }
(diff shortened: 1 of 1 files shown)


tgn3000:
Since `vplugindirs` is `std::vector<std::string>` of not a large size, is it easier to do just
```
const std::unordered_set<std::string> splugindirs(begin(vplugindirs), end(vplugindirs));
vplugindirs = std::vector<std::string>(begin(splugindirs), end(splugindirs));
```

cielavenir:
@tgn3000 Maybe I need to get opinion about preserving OPENRAVE_PLUGINS order. (Using set causes dictionary-sorted order). How do you think @rdiankov  ?

src/libopenrave/plugindatabase.h
```diff
@@ -441,6 +441,13 @@ class RaveDatabase : public boost::enable_shared_from_this<RaveDatabase>
         if( pOPENRAVE_PLUGINS != NULL ) {
             utils::TokenizeString(pOPENRAVE_PLUGINS, delim, vplugindirs);
         }
+        for(int i=vplugindirs.size()-1;i>0;i--){
+            int j=0;
+            for(;j<i;j++){
+                if(vplugindirs[i]==vplugindirs[j])break;
```
rdiankov, line 447: @cielavenir please use `{}` and new lines and instead of single char variables like `i`, how about `iplugindir`. thanks~

rdiankov:
thanks!

PR #783, opened 2020-04-09
make FillRapidJsonFromArray as template functions

tgn3000:


rdiankov:
@tgn3000 I just noticed this, isn't it better to call `Reserve` so that rapidjson can preallocate the memory?

python/bindings/openravepy_int.cpp
```diff
@@ -124,40 +124,59 @@ object toPyObject(const rapidjson::Value& value)
...
-        } \
-} \
-
-#define FILL_RAPIDJSON_FROMARRAY_2D(pyarrayvalues, T, rapidjsonsetfn) {                            \
-        const T *vdata = reinterpret_cast<T*>(PyArray_DATA(pyarrayvalues)); \
-        for (int i = 0; i < dims[0]; i++) { \
-            rapidjson::Value colvalues(rapidjson::kArrayType); \
-            for (int j = 0; j < dims[1]; j++) { \
-                rapidjson::Value elementValue; \
-                elementValue.rapidjsonsetfn(vdata[i*dims[0]+j]); \
```
tgn3000, line 142: I just noticed the index `i*dims[0]+j` is not right in the first place, is it? @rdiankov

tgn3000:
@rdiankov Not sure if it should be
`value.Reserve(numel, allocator);` or `value.Reserve(numel*sizeof(T), allocator);`
as neither `value` nor `allocator` seems to know the type `T`.

rdiankov:
@tgn3000 i think `Reserve` is for the number of elements rather than the size.

tgn3000:
@rdiankov  Then it is be good to merge.
Checked/tested with @bucketzxm the index should be `i*dims[1]+j`.

python/bindings/openravepy_int.cpp
tgn3000: yup, it was wrong; should be i*dims[1]+j as expected

rdiankov:
looks fancy, thanks~

python/bindings/openravepy_int.cpp
```diff
@@ -124,40 +124,67 @@ object toPyObject(const rapidjson::Value& value)
...
+                              rapidjson::Value &value,
+                              rapidjson::Document::AllocatorType& allocator,
+                              RapidJsonSetFn<T> f,
+                              npy_intp const* dims) 
+{ 
+    const T *pv = reinterpret_cast<T*>(PyArray_DATA(pyarr));
+    const size_t numel = dims[0] * dims[1];
+    value.Reserve(numel, allocator);
+    for (size_t i = 0, ij = 0; i < dims[0]; ++i) { 
+        rapidjson::Value colvalues(rapidjson::kArrayType);
```
ziyan, line 159: No `Reserve` for `colvalues`?

python/bindings/openravepy_int.cpp
```diff
@@ -124,40 +124,67 @@ object toPyObject(const rapidjson::Value& value)
...
+template <typename T>
+void FillRapidJsonFromArray2D(PyArrayObject* pyarr,
+                              rapidjson::Value &value,
+                              rapidjson::Document::AllocatorType& allocator,
+                              RapidJsonSetFn<T> f,
+                              npy_intp const* dims) 
+{ 
+    const T *pv = reinterpret_cast<T*>(PyArray_DATA(pyarr));
+    const size_t numel = dims[0] * dims[1];
+    value.Reserve(numel, allocator);
```
ziyan, line 157: Is this json supposed to be `[[1,2, 3], [4, 5, 6], [7, 8, 9]]`? If so, why `Reserve` total number of elements in the first level of array?

rdiankov: oops, you are right. just patched

PR #785, opened 2020-04-15
(WIP) experiments on json reference architecture

bucketzxm:


bucketzxm:
found this is wrong after discussion, we need to save reference only in kinbodyinfo for re-initialization

PR #786, opened 2020-04-15
Fix binary trajectory startwith space in command

bucketzxm:
there is a whitespace when combining binary trajectory with commands, like `worktraj`, which caused deserialization failed.


rdiankov:
@bucketzxm feels like a very big hack. i think you should fix the code that is generating the whitespace instead

bucketzxm:
close this since this is a hack.  Caller can pass `option=0x8000` when doing serialize to  use xml 
 string format.

PR #788, opened 2020-04-24
Release all grabbed bodies

AntoineRyu:
Release all grabbed bodies when removing a kinbody from environment, otherwise the grabbed bodies can hold references to the removed/invalidated robot

### DIFF
--- a/src/libopenrave-core/environment-core.h
+++ b/src/libopenrave-core/environment-core.h
@@ -2099,6 +2099,7 @@ protected:
             }
         }
 
+        (*it)->ReleaseAllGrabbed();
         if( (*it)->IsRobot() ) {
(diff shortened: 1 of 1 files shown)


src/libopenrave-core/environment-core.h
```diff
@@ -2099,6 +2099,7 @@ class Environment : public EnvironmentBase
             }
         }
 
+        (*it)->ReleaseAllGrabbed();
         if( (*it)->IsRobot() ) {
```
rdiankov, line 2103: @AntoineRyu let's remove the IsRobot check. Thanks!

AntoineRyu: I moved the `ReleaseAllGrabbed` out of the `IsRobot` check already based on our in person discussion. The remaining check is followed by a cast of `it` to `RobotBase`, so I think it makes sense to keep it.

rdiankov: yes, you are right, thanks!

PR #789, opened 2020-04-27
do not reset __pIkSolver if direction is not changed

Puttichai:


src/libopenrave/robotmanipulator.cpp
```diff
@@ -107,13 +107,12 @@ void RobotBase::Manipulator::SetLocalToolTransform(const Transform& t)
...
 {
-    _info._sIkSolverXMLId.resize(0);
-    __pIkSolver.reset();
-    // only reset the iksolver if transform is different. note that the correct way is to see if the hash changes, but that could take too much computation. also hashes truncate the numbers.
+    // only reset the iksolver if direction is different. note that the correct way is to see if the hash changes, but that could take too much computation. also hashes truncate the numbers.
     if( (_info._vdirection-direction).lengthsqr3() > g_fEpsilonLinear ) {
         // only reset if direction is actually used
         if( !!__pIkSolver && (__pIkSolver->Supports(IKP_TranslationDirection5D) || __pIkSolver->Supports(IKP_Direction3D)) ) {
             __pIkSolver.reset();
+            _info._sIkSolverXMLId.resize(0);
```
rdiankov, line 115: @Puttichai maybe we should reset _sIkSolverXMLId just when direction is different...

rdiankov: Or are you worried that it might affect 6D solvers?

Puttichai: I think if the solver does not need to be changed, we should not reset it so that the call to `manip.SetIkSolver` does not spend extra time calling `iksolver->Init`.

rdiankov: yea, right now we don't have a good way of figuring out if `_sIkSolverXMLId` looks at direction or not. and most 6D iksolvers don't even care about the local tool transform anymore either... for now i'll accept this, but let's figure out a better protocol in the future

rdiankov:
thanks!

PR #791, opened 2020-05-01
Clone _selfcollisionchecker when cloning KinBody

Puttichai:
When cloning KinBody, also clone `_selfcollisionchecker`.


rdiankov:
thanks!

PR #794, opened 2020-05-06
Do not let <json_data> in collada override name attribute of <extra>.

ziyan:


src/libopenrave-core/colladaparser/colladawriter.cpp
```diff
@@ -2411,7 +2411,12 @@ class ColladaWriter : public daeErrorHandler
 
                 rapidjson::Document rGripperInfo;
                 (*itGripperInfo)->SerializeJSON(rGripperInfo, rGripperInfo.GetAllocator());
-
+                if (rGripperInfo.HasMember("id")) {
+                    rGripperInfo.RemoveMember("id");
+                }
+                if (rGripperInfo.HasMember("name")) {
```
rdiankov, line 2417: @ziyan why removing `name`? isn't that valid?

ziyan: For COLLADA files, `name` is already an attribute on the `<extra>` tag, similar to how `manipulator`, `attach_sensor` works.

For consistency, we continue to use `name` attribute over `name` in `<json_data>`.

Therefore, to avoid confusing, we deliberately remove `name` in `<json_data>`, because the attribute `name` takes precedence, in COLLADA files.

rdiankov: i see, ok

rdiankov:
thanks~

PR #795, opened 2020-05-10
some fix from cielavenir and qt4 conflict with boost

lianghongzhuo:
This pull request contains two prats: some fix from @cielavenir and qt4 conflict with boost
For the qt4 and boost conflict, you can see here as a reference.
https://github.com/LARG/HFO/issues/79#issuecomment-482526111

PR #796, opened 2020-05-11
not requiring Boost::Python if pybind11 is used

cielavenir:
This MR does not matter as long as we use debian stretch, but now #795 appeared...


rdiankov:
thanks~

PR #797, opened 2020-05-11
fix rapidjson

lianghongzhuo:
as discussed in https://github.com/Tencent/rapidjson/issues/1712
the rapisjson versions installed using `apt-get` will face compile error. 
so, to compile without using master version rapidjson, we need some fix.


rdiankov:
interesting, thanks for the changes!

PR #798, opened 2020-05-11
Inherit parent reader's vOpenRAVESchemeAliases for connected body

braineo:


src/libopenrave-core/colladaparser/colladareader.cpp
```diff
@@ -3584,7 +3584,16 @@ class ColladaReader : public daeErrorHandler
                 }
 
                 RobotBasePtr pbody;
-                if( reader.InitFromURI(url, AttributesList()) ) {
+
+                AttributesList atts;
+                if (!_vOpenRAVESchemeAliases.empty()) {
+                  stringstream ss;
+                  std::copy(_vOpenRAVESchemeAliases.begin(), _vOpenRAVESchemeAliases.end(), std::ostream_iterator<std::string>(ss));
```
rdiankov, line 3591: @braineo  is this separated by space? this type of one-liner is really confusing to read..

braineo:
Okay. Will use for loop instead.

rdiankov:
thanks~

PR #801, opened 2020-05-14
Fixed ComputePoseArrayDistSqr

cielavenir:
Usually we should use ComputePoseDistSqr, but somehow I found this function and its import was wrong...

Now it should return the same value as ComputePoseDistSqr.

### DIFF
--- a/python/openravepy_ext.py
+++ b/python/openravepy_ext.py
@@ -202,9 +202,8 @@ transformInversePoints = TransformInversePoints # deprecated
 def ComputePoseArrayDistSqr(pose0, posearray, quatweight=1.0):
     """computes the squared distance between pose0 and all poses in posearray
     """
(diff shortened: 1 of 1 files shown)


rdiankov:
@cielavenir `pose0` should be 7-elements, you are treating it as a 2D array here. perhaps you need a new function, or you need to test for its shape

cielavenir:
I'm pretty sorry to have misunderstood the interface. pose0 is a pose and posearray is pose-list. I fixed it.

However, I can see major arithmetic difference from ComputePoseDistSqr. Could you check my calculation? (I would need to fix basicmath.py as well)

cielavenir:
(basicmath.py update sent)

cielavenir:
now that basicmath.py fix is merged, what about this one?

rdiankov:
looks good, thanks!

PR #803, opened 2020-05-15
Use dynamic string

AntoineRyu:


rdiankov:
@AntoineRyu @kanbouchou can we make `torquelimitsmode` an `enum` so that it's clear what the numbers mean? thanks

rdiankov:
thanks~

kanbouchou:
looks good to me @rdiankov

PR #804, opened 2020-05-16
Fix smoother accel scaling

Puttichai:
When trajectory segment checking fails with `CFO_CheckTimeBasedConstraints` but not because of tool velocity/acceleration violation, previously only velocity limits are scaled down. (This velocity limit scaling was added in commit 2d9336001be576aa87202697ce247c3510b5d82f.)

However, I think acceleration limits should also be scaled down in such cases since the failures can also be due to torque limits violation.


rdiankov:
@Puttichai interesting merge, let's have testing team test this on systems to see if there are any regressions in trajectory quality.

Puttichai:
@rdiankov Tests look fine so far. I also removed `fOverallTimeMult` in the latest commit.

rdiankov:
woot!~

PR #805, opened 2020-05-19
fix SetJsonValueByKey on document type

bucketzxm:
Swap document with value lose previous  value stored in document


rdiankov:
thanks!

PR #807, opened 2020-05-20
SetActiveTrajectory: serialize trajectory in XML form

cielavenir:
otherwise, ChuckGripper() will cause SIGFPE:

```
Thread 9 "python" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7fff8db47700 (LWP 32279)]
0x00007fffc8ebde7c in OpenRAVE::GenericTrajectory::GetNumWaypoints (this=0x7fff8402eea0)
    at /home/mujin/mujin/checkoutroot/openrave/src/libopenrave-core/generictrajectory.cpp:367
367	        return _vtrajdata.size()/_spec.GetDOF();
```


cielavenir:
Note: please merge EITHER #806 or #807 (not both).

rdiankov:
thanks~

PR #809, opened 2020-05-21
Robot posture describer plugin and module

tgn3000:
- Introduced `InterfaceType::PT_PostureDescriber` that contains classes/methods describing the posture of a kinematics chain (i.e., a baselink-eelink pair) of a robot.
- Added the abstract class `PostureDescriberBase`.
- Implemented the interface (derived) class `PostureDescriber` and the module `PostureDescriberModule`. Both have the same name "`posturedescriber`".
- Implemented Python binding class `PyPostureDescriber` and the methods thereof. Implemented Python binding functions in classes `PyRobotBase`  and  `PyManipulator`.
- The method used in `PostureDescriber` to compute "robot posture values" and "robot posture states" uses the directions and anchors of some joints along the designated kinematics chain.
- The key in the describer map stored in `RobotBase::_mPostureDescribers` is a baselink-eelink pair, and the value of the map is a `PostureDescriberBasePtr` that is capable of describing the posture from baselink to eelink. Since several manipulators have the same "link pair" except different local tool transforms (transform from end-effector frame to manipulator frame), it is reasonable to use the link pair as the key of the map. This design can handle robots with two arms, provided each arm posture can be described.
- Introduced two enum classes: `NeighbouringTwoJointsRelation` describes the geometric relations between a pair of consecutive two joint axes; `RobotPostureSupportType` contains the posture supported types so far analyzed. Currently we support (1) general 6R robots where the last three axes intersect, and (2) a kind of 4R robot with the last three parallel joint axes perpendicular to the first joint axis. After we study the kinematics of other types of (custom) robots, we may be able to support more posture types in future.


include/openrave/robot.h
```diff
@@ -1053,6 +1053,67 @@ class OPENRAVE_API RobotBase : public KinBody
...
+                                      const std::vector<double>& dofvalues = {},
+                                      const std::vector<int>& dofindices = {},
+                                      const KinBody::CheckLimitsAction claoption = KinBody::CheckLimitsAction::CLA_Nothing) const;
+
+    /// \brief Computes posture states to describe the posture of a kinematics chain (prescribed by the manipulator) at the current or specified dof values.
+    /// Computes posture state integers for describing the posture of links between the manipulator's baselink and its eelink.
+    /// If the manipulator is not specified, then we use the active manipulator.
+    /// \param [out] posturevalues     a vector of robot posture state (unsigned) integers at the current of specified dof values
+    /// \param [in]  pmanip            manipulator that prescribes a kinematics chain from its baselink to its eelink
+    /// \param [in]  dofvalues         dof values set along the kinematics chain from base link to end-effector link; its size should be either 0 (i.e., using the current dof values) or the same size of the dofs of the kinematics chain
```
kanbouchou, line 1109: please update comment about expected size of ``dofvalues``

include/openrave/posturedescriber.h
```diff
@@ -0,0 +1,115 @@
...
+/// An essential kinematics chain, if not empty, is always such that both the first joint and the last joint are active and revolute.
+OPENRAVE_API LinkPair ExtractEssentialKinematicsChain(const LinkPair& kinematicsChain);
+OPENRAVE_API LinkPair ExtractEssentialKinematicsChain(const RobotBase::ManipulatorPtr& pmanip);
+OPENRAVE_API LinkPair ExtractEssentialKinematicsChain(const RobotBase::ManipulatorConstPtr& pmanip);
+
+///< \brief Acquires from the manipulator the baselink-eelink pair.
+OPENRAVE_API LinkPair GetKinematicsChain(const RobotBase::ManipulatorConstPtr& pmanip);
+
+///< \brief Computes a kinematics hash from the baselink to eelink.
+OPENRAVE_API std::string ComputeKinematicsChainHash(const LinkPair& kinematicsChain, std::vector<int>& armindices);
```
kanbouchou, line 108: let's remove ``armindices`` from this function's parameter

include/openrave/robot.h
tgn3000: updated

include/openrave/posturedescriber.h
tgn3000: done

kanbouchou:
looks good to me, passing to Rosen.

tgn3000:
@kanbouchou thanks!
@rdiankov  woot!! pass on to you! thanks~

kanbouchou:
@tgn3000 can you resolve conflicts? Thanks!

include/openrave/posturedescriber.h
```diff
@@ -0,0 +1,116 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef OPENRAVE_POSTUREDESCRIBER_H
+#define OPENRAVE_POSTUREDESCRIBER_H
+
+#include <openrave/openrave.h>
+
+namespace OpenRAVE {
+
+using LinkPair = RobotBase::LinkPair; ///< a baselink-eelink pair
```
rdiankov, line 24: @tgn3000 posture should be a property of KinBody too, so why not `KinBody::LinkPair`?

include/openrave/posturedescriber.h
```diff
@@ -0,0 +1,116 @@
...
+
+private:
+    virtual const char* GetHash() const final;
+};
+
+using PostureDescriberBasePtr = boost::shared_ptr<PostureDescriberBase>;
+
+///< \brief Acquires from the manipulator the essential kinematics chain in the form of a link pair.
+/// An essential kinematics chain, if not empty, is always such that both the first joint and the last joint are active and revolute.
+OPENRAVE_API LinkPair ExtractEssentialKinematicsChain(const LinkPair& kinematicsChain);
```
rdiankov, line 101: @tgn3000 please prefix global openrave functions with `Rave`

plugins/posturedescriber/posturedescriberinterface.h
```diff
@@ -0,0 +1,167 @@
...
+{
+    return x = x & y;
+}
+
+template <typename T>
+inline T operator|=(T& x, T y)
+{
+    return x = x | y;
+}
+
```
rdiankov, line 68: @tgn3000 omg, this is super fancy. can we get rid of it and instead you do the manual casting when you need to? ie `(uint32_t)myenum1 & (uint32_t)myenum2`

include/openrave/robot.h
```diff
@@ -1247,6 +1247,69 @@ class OPENRAVE_API RobotBase : public KinBody
     /// \brief \ref KinBody::SetNonCollidingConfiguration, also regrabs all bodies
     virtual void SetNonCollidingConfiguration();
 
+    using LinkPair = std::array<LinkPtr, 2>; ///< a (base link, end-effector link) pair. We henceforth write baselink and eelink for brevity.
+
+    /// \brief Registers a posture describer for a kinematics chain from baselink to eelink
+    /// \param [in] kinematicsChain    a kinematics chain prescribed by a baselink-eelink pair
+    /// \param [in] pDescriber         a robot posture describer
+    /// \return true if either (1) pDescriber is not null and supports the kinematics chain, or (2) pDescriber is null and we successfully register the describer for the kinematics chain.
+    virtual bool SetPostureDescriber(const LinkPair& kinematicsChain, PostureDescriberBasePtr pDescriber);
```
rdiankov, line 1256: @tgn3000 why not have one PostureDescriber for the whole kinbody? And internally that will do the correct logic for base-end link algorithms. Why do you need to register  per base-endlink algorithms? That just makes it harder to use and setup and manage later.

cielavenir:
@tgn3000 really? I think this is useful to avoid solutionindices unstability and user-friendly posture configuration (see #1043)

cielavenir:
Talked with @kanbouchou and he still wants to have this done, so for now I reopen this pull request. /cc @lazydroid

Issue #808, opened 2020-05-20
Error compiling IKfast module on latest production branch


hungpham2511:
I am running the lastest production branch (as of a few days ago), on Ubuntu 16.04. Running `ikmodel.autogenerate` fail with the below error:

```
/home/hung/.openrave/kinematics.72948cfc3ff77d3858ae895ad25226f4/ikfast0x1000004b.IkParameterizationType.Transform6D.x86_64.15_16_18_19_20_21_f12_17.cpp:12404:11: error: ‘class ikfast::IkSingleDOFSolutionBase<double>’ has no member named ‘indices’
 vinfos[7].indices[0] = _ij21[0];
           ^
/home/hung/.openrave/kinematics.72948cfc3ff77d3858ae895ad25226f4/ikfast0x1000004b.IkParameterizationType.Transform6D.x86_64.15_16_18_19_20_21_f12_17.cpp:12405:11: error: ‘class ikfast::IkSingleDOFSolutionBase<double>’ has no member named ‘indices’
 vinfos[7].indices[1] = _ij21[1];
           ^
```

Look like there is a bug in the generation of the cpp file.

hungpham2511:
Any hints/pointers to the relevant source files are appreciated. Thanks!

tgn3000:
@hungpham2511 Hi, I wonder whether you could share the scene file (*.dae or *.xml) so that I try reproducing.

Can you also find the `ikfast.h` that you installed openrave with? See
```
template <typename T>
class IkSingleDOFSolutionBase {
...
    std::array<unsigned char, IKSINGLEDOFSOLUTIONBASE_INDICES_SIZE> indices; ///< unique index of the solution used to keep track on what part it came from. sometimes a solution can be repeated for different indices. store at least another repeated root
...
};
```

The `indices` member has been in this class for some years. Wondering which version did you upgrade from? Thanks!

hungpham2511:
Thanks for answering! I have a closer look at the compilation error and it turns out the compiler was looking for std=c++11. 
```
/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support \
```
This is probably because I compile OpenRAVE with C++11. I will close this issue as it's not an issue with the source code.

hungpham2511:
[...]

PR #810, opened 2020-05-21
Fix: Add c++11 flag to inversekinematics.py

hungpham2511:
Fix an error that breaks ikfast compilation, due to missing C++11 flag which is required by the array header introduced in 374297b26c6007ad54431df33dc60274d3881874

Reported in #808


rdiankov:
thanks!

PR #811, opened 2020-05-21
Add DynamicsConstraints enum. Update torquelimitmode from int to that…

AntoineRyu:
… new enum

Same as https://github.com/rdiankov/openrave/pull/803, but targeting production instead of teachworker branch.


include/openrave/kinbody.h
```diff
@@ -38,6 +38,12 @@ enum GeometryType {
     GT_Cage=6, ///< a container shaped geometry with removable side walls. The side walls can be on any of the four sides. The origin is at the bottom of the base. The inner volume of the cage is measured from the base to the highest wall.
 };
 
+enum DynamicsConstraints {
```
kanbouchou, line 41: what about ``DynamicsConstraintsType``?

Puttichai: I was going to ask the exact same question. Since it's an `enum` of type/option for torque checking, I think it might be more informative/less confusing if the name is something like `DynamicsConstraintsType`.

rdiankov: @Puttichai @kanbouchou i i agree, let's make it consistent

rdiankov:
yay~

PR #812, opened 2020-05-22
Improve planners

Puttichai:
- Added env id in planner logging
- Added checking callbacks in smoothers.
- Added checking for time limit in smoothers.
- Using `OPENRAVE_PLANNER_STATUS` to create a `PlannerStatus` to automatically get `errorOrigin`, etc.


rdiankov:
thanks~

PR #814, opened 2020-06-01
initliaze readable interface in *Info

bucketzxm:


rdiankov:
@ziyan @bucketzxm is this MR ready?

PR #817, opened 2020-06-08
Write connect body collada

ziyan:
- [x] all links, joints are being written to dae even though some of them could be belonging to connected body, need to skip writing those


rdiankov:
woot~ thanks~

PR #821, opened 2020-06-11
Planning Failure Analyzer (PickPlaceLog)

ShaysRebellion:
Initial work on collecting all robot joints values and colliding links for all planning collisions. Plan on later using robot joint values to visualize collision areas on UI.


src/libopenrave/planner.cpp
```diff
@@ -99,7 +99,16 @@ int AddStatesWithLimitCheck(std::vector<dReal>& q, const std::vector<dReal>& qde
     return status;
 }
 
-PlannerStatus::PlannerStatus() : statusCode(0)
+PlannerStatus::PlannerStatus() {}
+
+// To maintain backwards compatibility...
+PlannerStatus::PlannerStatus(const std::string& description, const int statusCode, CollisionReportPtr report) : description(description), statusCode(statusCode)
+{
+    // not sure if this works...
```
ziyan, line 107: No, it does not. `__FILE__` is a macro resolved at compile time, it will always be `planner.cpp`

include/openrave/planner.h
```diff
@@ -487,15 +487,20 @@ class OPENRAVE_API PlannerStatus
...
-    PlannerStatus(const std::string& description, const int statusCode, const IkParameterization& ikparam, CollisionReportPtr report=CollisionReportPtr());
-    PlannerStatus(const std::string& description, const int statusCode, const std::vector<dReal>& jointValues, CollisionReportPtr report=CollisionReportPtr());
+    PlannerStatus(const std::string& description, const int statusCode);
+
+    // PlannerStatus();
+    // PlannerStatus(const int statusCode);
+    // PlannerStatus(const std::string& description, const int statusCode, CollisionReportPtr report=CollisionReportPtr());
+    // PlannerStatus(const std::string& description, const int statusCode, const IkParameterization& ikparam, CollisionReportPtr report=CollisionReportPtr());
+    // PlannerStatus(const std::string& description, const int statusCode, const std::vector<dReal>& jointValues, CollisionReportPtr report=CollisionReportPtr());
+
```
ziyan, line 497: Please keep all the constructors exactly same as before. I don't think you have a good enough reason to change them.

include/openrave/planner.h
```diff
@@ -507,14 +512,18 @@ class OPENRAVE_API PlannerStatus
...
-    IkParameterization ikparam;      // Optional,  the ik parameter that failed to find a solution.
-    std::vector<dReal> jointValues; // Optional,  the robot's joint values in rad or m
-    CollisionReportPtr report;       ///< Optional,  collision report at the time of the error. Ideally should contents contacts information.
+    PlannerParametersConstPtr parameters;   ///< parameters used in the planner
+    std::string description;                ///< Optional, the description of how/why the error happended. Displayed to the user by the UI. It will automatically be filled with a generic message corresponding to statusCode if not provided.
+    uint32_t statusCode;                    // combination of PS_X fields (PlannerStatusCode)
+    IkParameterization ikparam;             // Optional, the ik parameter that failed to find a solution.
+    // IMPORTANT: indices of vRobotJointValues and vCollisionReportString should correspond to each other!
+    std::vector<std::vector<dReal>> vRobotJointValues;  // Records robot joint values at all collisions
+    std::vector<std::string> vCollisionReportString;    // Records all collision reports
```
ziyan, line 521: Please do not rename any of the members. I don't think you have a good enough reason to change them. Only add the new members you introduced.

plugins/rplanners/rplanners.h
```diff
@@ -448,6 +448,11 @@ class SpatialTree : public SpatialTreeBase
         return bHasAdded ? ET_Sucess : ET_Failed;
     }
 
+    // Covnert to boost shared pointer, e.g. CollisionReportPtr?
+    std::string GetCollisionReportString() {
+        return this->_constraintreturn->_report.__str__();
+    }
```
ziyan, line 454: You don't need `this->`

ziyan: What is `__str__`, is that python?!

src/libopenrave/planner.cpp
ShaysRebellion: Oh, I actually just copied the function to maintain backwards compability. I'll delete it.

include/openrave/planner.h
ShaysRebellion: I `grep`ed through the projects, and it seems like these constructors are not being used. Also, part of the motivation is because I changed the member variables.

include/openrave/planner.h
ShaysRebellion: Hm, all right, then.

plugins/rplanners/rplanners.h
ShaysRebellion: Hm, okay, sure.

include/openrave/planner.h
ziyan: I `grep`ed through the projects too, and I see plenty of usages. Please compile openrave before you ask for review.

plugins/rplanners/rrt.h
```diff
@@ -456,6 +457,13 @@ Some python code to display data::\n\
             // extend A
             ExtendType et = TreeA->Extend(_sampleConfig, iConnectedA);
 
+            if (et == ET_Failed && IS_DEBUGLEVEL(Level_Verbose)) {
+                std::vector<dReal> robotJointValues;
+                this->_getstatefn(robotJointValues); // Making the assumption that _getstatefn is shared throughout planning, e.g. shares same _getstatefn as DynamicsCollisionConstraint checker
+                planningstatus.vRobotJointValues.push_back(robotJointValues); // Is this slow?
+                planningstatus.vCollisionReports.push_back(this->_treeForward.GetCollisionReportString());
+            }
+
```
Puttichai, line 466: Why should we collect joint values every time a tree extension fails?

Puttichai:
Talking about collision-related stuff, I think one useful thing might be to also report colliding _geometry_ in collision report. Currently a collision report only reports colliding links.

plugins/rplanners/rrt.h
ShaysRebellion: I was thinking that we could visualize the locations where collision happens on the UI (though according to Kei and Antoine, teachworker team has a tool in which they dump the tree and visualize all the nodes)

ShaysRebellion:
> Talking about collision-related stuff, I think one useful thing might be to also report colliding _geometry_ in collision report. Currently a collision report only reports colliding links.

What's the difference between link and geometry collision? For the link collision, I'm mostly interested in the parent link name (KinBody name), so we can know what bodies are colliding

plugins/rplanners/rrt.h
Puttichai: The joint values collected here are ones that are _not_ in the tree. I think it will rarely be useful/informative to look at these joint values.

Tree dumping functionality is already there and can be used by `planner->SendCommand("DumpTree")`.

What problems are you trying to solve that require these information/ will be easier to solve with these information?

Puttichai:
There can be multiple geometries living inside one link. Reporting which geometries are colliding will give even more specific information of where the collision actually happens.

include/openrave/planner.h
```diff
@@ -491,12 +491,14 @@ class OPENRAVE_API PlannerStatus
...
     PlannerStatus(const std::string& description, const int statusCode, const IkParameterization& ikparam, CollisionReportPtr report=CollisionReportPtr());
     PlannerStatus(const std::string& description, const int statusCode, const std::vector<dReal>& jointValues, CollisionReportPtr report=CollisionReportPtr());
+    
     virtual ~PlannerStatus();
 
     PlannerStatus& SetErrorOrigin(const std::string& errorOrigin);
     PlannerStatus& SetPlannerParameters(PlannerParametersConstPtr parameters);
 
     void SaveToJson(rapidjson::Value& rPlannerStatus, rapidjson::Document::AllocatorType& alloc) const;
+    void SaveToJson2(rapidjson::Value& rPlannerStatus, rapidjson::Document::AllocatorType& alloc) const;
```
ziyan, line 501: What's this `SaveToJson2`? Please merge your additions into existing `SaveToJson`

include/openrave/planner.h
```diff
@@ -507,14 +509,17 @@ class OPENRAVE_API PlannerStatus
...
+    std::string description;                ///< Optional, the description of how/why the error happended. Displayed to the user by the UI. It will automatically be filled with a generic message corresponding to statusCode if not provided.
+    uint32_t statusCode;                    // combination of PS_X fields (PlannerStatusCode)
+    IkParameterization ikparam;             // Optional, the ik parameter that failed to find a solution.
+    std::vector<dReal> jointValues;         // Optional, the robot's joint values in rad or m
+    CollisionReportPtr report;              ///< Optional,  collision report at the time of the error. Ideally should contents contacts information.
+    // IMPORTANT: indices of vRobotJointValues and vCollidingBodies should correspond with each other!
+    std::vector< std::vector<dReal> > vRobotJointValues;                                 // Records robot joint values at all collisions
+    std::vector< std::vector< std::pair<std::string, std::string> > > vCollidingBodies;  // Records all collision reports
+
+    std::string errorOrigin;                // Auto, a string representing the code path of the error. Automatically filled on construction.
```
ziyan, line 522: Try to avoid these kind of cosmetic space changes, they make code reviewing much less efficient. It's hard to tell which ones you really changed and whether something is hidden in here.

plugins/rplanners/rrt.h
```diff
@@ -456,6 +457,13 @@ Some python code to display data::\n\
             // extend A
             ExtendType et = TreeA->Extend(_sampleConfig, iConnectedA);
 
+            if (et == ET_Failed && IS_DEBUGLEVEL(Level_Verbose)) {
+                std::vector<dReal> robotJointValues;
+                this->_parameters->_getstatefn(robotJointValues); // Making the assumption that _getstatefn is shared throughout planning, e.g. shares same _getstatefn as DynamicsCollisionConstraint checker
+                planningstatus.vRobotJointValues.push_back(robotJointValues);                           // Is this slow?
+                planningstatus.vCollidingBodies.push_back(this->_treeForward.GetCollisionBodies());    // Is this slow?
```
ziyan, line 466: 1. Colliding link will be more useful than colliding bodies.
2. I think you need to do the count collection at this level instead of letting the caller do it.
3. Saving `std::pair<KinBody::LinkConstPtr, KinBody::LinkConstPtr>` instead of `std::string` will probably be better, caller can figure out the body and the link.

bucketzxm:
btw you are targeting master instead of production branch

plugins/rplanners/rrt.h
ShaysRebellion: I see...I was hoping to collect the associated joint values for every sampled configuration, so that later we can move the robot to those joint locations and see where the robot (or transferred workpiece ) is colliding during planning (and essentially visualize where in the configuration space is being sampled).

To confirm, then we could essentially do the same with this `planner->SendCommand("DumpTree")`?

include/openrave/planner.h
ShaysRebellion: Okay, will do

plugins/rplanners/rrt.h
ShaysRebellion: I wonder if `_configurations` in `ConstraintFilterReturn` also essentially provides the same information (i.e. robot configuration used during collision checking)

ShaysRebellion:
Here's sample information from a planning problem involving collision: 
[sampleInfo.txt](https://github.com/rdiankov/openrave/files/4778717/sampleInfo.txt)

src/libopenrave/planner.cpp
```diff
@@ -99,18 +99,12 @@ int AddStatesWithLimitCheck(std::vector<dReal>& q, const std::vector<dReal>& qde
     return status;
 }
 
-PlannerStatus::PlannerStatus() : statusCode(0)
-{
-}
+PlannerStatus::PlannerStatus() {}
```
ziyan, line 102: By removing `statusCode(0)` you introduced a bug. Can you revert to exactly what it was before here?

src/libopenrave/planner.cpp
```diff
@@ -99,18 +99,12 @@ int AddStatesWithLimitCheck(std::vector<dReal>& q, const std::vector<dReal>& qde
...
 
 PlannerStatus::PlannerStatus(const std::string& description, const int statusCode, CollisionReportPtr report) :
     description(description),
     statusCode(statusCode),
-    report(report)
-{
-    // not sure if this works...
-    errorOrigin = str(boost::format("[%s:%d %s] ")%OpenRAVE::RaveGetSourceFilename(__FILE__)%__LINE__%__FUNCTION__);
-}
+    report(report) {}
```
ziyan, line 107:
```
report(report) {}
```

should be

```
report(report)
{
}
```

src/libopenrave/planner.cpp
```diff
@@ -180,6 +174,18 @@ PlannerStatus& PlannerStatus::SetPlannerParameters(PlannerParametersConstPtr par
     return *this;
 }
 
+// This assumes that collisionReport plink1 and plink2 fields are populated! Current implementation does not check if vLinkColliding is populated (when is it populated)?
+void PlannerStatus::UpdatePlannerStatusInfo(CollisionReport& collisionReport)
```
ziyan, line 178: I don't think this name conveys much useful information. Consider a better name that reflects what it does.

src/libopenrave/planner.cpp
```diff
@@ -180,6 +174,18 @@ PlannerStatus& PlannerStatus::SetPlannerParameters(PlannerParametersConstPtr par
...
+{      
+    if (!!collisionReport.plink1 && !!collisionReport.plink2) {
+        std::pair<KinBody::LinkConstPtr,KinBody::LinkConstPtr> collisionPair(collisionReport.plink1, collisionReport.plink2);
+        if (mCollidingLinksCount.find(collisionPair) == mCollidingLinksCount.end()) {
```
ziyan, line 182: Efficiency can be improved here. You did a `find()` operation, you might as well use the resulting iterator to set the value. Doing another `mCollidingLinksCount[collisionPair]` to increment means you are finding the same iterator again.

plugins/rplanners/rplanners.h
```diff
@@ -607,6 +607,10 @@ class SpatialTree : public SpatialTreeBase
         }
     }
 
+    ConstraintFilterReturnPtr GetConstraintReport() {
```
ziyan, line 610:
```
const ConstraintFilterReturnPtr& GetConstraintReport() const {
```

plugins/rplanners/rrt.h
```diff
@@ -456,6 +457,11 @@ Some python code to display data::\n\
             // extend A
             ExtendType et = TreeA->Extend(_sampleConfig, iConnectedA);
 
+            // Maybe should be able to pass options to CheckCollisionConstraint function in SpatialTree and also use it as a check here!
+            if (et == ET_Failed && IS_DEBUGLEVEL(Level_Verbose)) {
```
ziyan: Instead of `IS_DEBUGLEVEL`, use a flag passed down through `planningoptions`

ShaysRebellion: I just realized that the problem with using `planningoptions` is that it's passed by `binpickingmodule`, but `binpickingmodule` does not allow user to define `planningoptions`, so the only way to get this information is to hardcode the value from `binpickingmodule`

ziyan: You can change `binpickingmodule`

src/libopenrave/planner.cpp
```diff
@@ -99,7 +99,7 @@ int AddStatesWithLimitCheck(std::vector<dReal>& q, const std::vector<dReal>& qde
     return status;
 }
 
-PlannerStatus::PlannerStatus() : statusCode(0)
+PlannerStatus::PlannerStatus(): statusCode(0)
```
ziyan, line 102: Please add back the space.

plugins/rplanners/rplanners.h
```diff
@@ -607,6 +608,10 @@ class SpatialTree : public SpatialTreeBase
         }
     }
 
+    ConstraintFilterReturnPtr GetConstraintReport() const {
```
ziyan, line 611: Return `const ConstraintFilterReturnPtr&`, you shouldn't need to modify the content of the report or the pointer.

CMakeLists.txt
```diff
@@ -317,7 +317,11 @@ if(NOT MSVC71 AND NOT MSVC80)
     set(Boost_LIBRARY_DIRS $ENV{BOOST_LIBRARYDIR})
   endif()
   set(Boost_ADDITIONAL_VERSIONS "1.46.1" "1.45" "1.44" "1.43" "1.42" "1.41" "1.40" "1.39" "1.38" "1.37.0" "1.37" "1.35.0" "1.34.1" "1.34.0" "1.34" "1.33.1" "1.33.0" "1.33" ${Boost_ADDITIONAL_VERSIONS})
-  find_package(Boost COMPONENTS regex filesystem system python thread iostreams date_time)
+  if(OPT_PYTHON AND NOT USE_PYBIND11_PYTHON_BINDINGS)
+    find_package(Boost COMPONENTS regex filesystem system python thread iostreams date_time)
+  else()
+    find_package(Boost COMPONENTS regex filesystem system thread iostreams date_time)
+  endif()
```
ziyan, line 324: Why are these changes here? Seems unrelated to this PR

include/openrave/planner.h
```diff
@@ -491,11 +492,13 @@ class OPENRAVE_API PlannerStatus
     PlannerStatus(const std::string& description, const int statusCode, CollisionReportPtr report=CollisionReportPtr());
     PlannerStatus(const std::string& description, const int statusCode, const IkParameterization& ikparam, CollisionReportPtr report=CollisionReportPtr());
     PlannerStatus(const std::string& description, const int statusCode, const std::vector<dReal>& jointValues, CollisionReportPtr report=CollisionReportPtr());
+    
     virtual ~PlannerStatus();
 
     PlannerStatus& SetErrorOrigin(const std::string& errorOrigin);
     PlannerStatus& SetPlannerParameters(PlannerParametersConstPtr parameters);
 
+    void IncrementCollisionPairCount(CollisionReport& collisionReport); // This assumes that collisionReport plink1 and plink2 fields are populated!
```
ziyan, line 501: `IncrementCollisionPairCount` still sounds weird. Maybe `AddCollisionReport`.

Also, parameter should be const since you are not changing it, `const CollisionReport& collisionReport`.

plugins/rplanners/rplanners.h
```diff
@@ -126,7 +126,7 @@ class SpatialTreeBase
 
     /// extends toward pNewConfig
     /// \return true if extension reached pNewConfig
-    virtual ExtendType Extend(const std::vector<dReal>& pTargetConfig, NodeBasePtr& lastnode, bool bOneStep=false) = 0;
+    virtual ExtendType Extend(const std::vector<dReal>& pTargetConfig, NodeBasePtr& lastnode, bool bOneStep=false, bool saveStatusData=false) = 0;
```
ziyan, line 129: It's weird that you call it `saveStatusData` and have a enum named `PO_StatusDetail` and what it does is add `CFO_FillCollisionReport` to options to `CheckPathAllConstraints`. None of these 3 names are related to each other.

plugins/rplanners/rplanners.h
```diff
@@ -305,7 +305,7 @@ class SpatialTree : public SpatialTreeBase
         RAVELOG_VERBOSE("computed in %fs", (1e-9*(utils::GetNanoPerformanceTime()-starttime)));
     }
 
-    virtual ExtendType Extend(const vector<dReal>& vTargetConfig, NodeBasePtr& lastnode, bool bOneStep=false)
+    virtual ExtendType Extend(const vector<dReal>& vTargetConfig, NodeBasePtr& lastnode, bool bOneStep=false, bool saveStatusData=false)
```
ziyan, line 308: Instead of exposing saveStatusData, why not expose `constraintFilterOptions` with default `0xffff|CFO_FillCheckedConfiguration`

include/openrave/planner.h
```diff
@@ -62,6 +62,7 @@ enum PlannerStatusCode
 enum PlanningOptions
 {
     PO_NoStatusDetail = 1, ///< if set, then do not output any PlannerStatus details, just the finish code. This allows system to be faster.
+    PO_StatusDetail = 2 /// If set, outputs all PlannerStatus details. This will cause the system to be slower.
```
ziyan, line 65: Perhaps `PO_AddCollisionReport = 1 << 1`

plugins/rplanners/rplanners.h
ziyan: Or, `additionalConstraintFilterOptions`

src/libopenrave/planner.cpp
```diff
@@ -180,6 +178,20 @@ PlannerStatus& PlannerStatus::SetPlannerParameters(PlannerParametersConstPtr par
     return *this;
 }
 
+// This assumes that collisionReport plink1 and plink2 fields are populated! Current implementation does not check if vLinkColliding is populated (when is it populated)?
+void PlannerStatus::IncrementCollisionPairCount(CollisionReport& collisionReport)
+{      
+    if (!!collisionReport.plink1 && !!collisionReport.plink2) {
+        std::pair<KinBody::LinkConstPtr,KinBody::LinkConstPtr> collisionPair(collisionReport.plink1, collisionReport.plink2);
+        std::map< std::pair<KinBody::LinkConstPtr,KinBody::LinkConstPtr>, unsigned int >::iterator collideLinkPairKey = mCollidingLinksCount.find(collisionPair);
```
ziyan, line 186: No need for space after `<` and before `>`.

CMakeLists.txt
ShaysRebellion: Oh, I actually merged from `origin/production` (so that it effectively branches from that branch). So, this may be the result of that branch merge.

PR #823, opened 2020-06-12
Configurationspec json

bucketzxm:


ziyan:
@bucketzxm please increment version

rdiankov:
@bucketzxm @ziyan shouldn't we unify the JSON functions? Sometimes we use `SerializeJSON` and sometimes it is `SaveToJson`, so which is it? Please pick one.

src/libopenrave/configurationspecification.cpp
```diff
@@ -147,6 +147,50 @@ bool ConfigurationSpecification::operator!=(const ConfigurationSpecification& r)
...
+
+    rValue.SetObject();
+
+    rapidjson::Value rGroups;
+    rGroups.SetArray();
+    rGroups.Reserve(_vgroups.size(), alloc);
+    for(size_t iGroup = 0; iGroup < vgroupindices.size(); iGroup++) {
+        int iGroupIndex = vgroupindices[iGroup];
+        const ConfigurationSpecification::Group& group = _vgroups[iGroupIndex];
+        rapidjson::Value rGroup;
```
rdiankov, line 184: @bucketzxm you should add `rGroup.SetObject()` here rather than relying on SetJsonValueByKey to set it for you.

ziyan:
Let's change to `SerializeJSON`.

src/libopenrave/configurationspecification.cpp
bucketzxm: resolved

bucketzxm:
changed name from `SaveToJson/LoadFromJson` to `SerializeJSON/DeserializeJSON`

CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set( CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS TRUE )
 
 # Define here the needed parameters
 set (OPENRAVE_VERSION_MAJOR 0)
-set (OPENRAVE_VERSION_MINOR 40)
+set (OPENRAVE_VERSION_MINOR 41)
 set (OPENRAVE_VERSION_PATCH 8)
```
rdiankov, line 8: @bucketzxm if you are to update MINOR version, you can reset PATCH to 0

bucketzxm: resolved

rdiankov:
thanks!

PR #824, opened 2020-06-17
Tighter constraint on shortcutting

yoshikikanemoto:
this PR introduces 3 stage constraints relaxation

1. path search : `NSO_TighterConstraintsForSampling`
2. shortcutting : `NSO_TighterConstraintsForShortcutting`
3. smoothing : `NSO_OnlyHardConstraints`


tgn3000:
For our reference, this is part of planningcommon MR #1325

python/bindings/openravepy_ikparameterization.cpp
```diff
@@ -597,6 +600,7 @@ void init_openravepy_ikparameterization()
                                    .def("__repr__",&PyIkParameterization::__repr__)
                                    .def("__mul__",&PyIkParameterization::__mul__)
                                    .def("__rmul__",&PyIkParameterization::__rmul__)
+                                   .def("toDict",&PyIkParameterization::toDict)
```
rdiankov, line 603: @yoshikikanemoto is `toDict` really necessary?

include/openrave/planner.h
```diff
@@ -78,7 +80,9 @@ enum PlannerAction
...
-    NSO_OnlyHardConstraints=2, ///< if set, then the new neighbor should be as close as possible to q+qdelta, otherwise can prioritize other constraints and only use q+qdelta as a hint. This is used in smoothers when the path is already determined and user just wants to verify that hard constraints are met only; do not modify q+qdelta unless hard constraints fail.
+    NSO_OnlyHardConstraints=2, ///< if set, then the new neighbor should be as close as possible to q+qdelta. This is used in smoothers when the path is already determined and user just wants to verify that hard constraints are met only; do not modify q+qdelta unless hard constraints fail.
+    NSO_TighterConstraintsForShortcutting=4, ///< if set, then the new neighbor should be as close as possible to q+qdelta. the new neighbor needs to preserve tighter constraint than hard constraint to give a room for the later processing. This is used in shortcutter when the path is already determined and user just wants to verify that tighter constraints are met only; do not modify q+qdelta unless tighter constraints fail.
```
rdiankov, line 84: @yoshikikanemoto I guess the purpose here is to have the neighstate function internally decide how tight the constraints should be. But maybe in the future, we would want different behaviors depending on this. In that case, perhaps renaming the enums could be `NSO_FromPathSampling` (ie birrt), `NSO_FromPathShortcutting` (ie linear shortcut), `NSO_FromTrajectorySmoother` (ie parabolic smoohter)

yoshikikanemoto: yes. i want to change behavior of neighstate function depending on from where it is called. you suggestion of changing the enum is clearer because it differentiates shortcut and smoother. i actually felt a little bit inappropriate not giving anything to smoother since smoother can be called some sort of shortcutting algorithm. i'll change it accordingly

python/bindings/openravepy_ikparameterization.cpp
yoshikikanemoto: no, it came in accidentally. i removed

PR #825, opened 2020-06-18
Fix RapidJSON SetString call without allocator

ziyan:
Fix RapidJSON SetString call without allocator, which is dangerous if caller memory gets modified.

Before the fix:

```
      "confidence": "\u001c\u007f\u0000\u0000��K�\u001c\u007f\u0000\u0000`���\u001c\u007f\u0000\u0000�\u0019R\u001c\u001d\u007f\u0000\u0000\u0000\u0005\u0000\u00001974\u0000\u0000\u0000\u0000",
```

After the fix:

```
      "confidence": "{\"global_confidence\":1.0}",
```


rdiankov:
that's pretty serious, nice catch~

PR #826, opened 2020-06-18
robust manipulator restore

kanbouchou:
before this patch, if someone does following, restoring could shuffle manipulators.

```
assert len(robot.GetManipulators()) > 1
manip = robot.GetManipulators()[0]
info = manip.GetInfo()
robot.RemoveManipulator(manip)
robot.AddManipulator(info) # now manip is the last in manipulators
```

with this patch, above code does not cause problem as long as manipulator name does not change.


kanbouchou:
@rdiankov changed log level according to our conversation in person.

rdiankov:
thanks!

PR #827, opened 2020-06-25
Remove lock env in VerifyTrajectory

Puttichai:


rdiankov:
@Puttichai can you change external doc to say that env is not locked?
thanks

rdiankov:
thanks~

PR #828, opened 2020-06-25
IkParametrization ujson serializable

yoshikikanemoto:
to generate target hash for grasping model


python/bindings/openravepy_ikparameterization.cpp
```diff
@@ -597,6 +600,7 @@ void init_openravepy_ikparameterization()
                                    .def("__repr__",&PyIkParameterization::__repr__)
                                    .def("__mul__",&PyIkParameterization::__mul__)
                                    .def("__rmul__",&PyIkParameterization::__rmul__)
+                                   .def("toDict",&PyIkParameterization::toDict)
```
rdiankov, line 603: @yoshikikanemoto is `toDict` some python reserved word? If not, why not just call it `SerializeJSON`?

yoshikikanemoto: toDict is the function name that ujson tries to use when a given object is not compatible with python native types.

yoshikikanemoto: toDict is the function name that ujson tries to use when a given object is not compatible with python native types.

yoshikikanemoto: discussed with rosen. we better explicitly call `SerializeJSON` when we serialize JSON from openrave classes since conversion is not symmetric and we might not continue using ujson forever. then close this MR

PR #831, opened 2020-07-01
(WIP) support uri(kinbody) and referenceUri(connectedBody)

bucketzxm:


include/openrave/environment.h
```diff
@@ -229,7 +229,7 @@ class OPENRAVE_API EnvironmentBase : public boost::enable_shared_from_this<Envir
     virtual bool LoadData(const std::string& data, const AttributesList& atts = AttributesList()) = 0;
 
     /// \brief loads a scene from rapidjson document
-    virtual bool LoadJSON(const rapidjson::Value& doc, const AttributesList& atts = AttributesList()) = 0;
+    virtual bool LoadJSON(const rapidjson::Document& doc, const AttributesList& atts = AttributesList()) = 0;
```
rdiankov, line 232: @bucketzxm why not `rapidjson::Value`?

include/openrave/environment.h
```diff
@@ -748,6 +749,9 @@ class OPENRAVE_API EnvironmentBase : public boost::enable_shared_from_this<Envir
     /// \brief update EnvironmentBase according to new EnvironmentBaseInfo
     virtual void UpdateFromInfo(const EnvironmentBaseInfo& info) = 0;
 
+    std::map<std::string, rapidjson::Value> _mExpandedBodyValue;  ///< map of bodyId and expandedBodyValue
```
rdiankov, line 752: @bucketzxm not sure if it is wise to save `rapidjson::Value` like this, who has control of the allocator?

bucketzxm: @rdiankov This MR is **not** ready for review.  I haven't decided which is good,  saving a rapidjson::Value or saving an info structure for later serialize.  Also need to address a better way to allocate new document when doing deserialize

ziyan: @bucketzxm ping

PR #832, opened 2020-07-01
check ptempreport is not null before printing

Puttichai:
The debug level can change. We should not rely on the checking of `IS_DEBUGLEVEL(...)` to imply that `ptempreport` is initialized previously or not.
If the debug level changes in the middle from the level higher than verbose to verbose, `report.__str__()` will be called, causing segfault.


rdiankov:
thanks~

PR #833, opened 2020-07-03
Migrate tool size safety links

kanunikov-denys:
change `toolSizeSafetyLinks` -> `toolSizeSafetyLinkNames` in gripper info to have correct conversion of link name for connected bodies.


kanunikov-denys:
decided to do in a different way, same as adding prefixes for `linknames` in robotconnectedbody.cpp

PR #834, opened 2020-07-08
Fix basic RRT params processing

Puttichai:
Fixed the issue that RRT parameters (`_minimumgoalpaths`) are not processed.


rdiankov:
thanks~

PR #835, opened 2020-07-09
Get static joint value directly from it's dof limit; use _tLeft* as the transform from parent link to child link

tgn3000:
- If between a parent link (L0) and a child link (L1) is a static joint, then its joint value is always 0 by convention, and we store the complete transform from L0 to L1 in `_tLeft*`, and let `_tRight*` be identity.

### DIFF
--- a/src/libopenrave/kinbody.cpp
+++ b/src/libopenrave/kinbody.cpp
@@ -1837,7 +1837,7 @@ void KinBody::SetDOFValues(const std::vector<dReal>& vJointValues, uint32_t chec
                 break;
             }
         }
-        else {
+        else if ( !pjoint->IsStatic() ) {
             if( pjoint->GetType() == JointRevolute ) {
                 tjoint.rot = quatFromAxisAngle(pjoint->GetInternalHierarchyAxis(0), pvalues[0]);
                 pjoint->_doflastsetvalues[0] = pvalues[0];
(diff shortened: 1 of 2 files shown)


yoshikikanemoto:
thanks for digging into the problem.

i think it is better to move the logic to compute `tjoint` from `pvalue` to `Joint` class if you want to introduce the rule to store the complete transform into `tLeft`. that way, we can narrow scope of the local rule into `Joint` class. we can also skip multiplication to compute `tjoint` for static joint. how do you think about this idea?

Additionally to the above,  since `GetInternalHierarchy(Left|Right)Transform` is a public interface to users, changing behavior of `tLeft` and `tRight` might not be a good idea.

is there any reason behind forcing user to set `[0, 0]` joint limit? shouldn't `[-1, -1]` etc be allowed? 

FYI: not only `SetDOFValues`, `SetDOFVelocities` also can have the same improvement

rdiankov:
@yoshikikanemoto i told gary to put everything in tLeft so that we can drop tRight from any type of computation for static joints. i think this is the only way to save on computation, and i don't think it changes any conventions since joint transform should always be identity.

yoshikikanemoto:
@rosen computation-wise, putting everything to `tLeft` is ok. but the public interfaces to get anchor point of joint will be affected. some user might put some meaning to the anchor point of static joint

rdiankov:
@tgn3000 thanks! i patched up some forgotten variables.

yoshiki, whatever those people are trying to do with joint anchors on static links, i'm sure there are better ways, so i don't see why we need to support it.

tgn3000:
@yoshikikanemoto Thank you for your help and comments!
@rdiankov I saw your changes and those warnings better than my original forceful BOOST_ASSERT.
Thanks guys~

PR #836, opened 2020-07-10
Fix computing DOF velocities

yoshikikanemoto:
there were two mistakes related to computing DOF velocities.
1. `childindex` was wrongly computed
2. given link velocities were not used for one dof joint


rdiankov:
thanks!

PR #837, opened 2020-07-10
add cylinder mesh discretization parameter

kanunikov-denys:
adding a parameter to the ginfo to change the mesh accuracy for cylinders. 
I don't know what is a good name for such a parameter . For now it is ` numapproximate` 
I also noticed another issue with mesh generation for cylinders. We are replacing a cylinder with a mesh which is smaller than cylinder. We have to treat treat circle as inscribed into the mesh which we generate not as circumscribed.

### DIFF
--- a/src/libopenrave/kinbodygeometry.cpp
+++ b/src/libopenrave/kinbodygeometry.cpp
@@ -234,14 +234,15 @@ bool KinBody::GeometryInfo::InitCollisionMesh(float fTessellation)
     case GT_Cylinder: {
         // cylinder is on z axis
         dReal rad = GetCylinderRadius(), len = GetCylinderHeight()*0.5f;
-        int numverts = (int)(fTessellation*GetCylinderNumCircleDiscretiazationPoints());
-        // TODO: we should make collision obstacle bigger than cylinder, treat circle as inscribed circle not as circumscribed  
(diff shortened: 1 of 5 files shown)


src/libopenrave/kinbodygeometry.cpp
```diff
@@ -234,13 +234,15 @@ bool KinBody::GeometryInfo::InitCollisionMesh(float fTessellation)
     case GT_Cylinder: {
         // cylinder is on z axis
         dReal rad = GetCylinderRadius(), len = GetCylinderHeight()*0.5f;
-        int numverts = (int)(fTessellation*48.0f) + 3;
+        // int numverts = (int)(fTessellation*48.0f) + 3;
+        int numverts = max((int)(fTessellation*GetCylinderNumCircleApproximate()), 3); // should not be less than 3 to maintain backwards compatibility  (numverts = (int)(fTessellation*48.0f) + 3;)
```
rdiankov, line 238: @kanunikov-denys not sure if we need "tessellation" factor anymore since GeometryInfo is holding the discretized vertex count anymore... I would put this into its own new `int` parameter called `numVerticesWhenDiscretizing`. That way we can use it for spheres and other interesting primitives too. Perhaps we don't even need `fTessellation` in `InitCollisionMesh` anymore~

PR #840, opened 2020-07-14
light, efficient optimization about static joints and SetDOFValues

tgn3000:
This PR is a continuation of
https://github.com/rdiankov/openrave/pull/835.
We appreciate @yoshikikanemoto 's insight, advice, and help for this. A summary follows.

- Add `bool Joint::_bStatic` to store result from `IsStatic()` during initialization, before `_bInitialized` is set to `true` at the end of `Joint::_ComputeInternalInformation`.
- Use this `_bStatic` henceforth when calling `Joint::IsStatic()`. For static joints, do not derive `pvalues`, `tjoint` etc. variables, and directly use **0** as its joint value.
- Use references of `pjoint->_attachedbodies[0]` and `pjoint->_attachedbodies[1]` directly (because `KinBody::JointPtr` is a friend class of `KinBody::LinkPtr`), instead of calling `pjoint->GetHierarchyParentLink()` and `pjoint->GetHierarchyChildLink()`, because they copy shared pointers.
- Disable `lengthsqr4()` check for quaternion's 2-norm in [0.99f, 1.01f] when compiling in release mode, by defining `MATH_DISABLE_ASSERTS`.
Some reference: https://www.boost.org/doc/libs/1_73_0/libs/assert/doc/html/assert.html
- My signature code optimization in `SetDOFValues` to improve readability and understandability.


include/openrave/geometry.h
```diff
@@ -383,11 +383,15 @@ class RaveTransform
     template <typename U> RaveTransform(const RaveTransform<U>& t) {
         rot = t.rot;
         trans = t.trans;
-        MATH_ASSERT( rot.lengthsqr4() > 0.99f && rot.lengthsqr4() < 1.01f );
+        const T l = rot.lengthsqr4();
```
yoshikikanemoto, line 386: i think this is a good catch. for further improvement, computation of `l` can be skipped when assert is disabled. by doing like
```c++
#ifdef BOOST_ASSERT
#define MATH_ASSERT BOOST_ASSERT
#ifdef BOOST_DISABLE_ASSERTS
#define MATH_DISABLE_ASSERTS
#endif 
#else
#include <cassert>
#define MATH_ASSERT assert
#ifdef NDEBUG
#define MATH_DISABLE_ASSERTS
#endif 
#endif

...
#ifndef MATH_DISABLE_ASSERTS
        const T l = rot.lengthsqr4();
        MATH_ASSERT( l > 0.99f && l < 1.01f );
#endif
```

src/libopenrave/kinbody.cpp
```diff
@@ -1645,11 +1645,17 @@ void KinBody::SetDOFValues(const std::vector<dReal>& vJointValues, uint32_t chec
...
+    // have to compute the angles ahead of time since they are dependent on the link 
+    const int nActiveJoints = _vecjoints.size();
+    const int nPassiveJoints = _vPassiveJoints.size();
+    std::vector< std::vector<dReal> > vPassiveJointValues(nPassiveJoints);
+    for(int i = 0; i < nPassiveJoints; ++i) {
+        const KinBody::JointPtr& pjoint = _vPassiveJoints[i];
+        const boost::array<dReal, 3>& vlowerlimit = pjoint->_info._vlowerlimit;
+        const boost::array<dReal, 3>& vupperlimit = pjoint->_info._vupperlimit;
+        std::vector<dReal>& jvals = vPassiveJointValues[i];
+        if( !pjoint->IsMimic() ) {
```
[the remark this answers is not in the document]
tgn3000: done. thanks!!

include/openrave/geometry.h
tgn3000: good idea. incorporated. many thanks commenting!

yoshikikanemoto:
for the robot 7 out of 13 joints are static joint
before
```
In [1]: start=time.time(); a=[robot.SetActiveDOFValues(random.randn(6), checklimits=0) for i in range(1000000)]; time.time()-start
Out[1]: 5.586276054382324
```
after
```
In [1]: start=time.time(); a=[robot.SetActiveDOFValues(random.randn(6), checklimits=0) for i in range(1000000)]; time.time()-start
Out[1]: 3.719588041305542
```

for the robot 63 out of 80 joints are static joint:
before: 24.910272121429443 sec
after: 10.708690881729126 sec

3rdparty/crlibm-1.0beta4/atan_fast.c
```diff
@@ -65,17 +65,18 @@ static void atan_quick(double *atanhi,double *atanlo, int *index_of_e, double x)
         {
           /* compute i so that a[i] < x < a[i+1] */
           i=31;
-          if (x < arctan_table[i][A].d) i-= 16;
+          db_number TiA = arctan_table[i][A];
```
yoshikikanemoto, line 68: actually you cannot do this. i is changing while it goes through lines below

yoshikikanemoto:
tested with some real systems. i think this PR is ready @rdiankov

tgn3000:
thanks for the last reversion. I was out of my mind

include/openrave/kinbody.h
```diff
@@ -1476,6 +1476,7 @@ class OPENRAVE_API KinBody : public InterfaceBase
         Transform _tRightNoOffset, _tLeftNoOffset;         ///< same as _tLeft and _tRight except it doesn't not include the offset
         Transform _tinvRight, _tinvLeft;         ///< the inverse transformations of tRight and tLeft
         bool _bInitialized;
+        bool _bStatic = false; ///< IsStatic with both lower and upper limits equal 0
```
rdiankov, line 1479: @tgn3000 unfortunately, we don't use initializers like this in openrave headers (even though it is convenient). can you remove? thanks

src/libopenrave/kinbody.cpp
```diff
@@ -1642,128 +1642,145 @@ void KinBody::SetDOFValues(const std::vector<dReal>& vJointValues, uint32_t chec
...
+    const int nPassiveJoints = _vPassiveJoints.size();
+    std::vector< std::vector<dReal> > vPassiveJointValues(nPassiveJoints);
+    for(int i = 0; i < nPassiveJoints; ++i) {
+        const KinBody::JointPtr& pjoint = _vPassiveJoints[i];
+        if(pjoint->IsStatic()) {
+            continue;
+        }
+        const boost::array<dReal, 3>& vlowerlimit = pjoint->_info._vlowerlimit;
+        const boost::array<dReal, 3>& vupperlimit = pjoint->_info._vupperlimit;
+        std::vector<dReal>& jvals = vPassiveJointValues[i];
```
rdiankov, line 1656: @tgn3000 although minor and i like the reference, using `vector< vector >` is slow, and i think we can speed this up even further by using `std::vector< boost::array<dReal, 3> >` instead and creating a `Joint::GetValues(boost::array<dReal,3>& values)` implementation (we can do this since the max DOF per joint is 3)

include/openrave/kinbody.h
tgn3000: thanks, will address these comments tomorrow

rdiankov:
@tgn3000 for now, i allowed initializers and some c++11 stuff we also use at mujin. so looks good, thanks~

tgn3000:
@rdiankov  good to know. now I see how you want to dereference a shared pointer and use the reference henceforth. will follow that pattern in future merge requests. thanks!

PR #842, opened 2020-07-17
WIP: Pickplacelog ikfilteringinfo

ShaysRebellion:
Reports why `workspace` planner fails and what IK values (or potentially linearly interpolated IK values) cause the planner to fail

### DIFF
--- a/include/openrave/planner.h
+++ b/include/openrave/planner.h
@@ -62,7 +62,9 @@ enum PlannerStatusCode
 enum PlanningOptions
 {
     PO_NoStatusDetail = 1, ///< if set, then do not output any PlannerStatus details, just the finish code. This allows system to be faster.
-    PO_AddCollisionStatistics = 2 /// If set, Fill in collision to aid in planning failure debugging. This will slow down the system.
+    PO_AddBiRRTCollisionInfo = 2 /// If set, then will fill in collision information to aid in BiRRT planning failure debugging. This will slow down the system.
+    PO_AddLinearPlannerCollisionInfo = 4; /// If set, then will fill information to aid in linear planning failure debugging. This will slow down the system.
+    PO_AddWorkspacePlannerCollisionInfo = 0x8; /// If set, fill in collision to aid in IK filering failure debugging. This will slow down the system.
 };
 
 /// \brief action to send to the planner while it is planning. This is usually done by the user-specified planner callback function
@@ -504,7 +506,10 @@ public:
     PlannerStatus& SetPlannerParameters(PlannerParametersConstPtr parameters);
 
     void InitCollisionReport(CollisionReportPtr& collisionReport);
-    void AddCollisionReport(const CollisionReport& collisionReport);
+    void InitWorkspaceTraj(TrajectoryBaseConstPtr& workspaceTraj);
+
+    void UpdateLinkCollisionCount(const CollisionReport& collisionReport);
+
     void SaveToJson(rapidjson::Value& rPlannerStatus, rapidjson::Document::AllocatorType& alloc) const;
 
     inline uint32_t GetStatusCode() const {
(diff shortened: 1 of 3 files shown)


src/libopenrave/planner.cpp
```diff
@@ -217,7 +217,17 @@ void PlannerStatus::InitCollisionReport(CollisionReportPtr& newreport)
...
 
-void PlannerStatus::AddCollisionReport(const CollisionReport& collisionReport)
+void PlannerStatus::InitWorkspaceTraj(TrajectoryBaseConstPtr& newworkspacetraj)
+{
+    if ( !!newworkspacetraj ) {
+        // Copy boost::shared_ptr
+        workspaceTraj = newworkspacetraj;
+    } else {
+        workspaceTraj.reset(); // So we don't take up heap memory
+    }
```
ziyan, line 227: `workspaceTraj = newworkspacetraj;` will cover both cases.

src/libopenrave/planner.cpp
```diff
@@ -217,7 +217,17 @@ void PlannerStatus::InitCollisionReport(CollisionReportPtr& newreport)
     }
 }
 
-void PlannerStatus::AddCollisionReport(const CollisionReport& collisionReport)
+void PlannerStatus::InitWorkspaceTraj(TrajectoryBaseConstPtr& newworkspacetraj)
```
ziyan, line 220: `const TrajectoryBaseConstPtr& newworkspacetraj`

include/openrave/planner.h
```diff
@@ -524,10 +529,18 @@ class OPENRAVE_API PlannerStatus
...
 
+    // Filled in by BiRRT planner
     std::map< std::pair<KinBody::LinkConstPtr,KinBody::LinkConstPtr>, unsigned int > mCollidingLinksCount; // Counter for colliding links
-    uint32_t numPlannerIterations; ///< number of planner iterations before failure
+    uint32_t numPlannerIterations; ///< number of planner iterations before failure.
     uint64_t elapsedPlanningTimeUS; ///< us, elapsed time of the planner
+
+    // Filled in by workspace planner
+    // IkParameterization ikparam - in the case being returned by workspace planner, indicates which ikparam failed (may be linearly interpolated point) (conditionally populated)
+    TrajectoryBaseConstPtr workspaceTraj;   // maybe we can visualize this linear trajectory later; always populated by workspace planner (always populated)
```
ziyan, line 539: `pWorkspaceTraj`

src/libopenrave/planner.cpp
ziyan: 1. It should be called `SetWorkspaceTraj`
2. Actually why have this function? Since the member is already public.

include/openrave/planner.h
```diff
@@ -62,7 +62,9 @@ enum PlannerStatusCode
 enum PlanningOptions
 {
     PO_NoStatusDetail = 1, ///< if set, then do not output any PlannerStatus details, just the finish code. This allows system to be faster.
-    PO_AddCollisionStatistics = 2 /// If set, Fill in collision to aid in planning failure debugging. This will slow down the system.
+    PO_AddBiRRTCollisionInfo = 2, /// If set, then will fill in collision information to aid in BiRRT planning failure debugging. This will slow down the system.
+    PO_AddLinearPlannerCollisionInfo = 4, /// If set, then will fill information to aid in linear planning failure debugging. This will slow down the system.
+    PO_AddWorkspacePlannerCollisionInfo = 0x8 /// If set, fill in collision to aid in IK filering failure debugging. This will slow down the system.
```
ziyan, line 67: Why `0x8` but not the others? Perhaps `(1 << 0)`, `(1 << 1)` is easier to understand.

PR #843, opened 2020-07-21
Support depth rendering

cv3d:


rdiankov:
@cv3d in terms of API consistency, perhaps it is better to create a new function in the ViewerBase rather than overload `GetCameraImage`

cv3d:
@rdiankov Thanks for your comment. How about now?

PR #845, opened 2020-07-29
Fix container cage xml reader

osbertngok:
Fixing support for reading container / cage attributes in OpenRave XML file.


rdiankov:
thanks~

PR #846, opened 2020-07-31
fix json test failures

bucketzxm:


src/libopenrave/kinbody.cpp
```diff
@@ -5459,6 +5459,9 @@ void KinBody::ExtractInfo(KinBodyInfo& info)
     _ResolveInfoIds();
 
     info._id = _id;
+    if(info._id.empty()) {
```
ziyan, line 5462: @bucketzxm This seems out of place. Like other infos, perhaps you need to do it in UpdateFromInfo

bucketzxm: if we load a `.dae` file into `env`, and do the following command.  Then we need to set id before creating newkinbody,  or set id when we do extract info 
```
emptyEnv.UpdateFromInfo(env.ExtractInfo())
```

ziyan:
@bucketzxm does this PR fix all the test failures?

Please remember to increment minor version.

python/bindings/include/openravepy/openravepy_kinbody.h
```diff
@@ -152,6 +152,7 @@ typedef OPENRAVE_SHARED_PTR<PyGrabbedInfo> PyGrabbedInfoPtr;
     py::object _name = py::none_();
 #endif
     py::object _transform = ReturnTransform(Transform());
+    bool _isRobot = false;
```
rdiankov, line 155: @ziyan is it necessary to update MINOR version just because `_isRobot` was added here?

ziyan: Incremented PATCH version instead.

rdiankov:
thanks~

PR #847, opened 2020-07-31
[quick] bugfix  passing zero to hierarchical xml readable

kanbouchou:
recent change made options passed to ``HierarchicalXMLReadable::SerializeXML`` always zero.

This PR recovers behavior to correctly pass ``options``.


rdiankov:
thanks!

PR #850, opened 2020-07-31
handle changes because of last night's two pull requests tgn_staticJoint and orjson

tgn3000:


yoshikikanemoto:
i don't see any good reason to have _bStatic in JointInfo though

rdiankov:
@yoshikikanemoto i thought bStatic was good to have because it was a cache of some complex computation.

yoshikikanemoto:
what i'm suggesting here is moving _bStatic back to direct under the Joint class from JointInfo class.
therefore, you don't need to worry about cache. the problem happened when robot was cloned. while cloning, joint is copied by using the assignment operator. so, all members in Joint class should be copied (unless it is missed from implementation of assignment operator).

rdiankov:
@yoshikikanemoto right, bStatic should be in Joint

PR #851, opened 2020-07-31
(ManipulatorInfo, AttachedSensorInfo, GrabbedInfo) Check if members are None or not

kktk-KO:
These members are initialized to None (py_none) which is failed to convert to std::string and so on, and following code fails

```
p = ManipulatorInfo()
p.SerializeJSON()
```


kktk-KO:
AttachedSensorInfo and GrabbedInfo have same issue. I will add fix for them in this PR.

rdiankov:
thanks~

PR #853, opened 2020-08-02
Update gripperInfo in pnewgripperInfo

kktk-KO:
Currently `robot.GerGripperInfos` does not have `useLinkNames` with prefix.

I wonder if we should update gripeprInfo in connectedBodyInfo too. For now, I only update `pnewgripeprInfo`.


rdiankov:
thanks~
yes, looks like this is a regression, which somehow got introduced with the new JSON branches. before it was putting the new info correctly in `pnewgripperInfo`... sigh...

ziyan:
1. Simon did it correctly at first, on 6/18: https://github.com/rdiankov/openrave/commit/826413243086e6d012ddacfc9c8c6375d2c1113b
2. Then, later on 7/14 he merged production branch into his and resolved the conflict incorrectly: https://github.com/rdiankov/openrave/commit/9869f944b7329d76b5fd15bc01d6ca5c1625bc11

Sorry for the troubles guys~ We will be more careful when resolving merge conflict next time.

PR #854, opened 2020-08-03
Improve tracking mode camera manipulation

pablocael:
Previously camera motion to tracking object and calculate in each update call.

Using NodeTrackingManipulator in OSG achieves better camera control for rotation during track mode


plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -1029,98 +1033,9 @@ void QtOSGViewer::_UpdateCameraTransform(float fTimeElapsed)
 {
```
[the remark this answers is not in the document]
pablocael: What do you mean it does not take the parameter anymore?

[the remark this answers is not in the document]
braineo: `fTimeElapsed` is not needed anymore in `_UpdateCameraTransform`

1. It should be removed from arguments
2. It used to have this to sync with environment update for close-in animation when tracking anything. Is it ok to just use `QTime`?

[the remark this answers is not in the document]
pablocael: I see. _UpdateEnvironment(float fTimeElapsed) call UpdateCameraTransform.

QTime has enough precision for FPS measure, I can change to QTime.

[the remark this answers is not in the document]
pablocael: Wait, the only function that used delta time was _UpdateCameraTransform, but it doesnt use anymore.. Dont need to use QTime, do we?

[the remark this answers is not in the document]
pablocael: And actually I think maybe _UpdateCameraTransform can now be called _Update only, since it does other things now.. dont need to update camera transform, but other general things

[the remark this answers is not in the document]
braineo: I meant in your camera manipulator you are using a `QTime` to measure elapsed time, is it Ok not to use this `fTimeElapsed`? in order to in sync with environment update.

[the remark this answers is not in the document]
pablocael: Its ok. Its for the osg animation only. OSG also measure real frame delta time for everything so it can be framerate indepent. So in the end I could even use some other OSG function to measure, but I think using QTime is easier and cleaner imho.

plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -32,6 +33,143 @@
...
+        _transitionAnimationDuration = 1.0; //< transition animation time
+        _currentTransitionAnimationTime = 0;
+        _CreateTransitionAnimationPath(node, currentCamera, worldUpVector);
+        _time.restart();
+    }
+
+    // OSG overloaded methods
+public:
+
+    osg::Matrixd getMatrix() const {
```
braineo, line 64: `{` in next line

plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -32,6 +33,143 @@
...
+        osg::Vec3d nodeCenter;
+        osg::Quat nodeRotation;
+        computeNodeCenterAndRotation(nodeCenter, nodeRotation);
+
+        osg::Matrixd localToWorld;
+        computeNodeLocalToWorld(localToWorld);
+
+        nodeCenter = osg::Vec3d(_offset) * localToWorld;
+
+        return osg::Matrixd::translate(0.0,0.0,_distance)*osg::Matrixd::rotate(_rotation)*osg::Matrix::translate(nodeCenter);
```
braineo, line 74: Space between arguments and operators

plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -880,28 +1020,46 @@ void QOSGViewerWidget::SetViewType(int isorthogonal)
...
+    osg::Camera *hudcamera = _osghudview->getCamera();
+    hudcamera->setViewport(0, 0, width * scale, height * scale);
 
     double textheight = 12*scale;
     _osgHudText->setPosition(osg::Vec3(-width*scale/2+10, height*scale/2-textheight, -50));
     _osgHudText->setCharacterSize(textheight);
+    UpdateHUDAxisTransform(width, height);
+}
+
+void QOSGViewerWidget::UpdateHUDAxisTransform(int width, int height)
```
braineo, line 1057: No need to expose this?

plugins/qtosgrave/qtosgviewer.h
```diff
@@ -332,6 +332,9 @@ public slots:
 
     virtual void _SetCamera(RaveTransform<float> trans, float focalDistance);
     virtual void _SetCameraDistanceToFocus(float focalDistance);
+    virtual void _UpdateDistanceToFocusFromCurrentManipulator();
+    virtual void _SetTrackManipulatorToStopTracking();
+    virtual bool _SetTrackManipulatorToTrackLink(KinBody::LinkPtr link, const RaveTransform<float>& linkRelativeTranslation);
```
braineo, line 337: _StopTrackLink
_TrackLink

plugins/qtosgrave/osgviewerwidget.h
```diff
@@ -123,9 +131,19 @@ class QOSGViewerWidget : public QOpenGLWidget
     /// \brief handle case when link is selected
     void SelectOSGLink(OSGNodePtr node, int modkeymask);
 
+    /// \brief activate and configure trackmode manipulator to track given OSG node
+    /// \brief trackInfoText is the text to display in canvas about the current element being tracked
+    void StartTrackingNode(OSGNodePtr node, const std::string& trackInfoText, const osg::Vec3d& offset, double trackDistance, const osg::Vec3d& worldUpVector);
+    void StopTrackingNode();
```
braineo, line 137: TrackNode
StopTrackNode

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -1025,102 +1028,13 @@ void QtOSGViewer::_FillObjectTree(QTreeWidget *treeWidget)
     RAVELOG_VERBOSE("End _FillObjectTree....\n");
 }
 
-void QtOSGViewer::_UpdateCameraTransform(float fTimeElapsed)
+void QtOSGViewer::_Update()
```
braineo, line 1031: Do not call it `_Update` name it specifically what it is updating

braineo:
Tidy code style with rest plz. I'm not pointing out every place

plugins/qtosgrave/qtosgviewer.h
```diff
@@ -332,12 +332,19 @@ public slots:
 
     virtual void _SetCamera(RaveTransform<float> trans, float focalDistance);
     virtual void _SetCameraDistanceToFocus(float focalDistance);
+    virtual double _GetCameraDistanceToFocus();
+    virtual void _StopTrackingLink();
+    virtual bool _StartTrackingLink(const std::string& infoText, KinBody::LinkPtr link, const RaveTransform<float>& linkRelativeTranslation);
```
braineo, line 337: _StopTrackLink
_TrackLink

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -1395,45 +1360,72 @@ bool QtOSGViewer::WriteCameraImage(int width, int height, const RaveTransform<fl
...
+void QtOSGViewer::_StopTrackingLink()
+{
+    _ptrackinglink.reset();
+    _posgWidget->StopTrackNode();
+    if(_cameraMoveModeButton != NULL) {
+        _cameraMoveModeButton->setEnabled(true);
+    }
+}
+
+inline bool AreTransformsEqual(const RaveTransform<float>& t1, const RaveTransform<float>& t2, float tolerance=1e-3)
```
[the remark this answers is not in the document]
pablocael: Usually comparing vectors with == is not a good idea in general

plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -32,6 +33,149 @@
...
+class OpenRAVETracker : public osgGA::NodeTrackerManipulator {
+public:
+    OpenRAVETracker()
+    {
+        _currentTransitionAnimationTime = 0;
+        _transitionAnimationPath = new osg::AnimationPath();
+        _transitionAnimationPath->setLoopMode(osg::AnimationPath::NO_LOOPING);
+    }
+
+    void TrackNode(osg::Node* node, const osg::Vec3d& offset, double trackDistance, osg::Camera* currentCamera, const osg::Vec3d& worldUpVector)
```
rdiankov, line 45: @pablocael if there's no reason to use raw pointer, let's not. You can do `osg::Node& node`. thanks

plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -32,6 +33,149 @@
...
+        return osg::Matrixd::translate(-nodeCenter) * osg::Matrixd::rotate(_rotation.inverse()) * osg::Matrixd::translate(0.0,0.0,-_distance);
+    }
+
+    virtual void setByInverseMatrix(const osg::Matrixd& matrix)
+    {
+        _rotation = matrix.getRotate().inverse();
+    }
+
+private:
+    void _CreateTransitionAnimationPath(osg::Node* node, osg::Camera* currentCamera, const osg::Vec3d& worldUpVector)
```
rdiankov, line 103: @pablocael if there's no reason to use raw pointer, let's not. You can do osg::Node& node. thanks

plugins/qtosgrave/osgviewerwidget.cpp
braineo: Used `OsgNodePtr` instead

plugins/qtosgrave/osgviewerwidget.cpp
braineo: node not used here. Removed.

plugins/qtosgrave/osgviewerwidget.cpp
pablocael: Rosen, we changed to ref pointer since we are using ref pointers as arguments in many other places already. Copying the ref pointer is ok since it will copy the ref count. Its safer than raw pointer (if any exception happens, for instance).

PR #856, opened 2020-08-03
A mimic joint is not static

tgn3000:
Fixed a bug where a mimic joint has `IsStatic()` as `true`.

### DIFF
--- a/src/libopenrave/kinbodyjoint.cpp
+++ b/src/libopenrave/kinbodyjoint.cpp
@@ -622,25 +622,12 @@ bool KinBody::Joint::IsStatic() const
         return _bStatic;
     }
     
-    if( IsMimic() ) {
-        bool bstatic = true;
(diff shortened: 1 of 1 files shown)


yoshikikanemoto:
oops, i think we should be showing warning when mimic joint depends on static joint in this case

yoshikikanemoto:
@tgn3000 I checked returned value from `IsStatic()` of mimic joint which depended on non-static joint. It correctly returned `False`.

given the fact that `Joint::_ComputeInternalInformation` is called after `Joint::SetMimicEquations` in `KinBody::_ComputeInternalInformation`, i actually don't see any problem on the current implementation. am i missing anything?

rdiankov:
@tgn3000 i'm not sure that's a safe assumption. in any case, i committed the proper fix, so no need for this pull request anymore.

PR #857, opened 2020-08-03
Fix copying gripperInfo 

yoshikikanemoto:
`id` was missed from copying. this was causing unnecessary re-generation of cache because cloned robot returned different `id`.


rdiankov:
nice catch~

ziyan:
@bucketzxm Please add test for copying infos.

PR #859, opened 2020-08-04
cage geometry json  backward compatibility 

bucketzxm:


src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,25 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
-            {"nx", KinBody::GeometryInfo::SideWallType::SWT_NX},
-            {"px", KinBody::GeometryInfo::SideWallType::SWT_PX},
-            {"ny", KinBody::GeometryInfo::SideWallType::SWT_NY},
-            {"py", KinBody::GeometryInfo::SideWallType::SWT_PY},
-        };
-
-        if (sideWallTypeMapping.find(type) == sideWallTypeMapping.end()) {
-            throw OPENRAVE_EXCEPTION_FORMAT(_("unrecognized sidewall type %s for loading from json"), type, ORE_InvalidArguments);
+        int sideWallType = 0;
+        if (v.HasMember("type") && v["type"].IsNumber()) {
```
ziyan, line 567: Should we use `IsInt()` instead?

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,25 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
+            orjson::LoadJsonValueByKey(v, "type", sideWallType);
```
ziyan, line 568: You need to check for range of `sideWallType` after loading the integer.

ziyan:
@rdiankov this PR and https://github.com/rdiankov/openrave/pull/846 contain fixes for various test failures and back-compat issues.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,33 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
-            throw OPENRAVE_EXCEPTION_FORMAT(_("unrecognized sidewall type %s for loading from json"), type, ORE_InvalidArguments);
+        if(!v.HasMember("type")) {
+            throw OPENRAVE_EXCEPTION_FORMAT0(_("need sidewall type when loading from json"), ORE_InvalidArguments);
+        }
+        int sideWallType = 0;
+        if (v["type"].IsInt()) {
+            // backward compatibility, support enum sideWall type
+            orjson::LoadJsonValueByKey(v, "type", sideWallType);
+            if (!(sideWallType >= KinBody::GeometryInfo::SideWallType::SWT_NX
+                && sideWallType <= KinBody::GeometryInfo::SideWallType::SWT_PY)) {
```
ziyan, line 574: Usually, I would do:

```c++
        enum SideWallType
        {
            SWT_NX=0,
            SWT_PX=1,
            SWT_NY=2,
            SWT_PY=3,
            SWT_First=SWT_NX,
            SWT_Last=SWT_PY,
        };
```

And test for `SWT_First` and `SWT_Last`, so that when someone change the header to add more types, they will not forget this place.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,33 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
+            orjson::LoadJsonValueByKey(v, "type", sideWallType);
+            if (!(sideWallType >= KinBody::GeometryInfo::SideWallType::SWT_First
+                && sideWallType <= KinBody::GeometryInfo::SideWallType::SWT_Last)) {
+                throw OPENRAVE_EXCEPTION_FORMAT(_("unrecognized sidewall type enum range %d for loading from json"), sideWallType, ORE_InvalidArguments);
+            }
+        }
+        else {
+            std::string type = "";
+            orjson::LoadJsonValueByKey(v, "type", type);
+            std::map<std::string, KinBody::GeometryInfo::SideWallType> sideWallTypeMapping = {
```
rdiankov, line 581: @bucketzxm having std::map initialized everytime is slow for no reason. having a `switch` statement is better/faster.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,33 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
-        std::map<std::string, KinBody::GeometryInfo::SideWallType> sideWallTypeMapping = {
-            {"nx", KinBody::GeometryInfo::SideWallType::SWT_NX},
-            {"px", KinBody::GeometryInfo::SideWallType::SWT_PX},
-            {"ny", KinBody::GeometryInfo::SideWallType::SWT_NY},
-            {"py", KinBody::GeometryInfo::SideWallType::SWT_PY},
-        };
-
-        if (sideWallTypeMapping.find(type) == sideWallTypeMapping.end()) {
-            throw OPENRAVE_EXCEPTION_FORMAT(_("unrecognized sidewall type %s for loading from json"), type, ORE_InvalidArguments);
+        if(!v.HasMember("type")) {
```
rdiankov, line 566: @bucketzxm if `type` does not exist, just don't initialize the geometry. perhaps we have geometries like this under normal flow when a user is first creating the geometry. i think throwing exception is too strict...

rdiankov: Or just give a default `NX` wall.

src/libopenrave/kinbodygeometry.cpp
bucketzxm: This is a string mapping .. which can't do `switch` i guess ?

src/libopenrave/kinbodygeometry.cpp
bucketzxm: set default  to `NX`

src/libopenrave/kinbodygeometry.cpp
ziyan: You can use `if` and `==`, probably still faster.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,35 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
+            RAVELOG_WARN(_("sideWallType is missing in json, set to 'nx'"));
+            t.type = KinBody::GeometryInfo::SideWallType::SWT_NX;
+            return;
+        }
+        int sideWallType = 0;
```
ziyan, line 571: This belongs inside `if (v["type"].IsInt()) {`.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,35 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
+            std::map<std::string, KinBody::GeometryInfo::SideWallType> sideWallTypeMapping = {
+                {"nx", KinBody::GeometryInfo::SideWallType::SWT_NX},
+                {"px", KinBody::GeometryInfo::SideWallType::SWT_PX},
+                {"ny", KinBody::GeometryInfo::SideWallType::SWT_NY},
+                {"py", KinBody::GeometryInfo::SideWallType::SWT_PY},
+            };
+            if (sideWallTypeMapping.find(type) == sideWallTypeMapping.end()) {
+                throw OPENRAVE_EXCEPTION_FORMAT(_("unrecognized sidewall type %s for loading from json"), type, ORE_InvalidArguments);
+            }
+            sideWallType = sideWallTypeMapping[type];
```
ziyan, line 592: Set `t.type` directly, since you have the enum type here, no need to change to `int` and back`

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,35 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
-            {"nx", KinBody::GeometryInfo::SideWallType::SWT_NX},
-            {"px", KinBody::GeometryInfo::SideWallType::SWT_PX},
-            {"ny", KinBody::GeometryInfo::SideWallType::SWT_NY},
-            {"py", KinBody::GeometryInfo::SideWallType::SWT_PY},
-        };
-
-        if (sideWallTypeMapping.find(type) == sideWallTypeMapping.end()) {
-            throw OPENRAVE_EXCEPTION_FORMAT(_("unrecognized sidewall type %s for loading from json"), type, ORE_InvalidArguments);
+        if(!v.HasMember("type")) {
+            RAVELOG_WARN(_("sideWallType is missing in json, set to 'nx'"));
```
ziyan, line 567: Don't translate log.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -563,19 +563,35 @@ inline void LoadJsonValue(const rapidjson::Value& v, KinBody::GeometryInfo::Side
...
+            t.type = KinBody::GeometryInfo::SideWallType::SWT_NX;
+            return;
```
ziyan, line 569: Don't do this early return business here. If someone adds more code later, they will miss this.

rdiankov:
thanks~

PR #860, opened 2020-08-05
Add envid in logs

Puttichai:
Added environment id in logs from `SetDOFValues` and `SetDOFVelocities` to help figure out more easily what happens in which environment


rdiankov:
thanks~

PR #861, opened 2020-08-05
Removed submodule plugins/broadcast

bucketzxm:


ziyan:
@bucketzxm please be careful

PR #862, opened 2020-08-05
Fix deserialization of IkParameterization

yoshikikanemoto:
it tried to deserialize `customData` by using keyword `key` whereas serialization uses keyword `id`.


rdiankov:
nice catch, thanks! (FYI @bucketzxm )

PR #865, opened 2020-08-06
When viewer is destroyed, also destroy qapplication to close X11 conn…

ziyan:
…ection.

### DIFF
--- a/plugins/qtosgrave/qtosgrave.cpp
+++ b/plugins/qtosgrave/qtosgrave.cpp
@@ -16,9 +16,6 @@
 
 #include "qtosg.h"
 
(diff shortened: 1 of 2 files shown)


rdiankov:
@ziyan code that you erased was for attaching the openrave viewer to an existing QApplication (ie if Qt is created from python and managed there). Are you sure you want to remove that code?

ziyan:
@rdiankov I moved the uncommented code to `CreateQtOSGViewer`. It should be equivalent.

rdiankov:
sure, i guess we can always resurrect when needed again. thanks

PR #866, opened 2020-08-06
Need to clean up all the user data set on viewer destroy.

ziyan:


plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -195,6 +195,13 @@ QtOSGViewer::~QtOSGViewer()
     }
 
     _condUpdateModels.notify_all();
+
+    // have to remove all the _userdatakey set on bodies in the env before we go
+    FOREACH(itbody, _mapbodies) {
+        BOOST_ASSERT( itbody->first->GetUserData(_userdatakey) == itbody->second );
```
rdiankov, line 201: @ziyan let's have warning instead of assert.

ziyan: Ok done.

rdiankov:
thanks~

PR #868, opened 2020-08-07
better writing of closed loop kinematics

kanbouchou:
previously, same joint was written twice for body with closed loop which later led to error from colladareader

@tgn3000 @AntoineRyu


tgn3000:
@kanbouchou I refactored a litttttttttttttle bit. Hope you don't mind.

rdiankov:
thanks~

PR #870, opened 2020-08-11
do not write geometry twice for closed loop links

kanbouchou:
follow up after https://github.com/rdiankov/openrave/pull/868 which fixed writing same joint twice, 

This patch fixes writing same geometry twice for links who has multiple parents due to closed loop.

Before this patch, ``env.Save`` would double number of geometries for links with multiple parents.

@AntoineRyu


rdiankov:
thanks~

PR #871, opened 2020-08-12
Change map format in json.

ziyan:


src/libopenrave/kinbodyjoint.cpp
```diff
@@ -344,23 +344,35 @@ void KinBody::JointInfo::DeserializeJSON(const rapidjson::Value& value, dReal fU
         _mapFloatParameters.clear();
         for (rapidjson::Value::ConstValueIterator it = value["floatParameters"].Begin(); it != value["floatParameters"].End(); ++it) {
             std::string key;
-            orjson::LoadJsonValueByKey(*it, "key", key);
+            orjson::LoadJsonValueByKey(*it, "id", key);
```
rdiankov, line 347: @ziyan why not the following?

```c++
if( it->HasMember("id" ) {
    orjson::LoadJsonValueByKey(*it, "id", key);
}
else if( it->HasMember("key" ) {
    orjson::LoadJsonValueByKey(*it, "key", key);
}
else {
    continue:
}
```

ziyan: Okay, I added these checks and log statement.

rdiankov:
thanks~

PR #877, opened 2020-08-17
should specify checking options to check all constraints in _InitPlan…

Puttichai:
Explicitly specify constraint options `0xffff` when checking the initial configuration(s) of RRT.


rdiankov:
thanks~

PR #878, opened 2020-08-17
Fixes for webstack2

ziyan:
@bucketzxm Please make sure all these changes are covered in test, especially the ikparam related changes.


rdiankov:
@ziyan @bucketzxm Each KinBody has `_environmentid` which is unique, so why do you need `_ResolveBodyIds`?

rdiankov:
@ziyan @bucketzxm  still need a better way to manage body ids and how they related to kinbody's environment id, but for now, number of fixes are greater than problems, so let's merge. thanks~

PR #879, opened 2020-08-18
Add support for Fparser installed in system

strmio:
This patch modifies `CMakeLists.txt` to use `fparser` installed in system if available. If `fparser` is not installed in system, the local copy of `fparser` under the `3rdparty` directory will be used as before.


rdiankov:
thanks~

PR #881, opened 2020-08-26
Include header dir for rapidjson

PeterBowman:
Enables custom installation paths (other than `/usr` or `/usr/local`).


rdiankov:
thanks~

PR #882, opened 2020-09-01
Tell OpenRAVE downstreams to use path to RapidJSON headers

PeterBowman:
This is a follow-up to https://github.com/rdiankov/openrave/pull/881 (sorry for not having noticed this earlier). Since RapidJSON is transitively included from within OpenRAVE headers, projects that compile against OR must be able to find them as well. It is troublesome in case RapidJSON is installed to a non-standard location (neither `/usr/include` nor `/usr/local/include`), hence this patch.


rdiankov:
nice catch, thanks!

PR #883, opened 2020-09-01
Add new variable "restrictGraspSetNames" to Manipulator

strmio:
To store information about applicability of grasp sets on manipulators, this patch adds a new field `restrictGraspSetNames` and its corresponding accessors to `ManipulatorInfo` and `Manipulator`. This variable is useful under a multi gripper configuration.

```
std::vector<std::string> _vRestrictGraspSetNames; ///< When this parameter is non-empty, only grasp sets listed here are applicable for this manipulator.
```


rdiankov:
thanks~

PR #885, opened 2020-09-08
Initialize mass.

ziyan:


kanbouchou:
@ziyan @rdiankov 

Maybe we should use small but non-zero value like 1e-10 for default?
In other places, 

0 at https://github.com/rdiankov/openrave/blob/master/src/libopenrave/kinbodylink.cpp#L31
0 at https://github.com/rdiankov/openrave/blob/master/python/bindings/include/openravepy/openravepy_jointinfo.h#L93
1e-10 at https://github.com/rdiankov/openrave/blob/master/src/libopenrave-core/colladaparser/colladareader.cpp#L1801

rdiankov:
thanks~

PR #890, opened 2020-09-16
Fix updating environment from JSON while some body is being grabbed

ziyan:
The background of the problem is:

1. OpenRAVE env has robot grabbing a body
2. WebStack does not know about this grabbing state since it is dynamic and local to the planning process
3. WebStack API is used to remove the body, environment change notification is sent
4. After removing the body from OpenRAVE env, we tried to restore grabbing state, which is no longer valid, and throws exception

This fix is to give warning when this case happens instead of exception.

cc @kanbouchou


ziyan:
This case is now covered in testopenrave.

rdiankov:
thanks~

PR #891, opened 2020-09-21
Calibration board geometry type, visualization, and serialization for openrave viewer

Hsifnus:
Support for calibration board geometry.

Generate dot mesh for board automatically according to parameters. Serialized collada file will not contain the dots on the other hand.

Sample collada file can be found at `test/testdata/calibration_board.dae`

image

### DIFF
--- a/include/openrave/kinbody.h
+++ b/include/openrave/kinbody.h
@@ -44,6 +44,7 @@ enum GeometryType {
     GT_TriMesh = 4,
     GT_Container=5, ///< a container shaped geometry that has inner and outer extents. container opens on +Z. The origin is at the bottom of the base.
     GT_Cage=6, ///< a container shaped geometry with removable side walls. The side walls can be on any of the four sides. The origin is at the bottom of the base. The inner volume of the cage is measured from the base to the highest wall.
+    GT_CalibrationBoard=7, ///< a box shaped geometry with grid of cylindrical dots of two sizes. The dots are always on the +z side of the box and are oriented towards z-axis.
 };
 
 enum DynamicsConstraintsType {
--- a/include/openrave/kinbody.h
+++ b/include/openrave/kinbody.h
@@ -332,21 +332,21 @@ public:
         bool _bVisible = true; ///< if true, geometry is visible as part of the 3d model (default is true)
         bool _bModifiable = true; ///< if true, object geometry can be dynamically modified (default is true)
 
-        struct CalibrationBoardParams { ///< used by GT_CalibrationBoard
+        struct CalibrationBoardParameters { ///< used by GT_CalibrationBoard
             int numDotsX; ///< number of dots in x direction, minimum 3
             int numDotsY; ///< number of dots in y direction, minimum 3
-            float dotsDistanceX; ///< distance between center of dots in x direction
-            float dotsDistanceY; ///< distance between center of dots in y direction
-            RaveVector<float> dotColor; ///< color of dot mesh, which can differ from the color of the board mesh
+            dReal dotsDistanceX; ///< distance between center of dots in x direction
+            dReal dotsDistanceY; ///< distance between center of dots in y direction
+            RaveVector<dReal> dotColor; ///< color of dot mesh, which can differ from the color of the board mesh
             std::string patternName; ///< string that determines pattern of big and normal dots
                                      /// currently, the only supported type is "threeBigDotsDotGrid"
                                      /// any unsupported pattern will generate a full grid of normal dots
-            float dotDiameterDistanceRatio; ///< dot diameter divided by minimum of dot distances x and y
-            float bigDotDiameterDistanceRatio; ///< big dot diameter divided by minimum of dot distances x and y
+            dReal dotDiameterDistanceRatio; ///< dot diameter divided by minimum of dot distances x and y
+            dReal bigDotDiameterDistanceRatio; ///< big dot diameter divided by minimum of dot distances x and y
             ///< trimesh representation of the collision data of calibration board dot grid in this local coordinate system
             /// Shares the same caveats as _meshcollision.
             TriMesh _dotmeshcollision;
-            bool operator==(const CalibrationBoardParams& other) const {
+            bool operator==(const CalibrationBoardParameters& other) const {
                 return numDotsX == other.numDotsX
                        && numDotsY == other.numDotsY
                        && dotsDistanceX - other.dotsDistanceX < dotsDistanceX*g_fEpsilon
@@ -356,11 +356,11 @@ public:
                        && dotDiameterDistanceRatio - other.dotDiameterDistanceRatio < dotDiameterDistanceRatio*g_fEpsilon
                        && bigDotDiameterDistanceRatio - other.bigDotDiameterDistanceRatio < bigDotDiameterDistanceRatio*g_fEpsilon;
             }
-            bool operator!=(const CalibrationBoardParams& other) const {
+            bool operator!=(const CalibrationBoardParameters& other) const {
                 return !operator==(other);
             }
         };
-        CalibrationBoardParams _calibrationBoardParams;
+        CalibrationBoardParameters _calibrationBoardParams;
 
         /// \brief Generates the calibration board's dot grid mesh based on calibration board settings
         void GenerateCalibrationBoardDotMesh(float fTessellation=1);
@@ -607,7 +607,7 @@ public:
                 return _info._calibrationBoardParams._dotmeshcollision;
             }
             /// \brief returns the color of the calibration board's dot mesh
-            inline const RaveVector<float>& GetCalibrationBoardDotColor() const {
+            inline const RaveVector<dReal>& GetCalibrationBoardDotColor() const {
                 return _info._calibrationBoardParams.dotColor;
             }
             /// \brief returns x dimension (in dots) of the calibration board dot grid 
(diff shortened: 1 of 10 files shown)


include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+                       && dotColor == other.dotColor
+                       && patternName == other.patternName
+                       && dotDiameterDistanceRatio - other.dotDiameterDistanceRatio < dotDiameterDistanceRatio*g_fEpsilon
+                       && bigDotDiameterDistanceRatio - other.bigDotDiameterDistanceRatio < bigDotDiameterDistanceRatio*g_fEpsilon;
+            }
+            bool operator!=(const CalibrationBoardParams& other) const {
+                return !operator==(other);
+            }
+        };
+        CalibrationBoardParams _calibrationBoardParams;
```
ziyan, line 363: `Params` -> `Parameters`

include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
         bool _bVisible = true; ///< if true, geometry is visible as part of the 3d model (default is true)
         bool _bModifiable = true; ///< if true, object geometry can be dynamically modified (default is true)
 
+        struct CalibrationBoardParams { ///< used by GT_CalibrationBoard
+            int numDotsX; ///< number of dots in x direction, minimum 3
+            int numDotsY; ///< number of dots in y direction, minimum 3
+            float dotsDistanceX; ///< distance between center of dots in x direction
```
ziyan, line 338: Is there a particular reason why `float` is used instead of `dReal`? Same for `RaveVector`

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -600,6 +666,32 @@ inline void LoadJsonValue(const rapidjson::Value& rValue, KinBody::GeometryInfo:
...
+    }
+    const char *calibrationBoardParamNames[] = {
+        "numDotsX", "numDotsY", "dotsDistanceX",
+        "dotsDistanceY", "dotColor", "patternName", 
+        "dotDiameterDistanceRatio", "bigDotDiameterDistanceRatio"
+    };
+    for (int idx = 0; idx < 8; idx++) {
+        if (!rValue.HasMember(calibrationBoardParamNames[idx])) {
+            char paramError[1024];
+            sprintf(paramError, "Missing calibration board parameter \"%s\".", calibrationBoardParamNames[idx]);
```
braineo, line 681: Use `RAVELOG_ERROR_FORMAT` instead of `sprintf`

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -1131,6 +1272,12 @@ AABB KinBody::GeometryInfo::ComputeAABB(const Transform& tGeometryWorld) const
         }
         break;
     }
+    case GT_CalibrationBoard: // the tangible part of the board is basically the box
```
braineo, line 1275: Looks as same as `GT_Box`? `case` earlier?

braineo:
@Hsifnus Looks good in general. While you cannot generate dot patterns, worth to log it out using `RAVELOG_WARN`

Hsifnus:
A `RAVELOG_WARN` message is now sent out when dot pattern generation fails.

braineo (approved):

include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
         bool _bVisible = true; ///< if true, geometry is visible as part of the 3d model (default is true)
         bool _bModifiable = true; ///< if true, object geometry can be dynamically modified (default is true)
 
+        struct CalibrationBoardParameters { ///< used by GT_CalibrationBoard
+            int numDotsX; ///< number of dots in x direction, minimum 3
+            int numDotsY; ///< number of dots in y direction, minimum 3
+            dReal dotsDistanceX; ///< distance between center of dots in x direction
+            dReal dotsDistanceY; ///< distance between center of dots in y direction
+            RaveVector<dReal> dotColor; ///< color of dot mesh, which can differ from the color of the board mesh
```
rdiankov, line 340: Although not wrong, convention is `RaveVector<dReal> dotColor;  -> float`

include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
         bool _bVisible = true; ///< if true, geometry is visible as part of the 3d model (default is true)
         bool _bModifiable = true; ///< if true, object geometry can be dynamically modified (default is true)
 
+        struct CalibrationBoardParameters { ///< used by GT_CalibrationBoard
+            int numDotsX; ///< number of dots in x direction, minimum 3
```
rdiankov, line 336: - need constructor to set parameter initials, are have inline initializers.

include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+                                     /// any unsupported pattern will generate a full grid of normal dots
+            dReal dotDiameterDistanceRatio; ///< dot diameter divided by minimum of dot distances x and y
+            dReal bigDotDiameterDistanceRatio; ///< big dot diameter divided by minimum of dot distances x and y
+            ///< trimesh representation of the collision data of calibration board dot grid in this local coordinate system
+            /// Shares the same caveats as _meshcollision.
+            TriMesh _dotmeshcollision;
+            bool operator==(const CalibrationBoardParameters& other) const {
+                return numDotsX == other.numDotsX
+                       && numDotsY == other.numDotsY
+                       && dotsDistanceX - other.dotsDistanceX < dotsDistanceX*g_fEpsilon
```
rdiankov, line 352: dotsDistanceX - other.dotsDistanceX < dotsDistanceX*g_fEpsilon -> use `==`

include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+            RaveVector<dReal> dotColor; ///< color of dot mesh, which can differ from the color of the board mesh
+            std::string patternName; ///< string that determines pattern of big and normal dots
+                                     /// currently, the only supported type is "threeBigDotsDotGrid"
+                                     /// any unsupported pattern will generate a full grid of normal dots
+            dReal dotDiameterDistanceRatio; ///< dot diameter divided by minimum of dot distances x and y
+            dReal bigDotDiameterDistanceRatio; ///< big dot diameter divided by minimum of dot distances x and y
+            ///< trimesh representation of the collision data of calibration board dot grid in this local coordinate system
+            /// Shares the same caveats as _meshcollision.
+            TriMesh _dotmeshcollision;
+            bool operator==(const CalibrationBoardParameters& other) const {
```
rdiankov, line 349: Also create `Compare` fn with epsilon, use abs(x-y) <= epsilon

include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+            dReal dotsDistanceY; ///< distance between center of dots in y direction
+            RaveVector<dReal> dotColor; ///< color of dot mesh, which can differ from the color of the board mesh
+            std::string patternName; ///< string that determines pattern of big and normal dots
+                                     /// currently, the only supported type is "threeBigDotsDotGrid"
+                                     /// any unsupported pattern will generate a full grid of normal dots
+            dReal dotDiameterDistanceRatio; ///< dot diameter divided by minimum of dot distances x and y
+            dReal bigDotDiameterDistanceRatio; ///< big dot diameter divided by minimum of dot distances x and y
+            ///< trimesh representation of the collision data of calibration board dot grid in this local coordinate system
+            /// Shares the same caveats as _meshcollision.
+            TriMesh _dotmeshcollision;
```
rdiankov, line 348: - `TriMesh _dotmeshcollision;`  -> not sure why necessary unless somehow vision algorithms use this data...? if it is just for rendering, then you don't need it.

include/openrave/kinbody.h
```diff
@@ -330,6 +332,39 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+                       && dotColor == other.dotColor
+                       && patternName == other.patternName
+                       && dotDiameterDistanceRatio - other.dotDiameterDistanceRatio < dotDiameterDistanceRatio*g_fEpsilon
+                       && bigDotDiameterDistanceRatio - other.bigDotDiameterDistanceRatio < bigDotDiameterDistanceRatio*g_fEpsilon;
+            }
+            bool operator!=(const CalibrationBoardParameters& other) const {
+                return !operator==(other);
+            }
+        };
+        CalibrationBoardParameters _calibrationBoardParams;
```
rdiankov, line 363: - Unfortunately the current structure will add calibration parameters to every GeometryInfo, even ones that do not have it. If you can encode this in a way that would add minimal size to GeometryInfo, perhaps it is OK (For example encoding most of the parameters inside `_vGeomDataX`. I do think having `CalibrationBoardParameters` as the way you define it is OK, but the way we insert it into GeometryInfo could be better. For example, we can make it a pointer (but then it becomes tricky, one interesting way out of this is to use `std::vector<CalibrationBoardParameters>` and resize it to 1 whenever the geomtype is `GT_CalibrationBoard`.

braineo: Something like `_vSideWalls` I guess?

```cpp
std::vector<SideWall> _vSideWalls; ///< used by GT_Cage
```

python/bindings/openravepy_kinbody.cpp
```diff
@@ -4474,6 +4513,7 @@ void init_openravepy_kinbody()
                           .def_readwrite("_bVisible",&PyGeometryInfo::_bVisible)
                           .def_readwrite("_bModifiable",&PyGeometryInfo::_bModifiable)
                           .def_readwrite("_vSideWalls", &PyGeometryInfo::_vSideWalls)
+                          .def_readwrite("_calibrationBoardParams", &PyGeometryInfo::_calibrationBoardParams)
```
ziyan, line 4516: `Params` -> `Parameters`

braineo:
still have both`_calibrationBoardParams` and `_calibrationBoardParameters`

rdiankov:
@Hsifnus there was a lot of small issues, so I just added `calibration-board-geom` branch to the origin and edited it. In the future, please create branches/pull-requests to `rdiankov/openrave` so that others can edit. thanks~

PR #894, opened 2020-09-23
Fix robot configuration sampler

Puttichai:
Disable the sampler's active space tracking by default. When the sampler is used as in a PlannerParameters' sample function, it should not automatically update its own robot's lower/upper limits.


CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set( CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS TRUE )
 
 # Define here the needed parameters
 set (OPENRAVE_VERSION_MAJOR 0)
-set (OPENRAVE_VERSION_MINOR 58)
+set (OPENRAVE_VERSION_MINOR 59)
```
rdiankov, line 7: @Puttichai it's not clear why you are changing minor version...

Puttichai: Following our discussion, I reverted the change of the minor version and increased the patch version instead.

rdiankov:
thanks~

PR #895, opened 2020-09-24
Fix to openrave viewer GUI thread update truncation

Hsifnus:
Fixes issue in openrave viewer where GUI thread update functions would be ignored if more than 1000 were to be created and processed at the same time.

### DIFF
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -473,7 +473,7 @@ void QtOSGViewer::_UpdateViewerCallback()
         if (!_excessGUIFunctions.empty()) {
             boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
             int excessSize = _excessGUIFunctions.size();
-            for (int idx = 0; idx < excessSize && idx < 1000; idx++) {
+            for (int idx = 0; idx < excessSize && _listGUIFunctions.size() < 1000; idx++) {
                 _listGUIFunctions.push_back(_excessGUIFunctions.front());
                 _excessGUIFunctions.pop_front();
             }
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -472,9 +472,9 @@ void QtOSGViewer::_UpdateViewerCallback()
         }
         if (!_listGUIFunctionsBuffer.empty()) {
             boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
-            auto insertPos = _listGUIFunctions.end();
-            auto bufferStartPos = _listGUIFunctionsBuffer.begin();
-            auto bufferEndPos = _listGUIFunctionsBuffer.begin();
+            std::list<GUIThreadFunctionPtr>::iterator insertPos = _listGUIFunctions.end();
+            std::list<GUIThreadFunctionPtr>::iterator bufferStartPos = _listGUIFunctionsBuffer.begin();
+            std::list<GUIThreadFunctionPtr>::iterator bufferEndPos = _listGUIFunctionsBuffer.begin();
             int itemsToAdd = 1000 < _listGUIFunctions.size() ? 0 : 1000 - _listGUIFunctions.size();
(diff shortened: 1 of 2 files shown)


plugins/qtosgrave/qtosgviewer.h
```diff
@@ -396,6 +396,7 @@ public slots:
 
     //@{ Message Queue
     list<GUIThreadFunctionPtr> _listGUIFunctions; ///< list of GUI functions that should be called in the viewer update thread. protected by _mutexGUIFunctions
+    list<GUIThreadFunctionPtr> _excessGUIFunctions; ///< list of GUI functions that exceed the queue limit and are thus stashed.
```
braineo, line 399: `_excessGUIFunctions` => `_listGUIFunctionsBuffer`

in openrave you see variables like `vxxx` or `bxxx`, they are vector and boolean.

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -470,6 +470,14 @@ void QtOSGViewer::_UpdateViewerCallback()
                 }
             }
         }
+        if (!_excessGUIFunctions.empty()) {
```
braineo, line 473: I would make here the only place to modify `listGUIFunctions` use [splice](https://en.cppreference.com/w/cpp/container/list/splice) instead of pushing to 2 different lists judging by size.

braineo (commented):
Reading from code, the problem of following seems not addressed yet?

1. Creating a handle
2. call `handle.Close()` before processed in GUI thread
3. handle cannot be closed anymore

cv3d:
@Hsifnus Thanks for your efforts.
I tested it on large number of updates, and it worked as expected.

Please address the comments from @braineo first, and then let's get it merged.

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2131,7 +2133,7 @@ void QtOSGViewer::_PostToGUIThread(const boost::function<void()>& fn, bool block
     GUIThreadFunctionPtr pfn(new GUIThreadFunction(fn, block));
     if( _listGUIFunctions.size() > 1000 ) {
```
braineo, line 2134: So that we can always push to `_listGUIFunctionsBuffer` without checking the size

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -470,6 +470,16 @@ void QtOSGViewer::_UpdateViewerCallback()
                 }
             }
         }
+        if (!_listGUIFunctionsBuffer.empty()) {
+            boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
+            auto insertPos = _listGUIFunctions.begin();
+            auto bufferStartPos = _listGUIFunctionsBuffer.begin();
+            auto bufferEndPos = _listGUIFunctionsBuffer.begin();
```
braineo, line 477: Want something different from `_listGUIFunctionsBuffer.end()`?

braineo: Follow code style using `std::list<T>::iterator` instead of `auto`

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2120,13 +2129,9 @@ void QtOSGViewer::_PostToGUIThread(const boost::function<void()>& fn, bool block
...
     boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
-    if( _listGUIFunctions.size() > 1000 ) {
-        // can happen if system is especially slow
-        //RAVELOG_WARN_FORMAT("too many gui post commands, ignoring %d", _listGUIFunctions.size());
-        return;
-    }
     GUIThreadFunctionPtr pfn(new GUIThreadFunction(fn, block));
-    _listGUIFunctions.push_back(pfn);
+    // pushing to buffer directly is thread-safe
+    _listGUIFunctionsBuffer.push_back(pfn);
```
rdiankov, line 2134: @Hsifnus isn't this infinitely growing if viewer thread is not running?

braineo:
close in favor of #925

PR #896, opened 2020-09-24
fix setting camera transformation

bucketzxm:
how  to reproduce: 
```viewer = env.GetViewer(); viewer.SetCamera(viewer.GetCameraTransform(), viewer.GetCameraDistanceToFocus())```

### DIFF
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -1370,16 +1370,15 @@ bool QtOSGViewer::WriteCameraImage(int width, int height, const RaveTransform<fl
     return false;
 }
 
-void QtOSGViewer::_SetCameraTransform()
+void QtOSGViewer::_SetCameraTransform(const RaveTransform<float>& transform)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugins/qtosgrave/qtosgviewer.cpp [hunk 1]
```diff
@@ -1441,7 +1441,7 @@ RaveTransform<float> QtOSGViewer::GetCameraTransform() const
     boost::mutex::scoped_lock lock(_mutexGUIFunctions);
```
braineo, line 1441: Probably no need for this lock?

braineo:
Actually I would remove `_Tcamera`, and let `void QtOSGViewer::_SetCameraTransform()` take a transform.

plugins/qtosgrave/qtosgviewer.cpp
bucketzxm: other gui functions could change the camera transformation .. ?

braineo (approved):
LGTM

PR #898, opened 2020-09-29
Fixed parabolicsmoothers

Puttichai:
`parabolicsmoother`: fixed a bug that skips adding checked ramps into the resulting shortcut segment
`parabolicsmoother2`: fixed a bug that causes discrepancies in the initial joint values and velocities of shortcut segments


rdiankov:
thanks~

PR #904, opened 2020-10-13
Fix infinite loop when calling _ConvertIntToHex

tgn3000:
strlen["link"]=4, sizeof["link"]=5

See 
https://www.geeksforgeeks.org/difference-strlen-sizeof-string-c-reviewed/#:~:text=sizeof()%20operator%20returns%20actual,total%20size%20is%209%20bytes.


rdiankov:
thanks~

PR #907, opened 2020-10-21
Fix hash computation

Puttichai:
- Now serializing the rotational part of a pose by converting it to the rotation matrix representation first. Previously quaternion representation was used, leading to a problem as a quaternion `v` is equivalent to `-v`.
- In addition, when serializing a rotation matrix, only consider two columns as the third one can always be uniquely determined from the other two, to help save time.
- When serializing the translational part of a pose, consider only the first three values of the vector, instead of 4 (by using `SerializeRound3` instead of `SerializeRound`).
- Renamed `SerializeRound(std::ostream& o, const RaveVector<T>& v)` to `SerializeRoundQuaternion(std::ostream& o, const RaveVector<T>& v)` for clarity.
- Now clearing `__hashrobotstructure` when related properties change.

Thanks @kanbouchou  for the discussion.

### DIFF
--- a/src/libopenrave/libopenrave.h
+++ b/src/libopenrave/libopenrave.h
@@ -204,7 +204,7 @@ template <class T>
 inline void SerializeRound4(std::ostream& o, const RaveVector<T>& v)
 {
     // This function is used only for serializing quaternions. Need to
-    // take into account the face that v and -v represent the same
+    // take into account the fact that v and -v represent the same
     // rotation. Convert v to a rotation matrix instead to get a
     // unique representation. Then since the thrid column can be uniquely
     // determined given the first two, serializing only the first two
--- a/src/libopenrave/libopenrave.h
+++ b/src/libopenrave/libopenrave.h
@@ -230,9 +230,11 @@ inline void SerializeRound(std::ostream& o, const RaveTransform<T>& t)
 template <class T>
 inline void SerializeRound(std::ostream& o, const RaveTransformMatrix<T>& t)
 {
+    // Since the thrid column of the rotation matrix can be uniquely
+    // determined given the first two, serializing only the first two
+    // columns is sufficient for the purpose of hash computation.
     o << SerializationValue(t.m[0]) << " " << SerializationValue(t.m[4]) << " " << SerializationValue(t.m[8]) << " "
-      << SerializationValue(t.m[1]) << " " << SerializationValue(t.m[5]) << " " << SerializationValue(t.m[9]) << " "
-      << SerializationValue(t.m[2]) << " " << SerializationValue(t.m[6]) << " " << SerializationValue(t.m[10]) << " ";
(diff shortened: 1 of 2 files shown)


src/libopenrave/libopenrave.h
```diff
@@ -201,9 +201,17 @@ inline void SerializeRound(std::ostream& o, double f)
...
-    o << SerializationValue(v.x) << " " << SerializationValue(v.y) << " " << SerializationValue(v.z) << " " << SerializationValue(v.w) << " ";
+    // This function is used only for serializing quaternions. Need to
+    // take into account the fact that v and -v represent the same
+    // rotation. Convert v to a rotation matrix instead to get a
+    // unique representation. Then since the thrid column can be uniquely
+    // determined given the first two, serializing only the first two
+    // columns is sufficient for the purpose of hash computation.
+    RaveTransformMatrix<T> t = matrixFromQuat(v);
+    o << SerializationValue(t.m[0]) << " " << SerializationValue(t.m[4]) << " " << SerializationValue(t.m[8]) << " "
+      << SerializationValue(t.m[1]) << " " << SerializationValue(t.m[5]) << " " << SerializationValue(t.m[9]) << " ";
```
tgn3000, line 214: directly call `SerializeRound(o, matrixFromQuat(v))` as you've refactored out

tgn3000:
I think it's easier to change from the original
```cpp
inline void SerializeRound(...) {
    Vector v = t.rot;
    for(int i = 0; i < 4; ++i) { ... }
    ...
}
```
to
```cpp
inline void SerializeRound(...) {
    const Vector v = t.rot[0]>=0 ? t.rot : -t.rot;
    SerializeRound(o, v);	
    SerializeRound(o, t.trans);
}
```
Did I miss something?

src/libopenrave/libopenrave.h
Puttichai: `SerializeRound(std::ostream& o, const RaveTransformMatrix<T>& t)` serializes both rotational and translation parts.

Puttichai:
First, this statement `t.rot.x >= 0 ? t.rot : -t.rot` is prone to epsilon errors. You will get different results when serializing `[-1e-12, 0, 1.0, 0]` and `[1e-12, 0, 1.0, 0]`.
We can get around that by introducing some threshold to decide which numbers are zero. But then we can also just use the matrix representation, which is unique, instead.

tgn3000:
I see. I start to think whether this way of  "serialize floating-point numbers and compute hash" is a good idea throughout the repo now.

Calculations in matrixFromQuat is not very cheap. Hope it does not get called a lot.

Puttichai:
Stringstream operations are actually more expensive than the conversion.

But whether the current way of computing hash is good or not, or how it can be improved can be a separate discussion.

rdiankov:
thanks guys~

PR #900, opened 2020-10-02
WIP: Math: Compute mimic joints' second-order partial derivatives, their second-order time derivatives, and link accelerations

tgn3000:
TBW.

PR #908, opened 2020-10-22
Fix error due to automatically adding first partial variable

AntoineRyu:
Fix error where the first partial equation is automatically added when the pos equation have a variable. When the pos equation evaluates to 0, loading the saved dae will assert since the variable in the first partial doesn't exit in the pos equation anymore and will fail to set mimic equations on all following joints.


tgn3000:
LGTM. This section of code will be changed in #900 as well.

AntoineRyu:
> LGTM. This section of code will be changed in #900 as well.

From the look of it #900 have the same issue that this is correcting, but pushes it further by forcing a second partial equation when passing empty string.

rdiankov:
sure, thanks~

tgn3000:
Will have that in mind. Thanks @AntoineRyu

PR #913, opened 2020-11-03
Fix parabolicsmoother

Puttichai:
In `Shortcut2` (`parabolicsmoother.cpp`), when checking if the final ramp of the shortcut segment is valid, have to make sure that we accept the shortcut segment only when `!retcheck.bDifferentVelocity`.

### DIFF
--- a/plugins/rplanners/parabolicsmoother.cpp
+++ b/plugins/rplanners/parabolicsmoother.cpp
@@ -2302,29 +2302,30 @@ protected:
                             _totalTimeCheckPathAllConstraints_SegmentFeasible2 = 0;
 #endif
 
-                            if (retcheck.retcode == 0) {
-                                // The final segment is now good.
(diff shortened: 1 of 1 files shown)


plugins/rplanners/parabolicsmoother.cpp
```diff
@@ -2579,7 +2580,7 @@ class ParabolicSmoother : public PlannerBase, public ParabolicRamp::FeasibilityC
                 }
                 dReal diff = dummyEndTime - endTime;
                 vVisitedDiscretization.clear(); // have to clear so that can recreate the visited nodes
-#ifdef SMOOTHER_PROGRESS_DEBUG
+#ifdef SMOOTHER1_PROGRESS_DEBUG
```
rdiankov, line 2583: @Puttichai is this on purpose?

rdiankov: nm, looks like it is

rdiankov:
nice catch, thanks~

PR #914, opened 2020-11-06
Fix pybind11 build

cielavenir:
again I was playing with buster...

by the way I'm looking forward to bullseye beta (if my build is succeeding there, backend team will be helped so much...)


tgn3000:
thanks!!! @cielavenir

PR #915, opened 2020-11-09
Fix connected body attached sensor

ziyan:
`pattachedlink` pointer was assigned in `AttachedSensor` constructor. But in `_ComputeConnectedBodiesInformation` we modify the link name after the `AttachedSensor` is constructed.

This fix delays resolving of `pattachedlink` pointer until `_ComputeInternalInformation`. Similar things are being done for `Manipulator` and other class.


rdiankov:
thanks~

PR #919, opened 2020-11-13
Fix issue with `AttachedSensorInfo._linkname` misisng

ziyan:
Introduced by #915.

The problem is collada reader and xml reader sets `pattachedlink` without setting `info._linkname`. When `_ComputeInternalInformation` is called, `pattachedlink` gets set to null due to `_linkname` being empty.

This change makes sure `info._linkname` is set in collada reader and xml reader. This is consistent with how these 2 readers handle `Manipulator`.


src/libopenrave-core/colladaparser/colladareader.cpp
```diff
@@ -1056,11 +1056,17 @@ class ColladaReader : public daeErrorHandler
             if( !!pinterface_type ) {
                 if( pinterface_type->type == "kinbody" ) {
                     pbody = RaveCreateKinBody(_penv,pinterface_type->name);
+                    if (!pbody) {
```
ziyan, line 1059: These 2 new `if`s are requested by Osbert and Guangyang. They have members who cannot access the internal plugins, so cannot load vendor specific robot interface type. They want to in that case load the default robot class so they can use OpenRAVE to continue to edit scenes.

rdiankov: @ziyan can you give a warning message then?

ziyan: Done.

rdiankov:
thanks~

PR #920, opened 2020-11-19
Add workspace jitterer

Puttichai:
### Summary

Given a robot configuration, the current configuration jitterer produces a nearby configuration by adding small perturbations to each joint value independently. In some scenarios, however, it could more beneficial that a new configuration is produced so that the tool is moved from the initial position in some certain directions in the workspace. For those cases, jittering based on tool position in workspace could help find a feasible robot configuration more easily.

This pull request introduces a workspace jitterer. Given a robot configuration, a nearby configuration is produced by adding small perturbations to joint values so that the tool is moved along one of the predefined directions in its coordinate frame.


rdiankov:
thanks~

PR #921, opened 2020-11-19
wip: fix updating of link info

kanbouchou:
some information was not updated properly.

@bucketzxm @ziyan


kanbouchou:
``_mapExtraGeometries`` is still left not updated. https://github.com/rdiankov/openrave/blob/master/include/openrave/kinbody.h#L358

It is not serialized by ``SerializeJSON`` so I assumed we do not intend to update / save this information.

rdiankov:
thanks~

PR #922, opened 2020-11-25
Fix insufficient mimic velocities

tgn3000:
If the user did not (forgot to) provide enough mimic velocity formulas, then we print out a warning and set the first-order partial derivative as 0, instead of let index go out of bound of `std::vector`.

Related PR is #908.


rdiankov:
thanks~

Issue #893, opened 2020-09-22
Osg viewer truncates update viewer update commands 


braineo:
## How to reproduce

Draw more than 1000 axes in a loop
```python
import numpy
viewer = env.GetViewer()

indices  = range(2000)

Ts = []
offset = 0
for i in indices:
    T = numpy.eye(4)
    T [:3, 3] = [offset, 0 , 0]
    #T[:3, 3] = numpy.random.random(3)
    Ts += [T]
    offset += 0.01

handlers = []
for i in indices:
    handlers.append(misc.DrawAxes(env, Ts[i], dist=0.025, linewidth=4))
```

## Observation

Only 1000 Axes are drawn on the viewer, suspect the following piece truncating the number of adding child to scene graph

```cpp
// plugins/qtosgrave/qtosgviewer.cpp
void QtOSGViewer::_PostToGUIThread(const boost::function<void()>& fn, bool block)
{
    if( _nQuitMainLoop != -1 ) {
        // viewer quit, so anything posted won't get processed
        return;
    }

    boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
    if( _listGUIFunctions.size() > 1000 ) {
        // can happen if system is especially slow
        //RAVELOG_WARN_FORMAT("too many gui post commands, ignoring %d", _listGUIFunctions.size());
        return;
    }
```
@Hsifnus

Hsifnus:
The trivial solution to this would be to expand or remove the limit on `_listGUIFunctions.size()`. That being said, would it be better if certain functions sent to `_PostToGUIThread` were able to be condensed down into a single function pushed onto `_listGUIFunction` that accumulates similar function calls and calls each one in sequence?

PR #925, opened 2020-11-30
Priority-based GUI thread message queueing system (Issue #893 cont'd)

Hsifnus:
This pull request addresses issue #893 by categorizing all GUI thread message functions into three priority categories, with lesser-priority message being queued later in the queue and more likely to be dropped when the queue exceeds a certain size limit. High-priority messages are deemed essential and are able to enter the queue even when the viewer is unable to process messages (other messages are instead dropped on the spot in this case) as well as join the queue beyond the regular size limit, although a hard limit for high-priority messages still exists and can be set in the event that an extremely high amount of messages flood the queue. Such a solution allows important messages like handle closures to be handled to the best of the viewer's ability while devoting less resources to less essential messages if needed.

Feel free to suggest or edit in changes in the priority values given to each of the GUI thread message functions.

### DIFF
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -1187,10 +1187,12 @@ void QtOSGViewer::_ProcessApplicationQuit()
 {
     RAVELOG_VERBOSE("processing viewer application quit\n");
     // remove all messages in order to release the locks
-    boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
-
     FOREACH(itpriority, _mapGUIFunctionLists) {
-        list<GUIThreadFunctionPtr> listGUIFunctions = itpriority->second;
+        list<GUIThreadFunctionPtr> listGUIFunctions;
+        {
+            boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
+            listGUIFunctions.swap(itpriority->second);
+        }
         if( listGUIFunctions.size() > 0 ) {
             bool bnotify = false;
             FOREACH(it,listGUIFunctions) {
@@ -2164,10 +2166,13 @@ void QtOSGViewer::SetEnvironmentSync(bool bUpdate)
 
     if( !bUpdate ) {
         // remove all messages in order to release the locks
-        boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
         // Clear GUI function lists for all priorities
         FOREACH(itpriority, _mapGUIFunctionLists) {
-            list<GUIThreadFunctionPtr> listGUIFunctions = itpriority->second;
+            list<GUIThreadFunctionPtr> listGUIFunctions;
+            {
+                boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
+                listGUIFunctions.swap(itpriority->second);
+            }
             if( listGUIFunctions.size() > 0 ) {
                 bool bnotify = false;
                 FOREACH(it,listGUIFunctions) {
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -1353,10 +1353,10 @@ void QtOSGViewer::quitmainloop()
 void QtOSGViewer::Show(int showtype)
 {
     if( showtype ) {
-        _PostToGUIThread(boost::bind(&QtOSGViewer::show, this), GUIThreadQueuePriority::HIGH);
+        _PostToGUIThread(boost::bind(&QtOSGViewer::show, this), GUIThreadQueuePriority::VERY_HIGH);
     }
     else {
-        _PostToGUIThread(boost::bind(&QtOSGViewer::hide, this), GUIThreadQueuePriority::HIGH);
+        _PostToGUIThread(boost::bind(&QtOSGViewer::hide, this), GUIThreadQueuePriority::VERY_HIGH);
     }
 //    // have to put this in the message queue
 //    if (showtype ) {
@@ -2130,7 +2130,7 @@ void QtOSGViewer::_UpdateEnvironment()
 size_t QtOSGViewer::_GetQueueSizeForPriority(GUIThreadQueuePriority priority) {
     size_t queueSize = 0;
     for (uint8_t i = static_cast<uint8_t>(priority); i < 4; i++) {
-        queueSize += _mapGUIFunctionListLimits[static_cast<GUIThreadQueuePriority>(i)].size();
+        queueSize += _mapGUIFunctionLists[static_cast<GUIThreadQueuePriority>(i)].size();
     }
     return queueSize;
 }
@@ -2154,7 +2154,7 @@ void QtOSGViewer::_PostToGUIThread(const boost::function<void()>& fn, GUIThreadQ
         RAVELOG_WARN("GUI thread function of priority %d dropped due to function queue for given priority exceeding size of %d", priority, _mapGUIFunctionListLimits[priority]);
         return;
     }
-    _mapGUIFunctionLists[priority].push_back(pfn)
+    _mapGUIFunctionLists[priority].push_back(pfn);
     if( block ) {
         while(!pfn->IsFinished()) {
             _notifyGUIFunctionComplete.wait(_mutexGUIFunctions);
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -470,13 +470,16 @@ void QtOSGViewer::_UpdateViewerCallback()
                 }
             }
         }
-        if (!_excessGUIFunctions.empty()) {
+        if (!_listGUIFunctionsBuffer.empty()) {
             boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
-            int excessSize = _excessGUIFunctions.size();
-            for (int idx = 0; idx < excessSize && _listGUIFunctions.size() < 1000; idx++) {
-                _listGUIFunctions.push_back(_excessGUIFunctions.front());
-                _excessGUIFunctions.pop_front();
-            }
+            int excessSize = _listGUIFunctionsBuffer.size();
+            auto insertPos = _listGUIFunctions.begin();
+            auto bufferStartPos = _listGUIFunctionsBuffer.begin();
+            auto bufferEndPos = _listGUIFunctionsBuffer.begin();
+            std::advance(insertPos, _listGUIFunctions.size());
+            std::advance(bufferStartPos, 1);
+            std::advance(bufferEndPos, std::min(1000 - _listGUIFunctions.size(), _listGUIFunctionsBuffer.size())+1);
+            _listGUIFunctions.splice(insertPos, _listGUIFunctionsBuffer, bufferStartPos, bufferEndPos);
         }
     }
     catch(const std::exception& ex) {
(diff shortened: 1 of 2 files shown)


braineo:
Since we are executing all posted message, probably we don't need to keep track of indices of them. In the case of dropping viewer command it will become tedious work to maintain.

Instead we can simply store commands in different list and care about  the total size of all combined

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2112,21 +2127,34 @@ void QtOSGViewer::_UpdateEnvironment()
     }
 }
 
-void QtOSGViewer::_PostToGUIThread(const boost::function<void()>& fn, bool block)
+size_t QtOSGViewer::_GetQueueSizeForPriority(GUIThreadQueuePriority priority) {
+    size_t queueSize = 0;
+    for (uint8_t i = static_cast<uint8_t>(priority); i < 4; i++) {
+        queueSize += _mapGUIFunctionLists[static_cast<GUIThreadQueuePriority>(i)].size();
```
braineo, line 2133: This is doing something different judging from the name. The function name indicates you are getting viewer command number for a specific priority

plugins/qtosgrave/qtosgviewer.h
```diff
@@ -342,11 +342,22 @@ public slots:
     virtual void _PanCameraXDirection(float dx);
     virtual void _PanCameraYDirection(float dy);
 
+    /// \brief priority values used to determine how important it is to process certain GUI thread functions over others
+    enum GUIThreadQueuePriority : uint8_t {
```
braineo, line 346: GUIThreadQueuePriority -> ViewerCommandPriority

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -1180,19 +1188,23 @@ void QtOSGViewer::_ProcessApplicationQuit()
...
     // remove all messages in order to release the locks
     boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
-    if( _listGUIFunctions.size() > 0 ) {
-        bool bnotify = false;
-        FOREACH(it,_listGUIFunctions) {
-            (*it)->SetFinished();
-            if( (*it)->IsBlocking() ) {
-                bnotify = true;
+
+    FOREACH(itindex, _mapGUIFunctionLists) {
```
braineo, line 1192: itindex? itpriority?

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2112,21 +2127,34 @@ void QtOSGViewer::_UpdateEnvironment()
...
+    if (!_bUpdateEnvironment && priority < GUIThreadQueuePriority::HIGH) {
+        RAVELOG_WARN("GUI thread function of priority %d dropped due to _bUpdateEnvironment being set to false", priority);
         return;
     }
-    GUIThreadFunctionPtr pfn(new GUIThreadFunction(fn, block));
-    _listGUIFunctions.push_back(pfn);
+    // GUI thread function queue size limit
+    if (_GetQueueSizeForPriority(priority) >= _mapGUIFunctionListLimits[priority]) {
+        RAVELOG_WARN("GUI thread function of priority %d dropped due to function queue for given priority exceeding size of %d", priority, _mapGUIFunctionListLimits[priority]);
+        return;
```
braineo, line 2155: Maybe dropping old command makes more sense?

For example, if someone is setting camera position, we would want to apply the new value

braineo (changes requested):

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2145,14 +2157,16 @@ void QtOSGViewer::_PostToGUIThread(const boost::function<void()>& fn, GUIThreadQ
...
-        RAVELOG_WARN("GUI thread function of priority %d dropped due to _bUpdateEnvironment being set to false", priority);
+    if (!_bUpdateEnvironment && priority < ViewerCommandPriority::HIGH) {
+        RAVELOG_WARN("New GUI thread function of priority %d dropped due to _bUpdateEnvironment being set to false", priority);
         return;
     }
     // GUI thread function queue size limit
-    if (_GetQueueSizeForPriority(priority) >= _mapGUIFunctionListLimits[priority]) {
-        RAVELOG_WARN("GUI thread function of priority %d dropped due to function queue for given priority exceeding size of %d", priority, _mapGUIFunctionListLimits[priority]);
-        return;
+    if (_GetViewerCommandNumber(priority) >= _mapGUIFunctionListLimits[priority]) {
```
braineo, line 2165: Why we are comparing command number of several priority combined and a specific limit of one priority?

rdiankov:
@braineo let me know when this is ready for my review. thanks~

plugins/qtosgrave/qtosgviewer.cpp
Hsifnus: We get the command number using all priorities at or above the input priority as a way to ensure that lower priority commands will always have equal or greater command numbers than higher priority commands and are therefore will either force an old command to be dropped or be dropped themselves at the same point as or an earlier point than higher priority commands.

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2112,21 +2127,31 @@ void QtOSGViewer::_UpdateEnvironment()
...
+    // Block non-essential fucntions if viewer is not processing any messages
+    if (!_bUpdateEnvironment && priority < ViewerCommandPriority::HIGH) {
+        RAVELOG_WARN("New GUI thread function of priority %d dropped due to _bUpdateEnvironment being set to false", priority);
         return;
     }
-    GUIThreadFunctionPtr pfn(new GUIThreadFunction(fn, block));
-    _listGUIFunctions.push_back(pfn);
+    // GUI thread function queue size limit
+    if (_mapGUIFunctionLists[priority].size() >= _mapGUIFunctionListLimits[priority]) {
+        if(_mapGUIFunctionLists[priority].size() > 0) {
```
braineo, line 2146: Is this check necessary? you are inside `_mapGUIFunctionLists[priority].size() >= _mapGUIFunctionListLimits[priority]`

braineo (changes requested):

braineo (approved):

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -1180,19 +1188,23 @@ void QtOSGViewer::_ProcessApplicationQuit()
...
     boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
-    if( _listGUIFunctions.size() > 0 ) {
-        bool bnotify = false;
-        FOREACH(it,_listGUIFunctions) {
-            (*it)->SetFinished();
-            if( (*it)->IsBlocking() ) {
-                bnotify = true;
+
+    FOREACH(itpriority, _mapGUIFunctionLists) {
+        list<GUIThreadFunctionPtr> listGUIFunctions = itpriority->second;
```
rdiankov, line 1193: @Hsifnus don't you mean to use reference here? or perhaps you want to swap?

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2144,19 +2165,23 @@ void QtOSGViewer::SetEnvironmentSync(bool bUpdate)
...
         boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
-        if( _listGUIFunctions.size() > 0 ) {
-            bool bnotify = false;
-            FOREACH(it,_listGUIFunctions) {
-                (*it)->SetFinished();
-                if( (*it)->IsBlocking() ) {
-                    bnotify = true;
+        // Clear GUI function lists for all priorities
+        FOREACH(itpriority, _mapGUIFunctionLists) {
+            list<GUIThreadFunctionPtr> listGUIFunctions = itpriority->second;
```
rdiankov, line 2170: @Hsifnus don't you mean to use reference here? or perhaps use swap?

rdiankov:
@Hsifnus in general, I think `draw` should be MEDIUM priority and closing the handles (ie cleaning up) should be HIGH

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2089,18 +2103,21 @@ void QtOSGViewer::_UpdateEnvironment()
...
         }
 
-        FOREACH(itmsg, listGUIFunctions) {
-            try {
-                (*itmsg)->Call();
-            }
-            catch(const std::exception& ex) {
-                RAVELOG_WARN_FORMAT("gui thread function failed: %s", ex.what());
+        FOREACH(itpriority, mapGUIFunctionLists) {
+            list<GUIThreadFunctionPtr> listGUIFunctions = itpriority->second;
```
rdiankov, line 2113: @Hsifnus need reference, otherwise copying.

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -1179,20 +1187,26 @@ void QtOSGViewer::_ProcessApplicationQuit()
...
-        FOREACH(it,_listGUIFunctions) {
-            (*it)->SetFinished();
-            if( (*it)->IsBlocking() ) {
-                bnotify = true;
-            }
+    FOREACH(itpriority, _mapGUIFunctionLists) {
+        list<GUIThreadFunctionPtr> listGUIFunctions;
+        {
+            boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
+            listGUIFunctions.swap(itpriority->second);
```
rdiankov, line 1194: @Hsifnus why not swap the whole map before the for loop starts?

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -2143,20 +2166,27 @@ void QtOSGViewer::SetEnvironmentSync(bool bUpdate)
...
-                (*it)->SetFinished();
-                if( (*it)->IsBlocking() ) {
-                    bnotify = true;
-                }
+        // Clear GUI function lists for all priorities
+        FOREACH(itpriority, _mapGUIFunctionLists) {
+            list<GUIThreadFunctionPtr> listGUIFunctions;
+            {
+                boost::mutex::scoped_lock lockmsg(_mutexGUIFunctions);
+                listGUIFunctions.swap(itpriority->second);
```
rdiankov, line 2174: @Hsifnus why not swap the whole map before the for loop?

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -193,18 +199,20 @@ QtOSGViewer::~QtOSGViewer()
...
-            }
-            catch(const boost::bad_weak_ptr& ex) {
-                // most likely viewer
-            }
-            catch(const std::exception& ex2) {
-                RAVELOG_WARN_FORMAT("unexpected exception in gui function: %s", ex2.what());
+        FOREACH(itpriority, _mapGUIFunctionLists) {
+            FORIT(itmsg, itpriority->second) {
+                try {
+                    (*itmsg)->Call();
```
rdiankov, line 205: @Hsifnus I'm actually wondering why we need to keep `_mutexGUIFunctions` locked during the Call... perhaps we can get away with not..?

Hsifnus: Given the nature of where and when the destructor is typically called, perhaps going without the lock does makes sense.

rdiankov:
woot! finally. thanks~

PR #926, opened 2020-12-03
[quick] Bugfix deltatime

kanbouchou:
``deltatime`` was computed but not written to output


kanbouchou:
for ``Sample`` without ``spec``, it is correctly set as in https://github.com/rdiankov/openrave/blob/master/src/libopenrave-core/generictrajectory.cpp#L326

rdiankov:
nice catch~

PR #928, opened 2020-12-06
Add grabbed info hash

Puttichai:
- Added a function to compute a hash for `KinBody::GrabbedInfo`. This hash can be used to identify if the robot is grasping the same body the same way.
- Added `Save_JointResolutions` option to `KinBody::SaveParameters`.
- Fixes for python linkstatisticsmodel
  - Make sure that robot dof values are set to zeros and transform to identity before computation.
  - When setting robot dof values to zeros, need to specify `CLA_Nothing` to ignore joint limits.


rdiankov:
thanks~

PR #929, opened 2020-12-11
20x speed up trajectory sampling 

kanbouchou:
optiimizations in order of descending impact
--------------------------------------------------------------
- call ``ConfigurationSpecification::ConvertData`` just once
- ``lower_bound`` narrows down search space given time is monotonically increasing
- pass one double ``deltatime`` instead of vector of double. converting between numpy array and std::vector takes time.

in ipython, given ``traj``, 
```
from numpy import *
from openravepy import *
dt = 0.004
times = arange(0, traj.GetDuration() + dt, dt).tolist()
config = traj.GetConfigurationSpecification()
config.RemoveGroups(config.GetGroups()[-1].name) # modify from original config to avoid cheating.
%timeit traj.SamplePointsSameDeltaTime2D(dt, True, config)
%timeit traj.SamplePoints2D(times, config)
```

sample output for config with dof=29 and 353 elements in times is
```
In [10]: %timeit traj.SamplePointsSameDeltaTime2D(dt, True, config)
10000 loops, best of 3: 130 µs per loop

In [11]: %timeit traj.SamplePoints2D(times, config)
100 loops, best of 3: 3.97 ms per loop
```


rdiankov:
@kanbouchou thanks! this is a great idea. However, can we change the name to be a little more clear? How about `SamplePointsSameDeltaTime`?

kanbouchou:
@rdiankov I've changed the new api name to ``SamplePointsSameDeltaTime`` (and ``SamplePointsSameDeltaTime2D``).

rdiankov:
thanks~

PR #931, opened 2020-12-17
Add optional `IkChainEndLink` to manipulator and ikfast.

AntoineRyu:
Add optional `IkChainEndLink` to manipulator. Used to select a subset of the kinematics chain when using `iksolver`. Useful for actuated `ikChainEndLink` to `endEffector` chain that can change over time, but not by the `iksolver` itself when looking for a solution.


rdiankov:
thanks~ in the future, you can make PR to `production` branch~

PR #932, opened 2020-12-21
Add APIs for user to register forward kinematics functions

tgn3000:
Add functions
- `KinBody::SetTransformWithoutUpdateStampId`
- `KinBody::GetDependencyOrderedJointsAll`
- `KinBody::ProcessAfterSetDOFValues`
- `KinBody::IncrementUpdateStamp`

Add class
- `KinBody::ForwardKinematicsStruct`

A `SetDOFValues` function can compute transforms for all links, and then call `KinBody::IncrementUpdateStamp` with the number of links (`nlinks`)---this reduces `nlinks` calls of `GetParent()` that copies shared pointers.

When a `KinBody` is switching between different connected bodies, it is only associated with one kinbody calculator module `pCalculatorModule`, and we map each different kinematics geometry hash to a forward kinematics structure.


python/bindings/openravepy_global.cpp
```diff
@@ -1419,6 +1419,16 @@ void init_openravepy_global()
 // #endif
     ;
 
+#ifdef USE_PYBIND11_PYTHON_BINDINGS
+    m.def("GetMilliTime", utils::GetMilliTime, "get millisecond time");
```
rdiankov, line 1423: @tgn3000 you should really expose `GetMilliTime64` instead

include/openrave/kinbody.h
```diff
@@ -841,9 +841,14 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
         /// \brief Sets the transform of the link regardless of kinematics
         ///
-        /// \param[in] t the new transformation
+        /// \param[in] transform the new transform of link
         virtual void SetTransform(const Transform& transform);
 
+        /// \brief Sets the transform of the link regardless of kinematics without updating parent kinbody's stamp id
+        ///
+        /// \param[in] transform the new transformation
+        virtual void SetTransformWithoutUpdateStampId(const Transform& transform);
```
rdiankov, line 850: @tgn3000 this encourages bad programming if we expose it like this. there are better ways to do this....

include/openrave/kinbody.h
```diff
@@ -3066,6 +3083,25 @@ class OPENRAVE_API KinBody : public InterfaceBase
     ConfigurationSpecification _spec;
     CollisionCheckerBasePtr _selfcollisionchecker; ///< optional checker to use for self-collisions
 
+public:
+    // Gary registers forward kinematics functions
```
rdiankov, line 3087: @tgn3000 no need for names here ;0)

include/openrave/kinbody.h
```diff
@@ -3066,6 +3083,25 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+    struct ForwardKinematicsStruct {
+        ForwardKinematicsStruct();
+        ModuleBaseConstPtr pCalculatorModule = nullptr; ///< kinbody basic calculators module
```
rdiankov, line 3090: @tgn3000 `= nullptr` is redundant

include/openrave/kinbody.h
```diff
@@ -3066,6 +3083,25 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+        boost::function<bool(const std::vector<double>&)> pSetLinkTransformsFn; ///< function that sets links' transforms
```
rdiankov, line 3091: @tgn3000 use `dReal` instead of `double`

include/openrave/kinbody.h
```diff
@@ -3066,6 +3083,25 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+        boost::function<void(std::vector<double>&)> pGetDOFLastSetValuesFn; ///< function that updates kinbody's dof values
+        bool bInitialized = false; ///< indicator of successful initialization
```
rdiankov, line 3093: @tgn3000 what is the use of this? why not just check if `pSetLinkTransformsFn` is valid?

include/openrave/kinbody.h
```diff
@@ -3066,6 +3083,25 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
+
+    /// \brief Associate the kinbody's current kinematics geometry hash with a forward kinematics structure
+    /// \param[in] fkstruct a forward kinematics structure
+    /// \param[in] bOverWrite whether to replace a registered forward kinematics structure if it exists
+    ///
+    /// \return true if the forward kinematics structure to be registered has all the necessary facilitities we can use. 
+    bool RegisterForwardKinematicsStruct(const ForwardKinematicsStruct& fkstruct, const bool bOverWrite=false);
+
+protected:
+    std::map<std::string, ForwardKinematicsStruct> _mHash2ForwardKinematicsStruct; ///< maps a kinematics geometry hash to a forward kinematics structure
```
rdiankov, line 3104: @tgn3000 checking for a hash in the map every call is not very efficient. i think there should be a way for you to get a `ForwardKinematicsStruct _currentFK` variable that gets initialized from your DB of `ForwardKinematicsStruct` during the compute internal initialization. That way, you never have to check anything before using it.

python/bindings/openravepy_global.cpp
tgn3000:
```
    m.def("GetMilliTime", utils::GetMilliTime64, "get millisecond time");
```

include/openrave/kinbody.h
tgn3000: removed

include/openrave/kinbody.h
tgn3000: removed

include/openrave/kinbody.h
tgn3000: replaced

include/openrave/kinbody.h
tgn3000: `bInitialized==true` means all FK functions are properly initialized

include/openrave/kinbody.h
tgn3000: Admittedly if I call `SetTransform` then things still work.

I brought in `SetTransformWithoutUpdateStampId` because `Link::GetParent` copies the shared pointer and takes up noticeable amount of time in profiling. Since my FK module keeps these Links' parent body, I set their transforms and increase the body's `_nUpdateStampId` by the number of links at one time.
```
        KinBodyPtr KinBody::Link::GetParent(bool trylock=false) const {
            if( trylock ) {
                return _parent.lock();
            }
            else {
                return KinBodyPtr(_parent);
            }
        }

void KinBody::Link::SetTransform(const Transform& t)
{
    _info._t = t;
    GetParent()->_nUpdateStampId++;
}
```

rdiankov: @tgn3000 you can make the  SetTransform without update time internal and pass it in to the FK solver privately.

include/openrave/kinbody.h
rdiankov: @tgn3000 isn't that the same as checking `!!pSetLinkTransformsFn && !!pGetDOFLastSetValuesFn`?

tgn3000: ok

include/openrave/kinbody.h
tgn3000: Now I use `_pCurrentForwardKinematicsStruct` to cache the currently used  `ForwardKinematicsStruct` with the correct kinematics geometry hash.

include/openrave/kinbody.h
tgn3000: Now I hide it inside `ForwardKinematicsStruct::pSetSingleLinkTransformFn`.

tgn3000:
@rdiankov ping. Can you see whether the latest changes are appropriate? Thanks!

PR #933, opened 2020-12-24
Fix parabolicsmoother

Puttichai:
### Summary

In the initialization of the resulting trajectory in `parabolicsmoother.cpp`, there are cases when a `ParabolicRampND` that goes into the resulting trajectory is different from the one that has been validated with constraints.

### Changes

- Modified the part where a newly computed, unchecked ramp is put into the final trajectory. (`parabolicsmoother`)
- Added checking for acceleration violation in `SegmentFeasible2`. (`parabolicsmoother`)
- For safety, make sure that the ramps to be put into the final trajectory have valid check results with `bDifferentVelocity` being `false` (previously not considered). (`parabolicsmoother` and `parabolicsmoother2`)
- Minor fix on the condition to do trajectory sampling for verification at the end.


rdiankov:
thanks~

PR #938, opened 2021-01-20
Set camera focal point when setting camera transform with focus distance

braineo:


rdiankov:
@braineo  not sure if this fixes anything...?

braineo:
@rdiankov OSG does not set focal point when setting camera transform. `_SetCameraDistanceToFocus` will move camera to a random place without this fix

rdiankov:
sure, let's see what happens

PR #943, opened 2021-01-27
recreate controller on number of DOFs change

yoshikikanemoto:
This is to support number of DOFs change by connected body change.


yoshikikanemoto:
FYI @strmio

rdiankov:
thanks~

PR #946, opened 2021-02-05
Add Python binding for KinBody::Joint::IsActive()

strmio:
This patch adds a missing Python binding for `KinBody::Joint::IsActive()`.


rdiankov:
thanks~

PR #948, opened 2021-02-10
fix assert to non-existing

yoshikikanemoto:
it seems this assert checks pointer of `time()` function in `time.h`


tgn3000:
I saw this before. Another example of side-effect when people copy and paste code.
That line was from
```
void Sample(std::vector<dReal>& data, dReal time) const override
```

PR #949, opened 2021-02-11
serialize ikp none

kanbouchou:
otherwise, serializing none type ikparameter fails at https://github.com/rdiankov/openrave/blob/master/include/openrave/openrave.h#L2836


rdiankov:
thanks~

PR #952, opened 2021-02-18
Fix scalegeometry option for ReadKinBodyURI

justism:
This MR fixes scalegeometry in colladareader and adds support for this option in jsonreader.
Issue: https://git.mujin.co.jp/backend/backend/-/issues/322


rdiankov:
thanks~

PR #953, opened 2021-02-18
Migrate `sensor_reference` to `referenceAttachedSensorName`

ziyan:


rdiankov:
@ziyan in `robotconnectedbody.cpp`, you'll need to prefix `referenceAttachedSensorName`

rdiankov:
thanks~

PR #954, opened 2021-02-18
Optimize CheckCollision and generic physics engine access,  Recycle env id

kanbouchou:
background
=============

- on scenes with hundreds of kin bodies, large portion of collision check was spent in ``FCLSpace::GetInfo``.
- in applications where many objects are grabbed by robot, lot of computation time was spent in ``_UpdateGrabbedBodies`` mostly called from ``SetDOFValues``

Optimizations in this MR
======================

- For frequently accessed data (namely ``_currentpinfo`` in ``fclspace.h`` and ``_pysicsDataCache`` in ``genericphysicsengine.cpp``), use ``vector`` instead of ``map`` for constant search time.
- Cache generic physics engine user data
- ``_tmpSortedBuffer`` (previously ``_tmpbuffer``) is sorted to use ``lower_bound`` search which is more efficient than ``find``
- use reference instead of shared pointer for passing and returning when possible
- inline ``FCLSpace::GetInfo`` to avoid function overhead
- use range-based for loop instead of ``FOREACH``

Note
================

 - to use vector as data structure for holding kin body related data, we need to prevent from environment id from growing over time, so id of removed kin bodies are recycled. It  cannot be assumed anymore that an environment id maps to same kin body over time.

Performance
============

- with scene with 600 objects, 50% speed up of ``env.CheckCollision(robot)``
- with scene with small number of objects, about 10% speed up.

checkcollision

- SetDOFValues time increases 1.5us per grabbed item in production
- SetDOFValues time increases 0.7us per grabbed item in this MR

SetDOFValues


rdiankov:
woot! congratulations!

kanbouchou:
For developers and users, here is a short summary of behavior change in this PR.

1. environment body index (formerly environment id) can be reused over time.
2. now ``GetAttached`` correctly implements the spec ``If any bodies are already in setAttached, then ignores recursing on their attached bodies.``.  Previously it was not.

For 1.  environment body index of a kinbody (``body1``), say 9, in an env doesn't change as long as it is not removed from the env. Once ``body1`` is removed, and another kinbody (``body2``) is added later, ``body2`` may be assigned 9.

PR #955, opened 2021-02-21
Fix Updating Link Transform

justism:
This MR fixes updating a kin body link transform when it is updated from webstack by requiring reinitialization.  
Currently, when the link transform is updated from webstack, the changes are reset when `SetDOFValues` is called.   
Related issues:  
https://git.mujin.co.jp/dev/webstack/-/issues/33  
FR Shanghai: https://mattermost.mujin.co.jp/china-support/pl/ojahekhhn78mpjjwsa3j67zbfo


rdiankov:
@justism @ziyan i just made some changes that set the modified flags to 0xffffffff by default, and they can only be reset to 0 in the Extract call. please review.

ziyan:
@rdiankov if you remove setting `_modifiedFields = (uint32_t)-1` in python binding, then someone who did `ExtractInfo` in python and then updating the python object, and then `UpdateFromInfo` in python, will be confused since their change will not take effect

rdiankov:
@ziyan python binding always creates new Info, so it will always be modified now, so it will work fine.

rdiankov:
thanks~

include/openrave/geometry.h
```diff
@@ -491,11 +491,10 @@ class RaveTransform
         return !operator==(right);
     }
 
-    inline bool Compare(const RaveTransform<T>& rhs, T epsilon=1e-7) const {
-        T e1 = (rot-rhs.rot).lengthsqr4();
-        T e2 = (rot+rhs.rot).lengthsqr4();
-        T e = e1 < e2 ? e1 : e2;
-        return RaveSqrt((trans-rhs.trans).lengthsqr3() + e) <= epsilon;
+    /// \brief return true if any element of transform is different by mor ethan epsilon, otherwise false.
+    inline bool CompareTransform(const RaveTransform<T>& rhs, T epsilon) const {
```
Puttichai, line 495: @ziyan @rdiankov Is it intentional that we don't account for the fact that `quat` and `-quat` represent the exact same orientation here?

PR #958, opened 2021-02-26
Fix sphere always required reinitialize

kanbouchou:
also added some logs so that every location ``UFIR_RequireReinitialize`` is returned has verbose message.

cc @ziyan


rdiankov:
thanks~

PR #960, opened 2021-03-03
[quick] Fix typo

kanbouchou:
I guess following code is legal
```
bool func() { return true; }

int main()
{
    -func();
    return 0;
}
```


rdiankov:
thanks~

PR #962, opened 2021-03-04
Add functions for computing AABBs of geometries from specific group

Puttichai:
### Summary

Currently if we need to compute the AABB of a specific geometry group in a kinbody `body`, we need to first change the current geometry group of `body` via `body.SetLinkGeometriesFromGroup(desiredGroupName)` and then finally call `body.ComputeAABB()` or `body.ComputeLocalAABB()`. This involves many unnecessary steps (provided that the end goal is simply to compute the AABB) such as updating link collision meshes. 

This pull request adds functions that allow users to directly query for AABBs of specific geometry groups without having to change the current geometry group of the kinbody.


rdiankov:
thanks~

PR #964, opened 2021-03-16
Fix ignore_link_pair loaded from collada was not serialized into json

yoshikikanemoto:


src/libopenrave-core/colladaparser/colladareader.cpp
```diff
@@ -4968,6 +4968,7 @@ class ColladaReader : public daeErrorHandler
                                 continue;
                             }
                             pbody->_vForcedAdjacentLinks.emplace_back(plink0->GetName(), plink1->GetName());
+                            plink0->_info._vForcedAdjacentLinks.push_back(plink1->GetName()); // needs to set this as well because json serialization uses this.
```
ziyan, line 4971: This is going to fix one path. Alternatively, you can do this in either KinBody::ExtractInfo or KinBody::Link::ExtractInfo

rdiankov:
thanks~

PR #965, opened 2021-03-17
Prefix tool names in gripper info with connected body name

strmio:
As well as link names, tool names in gripper info need to be prefixed with the parent connected body name.


rdiankov:
@strmio i'm not sure i understand this MR... let's talk in person.

strmio:
@rdiankov I just removed the unnecessary line for `tools` as we discussed.

rdiankov:
thanks~

PR #967, opened 2021-03-18
Fix removing readable interfaces on update with KinBodyInfo

yoshikikanemoto:
It was using iterator invalidated by  `ClearReadableInterface`.


rdiankov:
@yoshikikanemoto nice catch! But i really don't like allocation in vectors~ ;0/

how about:
```
    FOREACH(itExisting, GetReadableInterfaces()) {
        bool bFound = false;
        FOREACHC(it, info._mReadableInterfaces) {
            if (itExisting->first == it->first) {
                bFound = true;
                break;
            }
        }
        if (!bFound) {
            ClearReadableInterface(itExisting->first);
            updateFromInfoResult = UFIR_Success;
            RAVELOG_VERBOSE_FORMAT("body %s updated due to readable interface %s removed", _id%itExisting->first);
        }
    }
```



By the way, your method can be improved by having `readableInterfaceKeys` only hold the interfaces that will be delcated... but i still think it's slower than the above iteration due to the allocation ;0/

yoshikikanemoto:
to be allocation-free, i don't have better idea for now. i changed to your version

yoshikikanemoto:
@rdiankov autotest all passed. please review this again

rdiankov:
thanks~

PR #969, opened 2021-03-24
Fix: Forced adjacent link names in connected body need to be prefixed with connected body name

strmio:
Otherwise link pairs marked as adjacent in a connected body are not handled as adjacent.


rdiankov:
thanks~

PR #970, opened 2021-03-26
Further clean up and fixes for NotifyEnvironmentChanged

ziyan:
This change adds handling of `__created__` flag in diff JSON. When applying diff JSON to env, having just `__deleted__` is not enough, we also need `__created__` to distinguish if a JSON object is new and complete, or if it is a partial diff. In the case of partial diff, we have to avoid creating new object with just partial info.

**NOTE: Before merging, look at the webstack changes first**


rdiankov:
@ziyan we'll need to talk about this in person...

rdiankov:
ziyan, just merged the isPartial branch. Is this branch still necessary?

rdiankov:
@ziyan ping, close this if not necessary.

PR #971, opened 2021-03-26
Fixes for fclmanager

Puttichai:
### Changes
- Commented out flooding (verbose) log message.
- Added checking `!!pbody` before using calling `pbody->GetEnv()->GetId()`.

cc @kanbouchou


rdiankov:
thanks!

PR #972, opened 2021-03-26
fix not checking difference of readable interface

yoshikikanemoto:
sorry for easy mistake. reported by @ziyan


rdiankov:
thanks!

PR #976, opened 2021-03-30
print env name

kanbouchou:
Background
==============
This MR aims to improve productivity by logging environment name in addition to env id (integer)

Changes
============
- print env name in addition to env id in many locations
- make env name settable only in initialization phase (constructor and cloning)
  -  ``SetName`` api is removed
  -  ``GetName`` does not lock any mutex, assuming name does not change after initialization.
  -  ``EnvironmentBaseInfo`` doesn't have ``_name`` anymore.


rdiankov:
thanks~

PR #977, opened 2021-03-31
wip: optimize _vForcedAdjacentLinks data structure

kanbouchou:
backgournd
============

- when there were many links, ``_vForcedAdjacentLinks`` access was slow

major change
===============

- previously a vector containing pair of strings were used
  -  unsorted, so searching took O(N) time
  - string was used causing lot of memory allocation and deallocation
- now vector containing vector of ints are used
  -  search and insertion takes constant time


rdiankov:
thanks~

PR #978, opened 2021-04-02
Add function AddGeometryToGroup

Puttichai:
### Summary

This pull request adds a new function `AddGeometryToGroup`, which adds the given geometry info to the designated geometry group.

Originally there was only one function `AddGeometry` that either adds the given geometry info to the current geometry group or all existing geometry groups.


rdiankov:
thanks~

PR #979, opened 2021-04-05
Basic Python3 support (if pybind11 is used)

cielavenir:
(I checked Python2 build with testopenrave)

by the way I understand iteritems change is arguable. `six.iteritems` is an idea as we already need to use six in metaclass implementation.

### DIFF
--- a/plugins/configurationcache/openravepy_configurationcache.cpp
+++ b/plugins/configurationcache/openravepy_configurationcache.cpp
@@ -193,7 +193,15 @@ typedef OPENRAVE_SHARED_PTR<PyConfigurationCache> PyConfigurationCachePtr;
 OPENRAVE_PYTHON_MODULE(openravepy_configurationcache)
 {
     using namespace configurationcachepy;
-    import_array();
+
+    // expansion of the macro `import_array()` in
+    // numpy/core/include/numpy/__multiarray_api.h
+    if (_import_array() < 0) {
+        PyErr_Print();
+        PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
+        return;
+    }
+
 #ifdef USE_PYBIND11_PYTHON_BINDINGS
     m.attr("__doc__") = "The module contains configuration cache bindings for openravepy\n";
 #else
--- a/python/bindings/convexdecompositionpy.cpp
+++ b/python/bindings/convexdecompositionpy.cpp
@@ -178,7 +178,14 @@ BOOST_PYTHON_FUNCTION_OVERLOADS(computeConvexDecomposition_overloads, computeCon
 
 OPENRAVE_PYTHON_MODULE(convexdecompositionpy)
 {
-    import_array();
+    // expansion of the macro `import_array()` in
+    // numpy/core/include/numpy/__multiarray_api.h
+    if (_import_array() < 0) {
+        PyErr_Print();
+        PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import");
+        return;
+    }
+
 #ifndef USE_PYBIND11_PYTHON_BINDINGS
     numeric::array::set_module_and_type("numpy", "ndarray");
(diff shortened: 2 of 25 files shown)


cielavenir:
|Python3 PR wave||
|:--|:--|
|Lint print/except/sort/next/dict.has_key|#973|
|Lint misc|#974|
|Lint izip/StringIO/queue|#975|
|Lint dict.iteritems/metaclass, Update bindings|#978|
|Attempt to fix sympy 0.7.3 comaptibility|---|

plugins/configurationcache/openravepy_configurationcache.cpp
```diff
@@ -193,6 +193,8 @@ typedef OPENRAVE_SHARED_PTR<PyConfigurationCache> PyConfigurationCachePtr;
 OPENRAVE_PYTHON_MODULE(openravepy_configurationcache)
 {
     using namespace configurationcachepy;
+#undef NUMPY_IMPORT_ARRAY_RETVAL
+#define NUMPY_IMPORT_ARRAY_RETVAL
```
rdiankov, line 197: @cielavenir i don't understand this, can you comment on it?

python/bindings/convexdecompositionpy.cpp
```diff
@@ -178,6 +178,8 @@ BOOST_PYTHON_FUNCTION_OVERLOADS(computeConvexDecomposition_overloads, computeCon
 
 OPENRAVE_PYTHON_MODULE(convexdecompositionpy)
 {
+#define NUMPY_IMPORT_ARRAY_RETVAL
+#define NUMPY_IMPORT_ARRAY_RETVAL
```
rdiankov, line 182: @cielavenir two `#define`??

plugins/configurationcache/openravepy_configurationcache.cpp
cielavenir: That NUMPY_IMPORT_ARRAY_RETVAL comes from https://github.com/numpy/numpy/commit/2fe642a524a4bd2f870e7def51140f94d136af73 , which exists until 1.18.x.

This import_array is meant to be called in `init##modname` (Python2) or `PyInit_##modname` (Python3). As these functions return types are different, import_array macro implementation is different in Python2 version and Python3 versions.

This concept is incompatible with [PYBIND11_MODULE](https://github.com/pybind/pybind11/blob/master/include/pybind11/detail/common.h#L332) (and [BOOST_PYTHON_MODULE](https://github.com/boostorg/python/blob/master/include/boost/python/module_init.hpp)).
There are several works to get rid of this matter such as ethz-asl/numpy_eigen#6 and https://github.com/EttusResearch/uhd/blob/master/host/python/pyuhd.cpp#L53 .

Another idea is to import oldnumeric.h https://github.com/numpy/numpy/blob/main/numpy/core/include/numpy/oldnumeric.h

Note that I locally tested numpy 1.18.5. I'm not sure if numpy 1.19+ are supported or not.

cielavenir: It seems that Gary manually expanded the macro.
https://github.com/tgn3000/openrave3/blob/python3/src/bindings/openravepy_int.cpp#L2254

cielavenir: Well it would be possible to cherry-pick https://github.com/tgn3000/openrave3/commit/2fcaccab669a6da9481913bfce75b8b06afb7309

python/bindings/convexdecompositionpy.cpp
rdiankov: @cielavenir please comment

cielavenir: no need anymore

cielavenir:
I changed mind and took Gary's way. We should be able to go numpy 1.19+.

rdiankov:
thanks~

PR #981, opened 2021-04-09
half collision checking time

kanbouchou:
Details
------------------------

- ``GetAttachedEnvironmentBodyIndices`` to just return indices instead of kin bodies
- use custom iterator to optimize copying struct member to vector (``GetLinkTransformations`` and ``GetLinkEnableStates``)
- change in https://github.com/rdiankov/fcl/ which I haven't been able to push due to permission issue? ([fcl_patch.zip](https://github.com/rdiankov/openrave/files/6285366/fcl_patch.zip))

- use ``vector`` instead of ``set`` when reasonable.
- minimize shared pointer construction, destruction, referencing (``->``)

Performance
--------------------

- tested on scenes used for internal testing. For scenes with more than 10 bodies, computation time is reduced to 50-70% of production.
performance


rdiankov:
@kanbouchou `vector<int8_t>`  is only useful when looking up an index and immediately finding if it is present or not. In your case, the point is iteration, and actually number of attached bodies are always significantly less than the number of objects in the scene.  So i think having a `vector<int>` of indices is much better and faster...

kanbouchou:
> @kanbouchou `vector<int8_t>` is only useful when looking up an index and immediately finding if it is present or not. In your case, the point is iteration, and actually number of attached bodies are always significantly less than the number of objects in the scene. So i think having a `vector<int>` of indices is much better and faster...


Thank you for your comment. I think this is referring to ``GetAttachedEnvironmentBodyIndices``.

It is true that number of attached bodies is usually much less than number of bodies in the environment. The reason for using  ``vector<int8_t>``  is that it was the internally used data structure in ``_GetNumAttached`` so we don't need to copy to intermediate temporary data structure.

Let me test if ``vector<int>`` which includes indices of only attached bodies is faster / easier to use.

kanbouchou:
I've made this change to return ``vector<int>`` containing environment body index of attached bodies. I think it's easier to use, and speed is not different by measure.

Also I've made other improvements as well.
- ``GetLinkTransformationPointers`` to return pointers to transformations to avoid copying ``Transform`` of links repeatedly
- ``GetLinkEnableStatesMasks``  which caches ``LinkInfo::_bIsEnabled`` as vector of bitmask for fast enable / disable check.
  - I've also tried a function to return ``vector<bool*>`` for pointer to ``LinkInfo::_bIsEnabled`` and it was faster than ``GetLinkEnableStates`` by 10%. However  dereferencing of ``bool*`` now became the bottleneck for ``vector<bool*>``.  ``GetLinkEnableStatesMasks`` is faster by 10x than ``GetLinkEnableStates``
- ``GetBodiesFromEnvironmentBodyIndices`` which takes multiple indices and return multiple bodies for them. This is to reduce construction / deconstruction of scoped lock by calling ``GetBodyFromEnvironmentBodyIndex`` in loop which turned out to take non-negligible computation time.

Please continue reviewing.

kanbouchou:
discussed some changes off line, wip until changes are implemented and tested.

kanbouchou:
implemented and tested following changes that we discussed

- use bit wise operations for link enable state masks
- remove GetLinkTransformationPointers
- GetAttachedEnvironmentBodyIndices takes sorted vector instead of set

Please continue to review.

kanbouchou:
I implemented and tested following offline comments
- ``GetAttached*`` functions clears input vector / set, 
- cache some vector in fcl to minimize memory allocation.

Please continue to review.

rdiankov:
thanks~

PR #984, opened 2021-04-20
Remove clone ignore attached bodies

kanbouchou:
background
=============

- When cloning kinbody who is grabbing other bodies without ``Clone_IgnoreAttachedBodies`` option,  ``_vGrabbedBodies`` and ``_listAttachedBodies`` of the cloned body can become inconsistent, leading to lots of warnings.
- I cannot think of use case that ``Clone_IgnoreAttachedBodies``  should not be specified, so make it a default behavior and remove that option.


rdiankov:
thanks~

PR #988, opened 2021-05-07
[quick] warning on empty geometry name

kanbouchou:
to make debugging easy.

### DIFF
--- a/src/libopenrave/kinbody.cpp
+++ b/src/libopenrave/kinbody.cpp
@@ -654,8 +654,13 @@ bool KinBody::InitFromGeometries(const std::vector<KinBody::GeometryInfoConstPtr
     plink->_index = 0;
     plink->_info._name = "base";
     plink->_info._bStatic = true;
-    FOREACHC(itinfo,geometries) {
-        Link::GeometryPtr geom(new Link::Geometry(plink,**itinfo));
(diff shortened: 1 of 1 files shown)


Puttichai:
I think this change will produce too many warning messages. Empty geometry name is not entirely unexpected.

rdiankov:
@kanbouchou perhaps do this in `KinBody::_ComputeInternalInformation` instead..

kanbouchou:
> @kanbouchou perhaps do this in `KinBody::_ComputeInternalInformation` instead..

@rdiankov I have second thought. I want to assign a random (based on monotonically increasing counter or timestamp) name if name is empty, not just a warning. otherwise, it is hard to refer to geometries once added.

For example, currently we cannot remove geometry with empty name.
```
void KinBody::Link::RemoveGeometryByName(const std::string& geometryname, bool removeFromAllGroups)
{
    OPENRAVE_ASSERT_OP(geometryname.size(),>,0);
```

What do you think about assigning unique name automatically if geometry name is empty?

rdiankov:
@kanbouchou @Puttichai  the new collision report structure sort of needs unique geometry names, otherwise it won't be able to display the right information. perhaps we should think about auto-assigning unique geometry names

PR #989, opened 2021-05-14
Add GeometryConstPtr to CollisionReport

MiraiHattori:
In order to get which geometry is colliding from collisionreport, I added pyGeometry in CollisionReport.
Also changed FCLSpace to store GeomtryPtr in pinfo->linkinfo->vgeominfos[]->_pgeom.
I also modified function ```std::string CollisionReport::__str__() const``` for debugging purpose to know which geometry is colliding.


MiraiHattori:
I'll close this one

PR #990, opened 2021-05-17
Add geom to collisionreport

MiraiHattori:
In order to get which geometry is colliding from collisionreport, I added pyGeometry in CollisionReport.
Also changed FCLSpace to store GeomtryPtr in pinfo->linkinfo->vgeominfos[]->_pgeom.
I also modified function ```std::string CollisionReport::__str__() const``` for debugging purpose to know which geometry is colliding.


plugins/fclrave/fclspace.h
```diff
@@ -265,12 +296,17 @@ class FCLSpace : public boost::enable_shared_from_this<FCLSpace>
             else {
                 const std::vector<KinBody::Link::GeometryPtr> & vgeometries = plink->GetGeometries();
                 FOREACH(itgeom, vgeometries) {
+                    const KinBody::GeometryPtr& pgeom = *itgeom;
```
kanbouchou, line 299: ``itgeom`` is dereferenced twice unneessarily, let's avoid it by using range based for loop (``for (const KinBody::GeometryPtr& pgeom : vgeometries)``), it's faster than iterator based.

plugins/fclrave/fclspace.h
```diff
@@ -111,13 +137,17 @@ class FCLSpace : public boost::enable_shared_from_this<FCLSpace>
...
+                // make sure to clear vgeominfos after vgeoms because the CollisionObject inside each vgeom element has a corresponding vgeominfo as a void pointer.
+                vgeominfos.resize(0);
             }
 
             KinBody::LinkPtr GetLink() {
                 return _plink.lock();
             }
 
             KinBody::LinkWeakPtr _plink;
+            vector< boost::shared_ptr<GeometryInfo> > vgeominfos; ///< info for every geometry of the kinbody
```
kanbouchou, line 150: ``info for every geometry of the kinbody`` may be a bit misleading, maybe ``info for every geometry of the link`` is more accurate?

plugins/fclrave/fclcollision.h
```diff
@@ -1225,7 +1237,7 @@ class FCLCollisionChecker : public OpenRAVE::CollisionCheckerBase
             LinkConstPtr plink = link_raw->GetLink();
             if( !plink ) {
                 if( link_raw->bFromKinBodyLink ) {
-                    RAVELOG_WARN_FORMAT("The link %s was lost from fclspace (env %d) (userdatakey %s)", GetEnv()->GetId()%link_raw->bodylinkname%_userdatakey);
+                    RAVELOG_WARN_FORMAT("The link %s was lost from fclspace (env %d) (userdatakey %s)", link_raw->bodylinkname%GetEnv()->GetId()%_userdatakey);
```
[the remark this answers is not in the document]
MiraiHattori: I am fixing the former suggestion using `GetEnv()->GetNameId()`.
For the latter part, I don't think we can print the parent body name since `plink` is `nullptr` and `link_raw` does not have parent link information so far.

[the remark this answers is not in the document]
MiraiHattori: We can get `link_raw->bodylinkname` because `bodylinkname` in `FCLSpace::KinBodyInfo::LinkInfo` stores it for debugging purpose. Maybe we can store `parentbodyname` when constructed?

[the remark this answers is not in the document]
MiraiHattori: Parent body was already included.

plugins/fclrave/fclspace.h
MiraiHattori: In order to set `enclosingBV` for the first time in L318 (originally L282), itgeom is used.
Should we compare pgeom and vgeommetries[0], or should we rewrite L300 (originally L268) as
```
const KinBody::GeometryInfo& geominfo = pgeom->GetInfo();
```
?

MiraiHattori:
@kanbouchou I've fixed codes according to your review. Please review the code I've just commited.
Removed `Reset()` function.
Printed envnameid instead of envid
Remove warning coming from
Used dereferenced `pgeom` instead of `(*itgeom)` since `itgeom` is used after to detect its the first time of the for loop.
Fixed comment.

plugins/fclrave/fclspace.h
kanbouchou: then let's keep ``itgeominfo``, and make a new variable to store const reference to ``*pgeominfo``. Currently, ``pgeominfo`` is deferenced three times by ``*`` and ``->`` operators which is time consuming.

MiraiHattori:
@kanbouchou avoided dereferencing pgeominfo. Please review.

MiraiHattori:
Fixed segmentation fault caused by `pfclgeom->setUserData();` before `pfclgeom` nullptr check

plugins/fclrave/fclspace.h
```diff
@@ -82,6 +84,27 @@ class FCLSpace : public boost::enable_shared_from_this<FCLSpace>
...
+            GeometryInfo() : bFromKinBodyGeometry(false) {
+            }
+
+            GeometryInfo(KinBody::GeometryPtr pgeom) : _pgeom(pgeom), bFromKinBodyGeometry(true) {
+            }
+
+            virtual ~GeometryInfo() {
+            }
+
+            KinBody::GeometryPtr GetGeometry() {
```
kanbouchou, line 99: let's inline this, and also ``LinkInfo::GetLink``

MiraiHattori: I slipped my mind. I made them inline now.

kanbouchou:
Looks good to me, passing to Rosen.

rdiankov:
thanks~

PR #991, opened 2021-05-17
Dismiss configurationcache bindings if `OPT_PYTHON` is `OFF`

PeterBowman:
Even though `HAVE_ALL_PYTHON_HEADERS` is `TRUE`, don't attempt to generate bindings for the configurationcache plugin in case `OPT_PYTHON` was set to `OFF`. Otherwise, a compilation error like this may be thrown:

```
[ 67%] Building CXX object plugins/configurationcache/CMakeFiles/openravepy_configurationcache.dir/openravepy_configurationcache.cpp.o
/home/bartek/git/openrave/plugins/configurationcache/openravepy_configurationcache.cpp:16:10: fatal error: pyconfig.h: No such file or directory
   16 | #include <pyconfig.h>
      |          ^~~~~~~~~~~~
```


rdiankov:
thanks~

PR #992, opened 2021-05-27
fix _mapExtraGeometries values were shared after cloning

kanbouchou:
Overview
=============
- Due to shallow copy, original body and cloned body shared values (vector of ``GeometryInfoPtr``) of ``_mapExtraGeometries``.  This led to a surprise later that ``ConvertUnitScale`` on cloned body modifies geom of original body.

Fix
========

- deep copy values of ``_mapExtraGeometries``, just like we already do for ``_vGeometries``


rdiankov:
thanks~

PR #993, opened 2021-05-28
Skip creating bodies with empty name.

ziyan:


rdiankov:
@ziyan what if we have an external reference, which specifies the referenceUri and id, but not name?

rdiankov:
thanks~

PR #994, opened 2021-05-31
Support conversion of numpy scalar types to RapidJson

cv3d:
Supports numpy scalar inputs such as `numpy.string_("str")` or `numpy.uint8(5)`


rdiankov:
@cv3d one thing i worry about with serializing numpy structures is that we also don't deserialize them into numpy structures. So what comes in to openrave is different from what comes out when some component wants to read that information. And it will actually lead to more problems in the code that is using this since some objects can be numpy objects and others not, and it changes the behavior.

cv3d:
> what comes in to openrave is different from what comes out when some component wants to read that information

@rdiankov That is right, but I am afraid it is already happening in the same file just above this proposal. The block starting with `if (PyArray_Check(obj.ptr()) ) { ... }` is converting arrays into lists (more wild than this proposal), and then convert their internal scalar types to C++ ones (see `FillRapidJsonFromArray<T>` calls where `T` is float, double, int8_t etc), which is exactly what this proposal is doing, except that this proposal is for scalar types, not arrays. You can think of it as a complementary one.

That being said, I do not think it is OpenRAVE problem. It is the user responsibility to pass these types or not. If they pass them, then either to accept it (which is done for the arrays - part of the problem), or to raise an exception saying `Unsupported python class '%s' for object %s` (which is happening for the scalar types). All I am asking for here is homogeneity.

After all, which is better, to convert (for example `numpy.string_` to `std::string`), to raise? It does not make sense for me to accept Numpy arrays but reject their basic building blocks; Numpy scalars.

rdiankov:
@cv3d that's a good point, i didn't realize how numpy arrays had already snuck up in the codebase ;0/
given that the system is already doing this weird conversation, and i'm afraid of removing it since now it will break code, then i think it is ok for your change to go forward.
can you increase the PATCH level?
thanks

cv3d:
@rdiankov Bumped the patch version. Many thanks~

rdiankov:
thanks!

PR #995, opened 2021-06-03
skip multiply by identity rotation matrix for aabb

kanbouchou:
overview
=============

- often times ``tglobal`` is identity, or just has translational component, so skip rotating vertices which does nothing.
- with this change, computing aabb becomes 2x fast (took 33 us with production, 14us with feature for vertices with 10,000 vertices


rdiankov:
thanks~

PR #996, opened 2021-06-04
Fix parabolic interpolator

Puttichai:
Previously when doing `std::vector<ParabolicCurve>` to `std::vector<RampND>`, we use the duration of the first `ParabolicCurve` was used as the duration of `RampND`. However, it is also possible that some other `ParabolicCurve` has longer duration than the first (while the discrepancy is within the specified epsilon).

This pull request adds a fix so that we always use the longest duration of all `ParabolicCurve`s as the duration of `RampND`s.


rdiankov:
thanks~

PR #998, opened 2021-06-10
[WIP] add time conversion to msgpack unpacker

cielavenir:
Time spec is written in https://github.com/msgpack/msgpack/blob/master/spec.md#timestamp-extension-type

@ziyan Could you check if the concept is ok? It is possible my patch aims for mujin-specific format (if yes I need to close it). Also I suspect read64be() impl is dirty.

### DIFF
--- a/src/libopenrave/openravemsgpack.cpp
+++ b/src/libopenrave/openravemsgpack.cpp
@@ -22,6 +22,16 @@
 #include <msgpack.hpp>
 #include <rapidjson/document.h>
 
+static inline unsigned int read32be(const void *p){
+	const unsigned char *x=(const unsigned char*)p;
+	return x[3]|(x[2]<<8)|(x[1]<<16)|((unsigned int)x[0]<<24);
+}
+
+static inline unsigned long long int read64be(const void *p){
+	const unsigned char *x=(const unsigned char*)p;
+	return x[7]|(x[6]<<8)|(x[5]<<16)|((unsigned int)x[4]<<24)|( (unsigned long long int)(x[3]|(x[2]<<8)|(x[1]<<16)|((unsigned int)x[0]<<24)) <<32);
+}
+
 namespace msgpack {
 
 MSGPACK_API_VERSION_NAMESPACE(MSGPACK_DEFAULT_API_NS) {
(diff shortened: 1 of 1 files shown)


rdiankov:
i'm not sure this is a good idea....

cielavenir:
@rdiankov thank you for the comment

@ziyan could you tell me the policy of the content of webstack mount?

cielavenir:
fyi now we can emulate webstack policy in Python:

```
#!/usr/bin/python

from datetime import datetime
from struct import unpack

import json
import msgpack

import sys
from collections import OrderedDict

if sys.version_info[0]>=3:
    binstdin = sys.stdin.buffer
else:
    binstdin = sys.stdin

def ext_hook(code, data):
    # https://github.com/neuecc/MessagePack-CSharp/issues/264#issuecomment-393583730
    if code == -1:
        if len(data) == 4:
            secs, = unpack('>I', data)
            nsecs = 0
        elif len(data) == 8:
            data, = unpack('>Q', data)
            secs = data & 0x00000003ffffffff
            nsecs = data >> 34
        elif len(data) == 12:
            nsecs, secs = unpack('>IQ', data)
        else:
            raise AssertionError("Not reached")

        retAsDatetime = datetime.utcfromtimestamp(secs + nsecs / 1e9)
        retAsString = retAsDatetime.strftime('%Y-%m-%dT%H:%M:%S.%f')+'+09:00'
        return retAsString
    return msgpack.ExtType(code, data)

content = msgpack.load(binstdin, ext_hook=ext_hook, raw=False, object_pairs_hook=OrderedDict)
json.dump(content, sys.stdout, indent=2, separators=(',', ': '))
```

ziyan:
@cielavenir even though webstack outputs the timestamp, openrave has no use of it, so it doesn't need to parse it.

cielavenir:
ok for offline conversion I use the above Python impl (hopefully).

@ziyan thank you for confirmation.

cielavenir:
NOTE: it is arguable which webstack_archive should download, msgpack or json because json helps checking difference among controller backup history.

cielavenir:
anyway webstack_archive is not openrave, closing

PR #999, opened 2021-06-11
inherit CMAKE_MODULE_PATH

cielavenir:
We want to specify CMAKE_MODULE_PATH externally

/cc @kktk-KO


cielavenir:
The situation is the same as controllercommon/CMakeLists.txt:9

cielavenir:
@rdiankov Takuya has issues around custom cmake module path.

takuya-smz:
@rdiankov I have issues in overriding CMAKE_MODULE_PATH in .jhbuildrc

rdiankov:
thanks

PR #1001, opened 2021-06-28
wip: Speedup parameter planner parameters assignment

kanbouchou:
Summary
============

- this MR tries to reduce planning time by optimizing (and separating) assignment and post processing parameter loading 

Production behavior
=====================
- assignment operator was loading members in derived class by serialize->deserialize with xml format. This was taking hundreds of micro seconds.
- assignment operator was doing more than assignment, namely, it was parsing ``_sPostProcessingPlanner`` and ``_sPostProcessingParameters`` from ``_sExtraParameters``

Feature behavior
==============
- assignment operator(``=``) copies members by down casting, without going through xml string.
- parsing of ``_sPostProcessingPlanner`` and ``_sPostProcessingParameters`` are done explicitly by ``LoadExtraParameters``

Discussion points
================
- ``_ptarget`` of ``GraspSetParameters`` is copied by environment body index as it is the behavior in production. Do we have a better identifier such as name or id?

Future TODO
==============
- start using json string instead of xml.

### DIFF
--- a/include/openrave/planner.h
+++ b/include/openrave/planner.h
@@ -445,7 +445,9 @@ protected:
     ///
     /// This is called from assignment operator(=), each derived class is responsible for overriding this.
     /// overriding _Copy should copy member variables introduced in derived class, and also call Base class' _Copy to make sure member variables in base class is also copied
-    virtual void _Copy(const PlannerParameters& other);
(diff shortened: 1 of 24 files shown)


tgn3000:
Back then I did have a version that uses only json to assign planner parameters between derived classes.

PR #1002, opened 2021-06-29
clarify and fix _listNonCollidingLinks

kanbouchou:
There was following inconsistency
- In ``KinBody::_PostprocessChangedParameters``, a link is considered non colliding even if it is disabled.
- In ``Grabbed::ProcessCollidingLinks``, a link is not considered non colliding if it is disabled.

As a result, following was observed in production.
```
robot.Enable(True)
link0 = robot.GetLinks()[0]
link0.Enable(False)
robot.Grab(body)
assert link0.GetName() in robot.GetGrabbedInfo()[0].SerializeJSON()['ignoreRobotLinkNames']
link1 = robot.GetLinks()[1]
link1.Enable(False) # trigger ProcessCollidingLinks, link0 remain disabled.
assert link0.GetName() in robot.GetGrabbedInfo()[0].SerializeJSON()['ignoreRobotLinkNames'] # assert fails, link0 is not in ``ignoreRobotLinkNames`` anymore
```

In this MR, ``link0`` is not in ``ignoreRobotLinkNames`` right after Grab, and remain so when ``ProcessCollidingLinks`` is called.

### DIFF
--- a/src/libopenrave/kinbody.cpp
+++ b/src/libopenrave/kinbody.cpp
@@ -5721,12 +5721,14 @@ void KinBody::_PostprocessChangedParameters(uint32_t parameters)
             }
             else {
                 // add since it is disabled?
-                FOREACH(itgrabbed,_vGrabbedBodies) {
-                    GrabbedPtr pgrabbed = boost::dynamic_pointer_cast<Grabbed>(*itgrabbed);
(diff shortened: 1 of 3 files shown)


rdiankov:
thanks~

PR #1006, opened 2021-07-11
Fix unit saving in openrave json/msgpack scene

ziyan:
Unit was not being saved into json/msgpack scene files. This pull request fixes the issue.

cc @AntoineRyu


rdiankov:
thanks~

PR #1007, opened 2021-07-15
Fixed pybind11 (again)

cielavenir:


cielavenir:
interesting test is passing, assigning

rdiankov:
thanks~

PR #1008, opened 2021-07-15
Add explicit __deepcopy__ to all pickleable classes. Simplified some __deepcopy__ implementations.

cielavenir:
please check https://github.com/rdiankov/openrave/pull/1007 first

PR #1009, opened 2021-07-22
make sure to initialize pycollision

cielavenir:
detector is using this method heavily.

for pybind series, maybe please prioritize this one (then #1007 then #1008). I suppose this one has large impact not limited to detectors.


rdiankov:
thanks

PR #1010, opened 2021-07-27
Init plan return planner status

kanbouchou:
Background
-----------------------

- Failure from ``InitPlan`` used to be propagated as exception, but currently is just a log message.
- There is a bug that ``PlannerStatus::statusCode``, return value of ``PlanPath``  is treated like a bool, so for example returning  ``PS_FailedDueToCollision``, whose value is 0x30000, would evaluate to ``true``.

In this MR
----------------
- ``InitPlan`` returns ``PlannerStatus``, just like ``PlanPath`` so that the failure information can be treated in the same way at application side
- fixed a bug of how to check ``statusCode``


kanbouchou:
wip until compilation error is fixed and tested against after merging with latest production.

rdiankov:
thanks~

PR #1013, opened 2021-08-16
Made buildable without Boost_NO_BOOST_CMAKE

cielavenir:
OpenRAVE has build issue without Boost_NO_BOOST_CMAKE, this commit eliminates the necessity.

cf https://github.com/cielavenir/mujin_recruiting/blob/fb3b4b54631a4010b974fb86a1e1496cfe75e586/cookbooks/mujin/recipes/default.rb#L199


cielavenir:
btw openrave.pc still needs -DBoost_NO_BOOST_CMAKE=1 to be generated correctly, but our jhbuild system does not use openrave.pc anyway.

rdiankov:
thanks~

PR #1016, opened 2021-09-21
Expose Grab(..., setIgnoreBodyLinkNames) in RobotBase

strmio:
In the current implementation one of the `Grab()` function definitions in `KinBody` is not explicitly exposed in `RobotBase` and is shadowed by having definitions of other overloading functions under the same name. To allow calling `Grab(..., setIgnoreBodyLinkNames)` in `KinBody` through an instance of `RobotBase`, this patch adds an explicit definition of the function in `RobotBase` and a corresponding Python binding.


rdiankov:
thanks

PR #1019, opened 2021-10-08
Require C++14 and drop redundant braces for Clang

PeterBowman:
OpenRAVE cannot be compiled with Clang (tested with 9 and 11) after the last merge to master. There are two issues this patch aims to fix:

* Commit https://github.com/rdiankov/openrave/commit/0d27e082d67a2dda08236018c5c1ed6d7cc500b6 switched the C++11 requirement to C++14, but it did so by forcing the `-std=gnu++14` compiler flag in a way that affects both C and C++ sources:
  https://github.com/rdiankov/openrave/blob/748891e955b01c660afd4d9e869ded32f03b2205/CMakeLists.txt#L397
  This is causing trouble on C files compiled with Clang, seeing lots of `invalid argument '-std=gnu++14' not allowed with 'C'` errors. I don't know why `typeof` is mentioned in the comment, that commit seems to be related to the introduction of `std::shared_timed_mutex` only (standard C++14). I'm bringing back the `CMAKE_CXX_STANDARD` variable.
* The syntax `py::array_t<T> arr({value})` produces several "braces around scalar initializer" warnings (said braces are needless) and can escalate to an error if implicit narrowing conversions are involved, e.g. `non-constant-expression cannot be narrowed from type 'int' to 'pybind11::size_t' (aka 'unsigned long') in initializer list [-Wc++11-narrowing]`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@ cmake_policy(SET CMP0002 NEW)
 # http://www.cmake.org/cmake/help/cmake-2.6.html#policy:CMP0003
 cmake_policy(SET CMP0003 NEW)
 
+set(CMAKE_CXX_STANDARD 14) # C++14 for std::shared_timed_mutex
 # Use, i.e. don't skip the full RPATH for the build tree
 set(CMAKE_SKIP_BUILD_RPATH  FALSE)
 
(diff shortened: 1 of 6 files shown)


cielavenir:
[memo] (review ongoing)

CXX_EXTENSIONS is ON by default, so setting `CMAKE_CXX_STANDARD 14` selects gnu++14, not c++14

cielavenir:
for mujinian: this change is now also in production branch as well

PR #1022, opened 2021-10-21
Improve OSG viewer HUD text appearance

Hsifnus:
### **Background**

* The `osgText::Text` instance `_osgHudText`, used to display various types of HUD text around the top of the viewer screen in `osgviewerwidget.cpp`, did not use line breaking behavior, which resulted in the HUD text running out of the screen if too much text is present in a line of HUD text.
* The HUD text used default font and typically displays units in meters.

### **Changes**

* Enforced a maximum width on the HUD text, which causes OSG to break any lines of text exceeding the width.
* Changed HUD text font to be the same as the one used for labels, and changed displayed units to millimeters in order to improve appearance and readibility of the text.
* By default, OSG will try to break lines early based on any spaces or hyphens found. As such, in order to make line breaking behavior less intrusive on text readability, spaces inside coordinate pairs and other segments of the HUD text string are replaced with non-breakable spaces, and hyphens (which are used in negative numbers) are replaced with a logical negation (¬) sign.
  * Not too sure if this is the best solution in this case, but similar substitutes for the hyphen are rather limited inside the single-byte character space (the improved font used here is a Unicode font, but multi-byte characters do not render properly in the HUD text.)

### DIFF
--- a/plugins/qtosgrave/osgviewerwidget.cpp
+++ b/plugins/qtosgrave/osgviewerwidget.cpp
@@ -923,9 +923,9 @@ void QOSGViewerWidget::HandleRayPick(const osgUtil::LineSegmentIntersector::Inte
 
             if( !!item ) {
                 osg::Vec3d pos = intersection.getWorldIntersectPoint();
-                pos *= 1000.0;
+                pos *= 1000.0 * _metersinunit;
                 osg::Vec3d normal = intersection.getWorldIntersectNormal();
-                normal *= 1000.0;
+                normal *= 1000.0 * _metersinunit;
                 KinBody::LinkPtr link = item->GetLinkFromOSG(node);
                 std::string linkname;
                 if( !!link ) {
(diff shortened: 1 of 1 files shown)


plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -917,7 +923,9 @@ void QOSGViewerWidget::HandleRayPick(const osgUtil::LineSegmentIntersector::Inte
 
             if( !!item ) {
                 osg::Vec3d pos = intersection.getWorldIntersectPoint();
+                pos *= 1000.0 * _metersinunit;
                 osg::Vec3d normal = intersection.getWorldIntersectNormal();
+                normal *= 1000.0 * _metersinunit;
```
Puttichai, line 928: @Hsifnus @rdiankov Actually I think the normal vector should have unit length.

PR #1023, opened 2021-10-25
Smoother HUD font on initialization in qtosg

Hsifnus:
**Background:**
PR #1022 introduces an improved font for the HUD in qtosg. However, it was discovered during more testing that it is possible for the HUD font to be initialized in an aliased, jagged state if there are no pre-existing labels in the loaded scene. This is because GL blending is by default turned off in the HUD section of the scene graph.

**Changes:**
* Make `QOSGViewerWidget` use `OSGLODLabel`'s approach of loading the font to a static member so that all needed font loading can be performed in one spot inside `QtOSGViewer::QtOSGViewer`.
* Adjusted font resolution and enabled GL blending for HUD text to fix initialization aliasing issue.
* As suggested by Puttichai in #1022, removed mm scaling from normal vector when displayed in HUD text.

### DIFF
--- a/plugins/qtosgrave/osgviewerwidget.cpp
+++ b/plugins/qtosgrave/osgviewerwidget.cpp
@@ -931,7 +931,6 @@ void QOSGViewerWidget::HandleRayPick(const osgUtil::LineSegmentIntersector::Inte
                 osg::Vec3d pos = intersection.getWorldIntersectPoint();
                 pos *= 1000.0 * _metersinunit;
                 osg::Vec3d normal = intersection.getWorldIntersectNormal();
-                normal *= 1000.0 * _metersinunit;
                 KinBody::LinkPtr link = item->GetLinkFromOSG(node);
                 std::string linkname;
                 if( !!link ) {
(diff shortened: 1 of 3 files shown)


plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -936,7 +941,7 @@ void QOSGViewerWidget::HandleRayPick(const osgUtil::LineSegmentIntersector::Inte
                 if( !!geom ) {
                     geomname = geom->GetName();
                 }
-                _strRayInfoText = str(boost::format("mouse\xa0on\xa0%s:%s:%s: (%.2f,\xa0%.2f,\xa0%.2f), n=(%.2f,\xa0%.2f,\xa0%.2f)")%item->GetName()%linkname%geomname%pos.x()%pos.y()%pos.z()%normal.x()%normal.y()%normal.z());
+                _strRayInfoText = str(boost::format("mouse\xa0on\xa0%s:%s:%s: (%.2f,\xa0%.2f,\xa0%.2f), n=(%.5f,\xa0%.5f,\xa0%.5f)")%item->GetName()%linkname%geomname%pos.x()%pos.y()%pos.z()%normal.x()%normal.y()%normal.z());
```
braineo, line 944: do we need .5f?

PR #1024, opened 2021-10-25
Clean up parabolicsmoother2 logs

Puttichai:
### Summary

This pull request cleans up some logging in `parabolicsmoother2`. Also, now allowing the smoother to run with `_nMaxIterations == 0`, in which case the smoother will only do the initial time-parameterization and no shortcutting.


rdiankov:
thanks~

PR #1025, opened 2021-11-03
remove extra ReadKinBodyURI call

KazutoMurase:
Somehow `ReadKinBodyURI` is called twice. I think 1st call is not needed.
And GIL is never unlocked in the 2nd call, which is not efficient.


KazutoMurase:
@rdiankov

rdiankov:
thanks~

PR #1027, opened 2021-11-04
set py_err_class to propagate std::runtime_error properly

KazutoMurase:
@rdiankov 
 When `std::runtime_error` is thrown, boost::python cannot propagate error message properly as "_std_runtime_error_" doesn't have attribute "py_error_class".

The error message caught in client is as follows:
```
'_std_runtime_error_' object has no attribute 'py_err_class'
```

### DIFF
--- a/python/openravepy.__init__.py
+++ b/python/openravepy.__init__.py
@@ -44,7 +44,7 @@ Available methods for an exception e are
 if openravepy_int.__pythonbinding__ == 'pybind11':
     from .openravepy_int import _OpenRAVEException as OpenRAVEException
 else:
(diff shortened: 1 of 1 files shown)


KazutoMurase:
The above error message is generated here:

boostpython_bindings.h
```c++
    static void translate( const T& err )
    {
        py::object pimpl_err( err );
        py::object pyerr_class = pimpl_err.attr( "py_err_class" );
        py::object pyerr = pyerr_class( pimpl_err );
        PyErr_SetObject( pyerr_class.ptr(), py::incref( pyerr.ptr() ) );
    }
```

rdiankov:
thanks~

PR #1029, opened 2021-11-16
Fix serialization of readable interfaces in trajectory

yoshikikanemoto:
`deserialize` always reads type information of readable. however, `serialize` wrote its type only when it was `HierarchicalXMLReadable`


yoshikikanemoto:
@rdiankov ping. this is a fix to serialize/deserialize multiple `ReadableInterface`s stored in trajectory.

rdiankov:
thanks~

PR #1030, opened 2021-11-17
wip: print fcl cache size

kanbouchou:


rdiankov:
just curious, what's the point of tracking this max?

kanbouchou:
> just curious, what's the point of tracking this max?

I originally wanted to print a message every time size of ``_bodymanagers`` or ``_envmanagers`` was changed, but it was a bit too much. So, I thought I would print at least the max size of them, and max size does not necessary happen at the time of destruction. If these values are very high compared to number of bodies in the env, that might indicate a bug in cache management or memory leak.

rdiankov:
ok, let's see what happens~

PR #1031, opened 2021-11-22
Fix boost python exception message signature

cielavenir:
@@@ This PR is concept demo @@@

python exception message field is in https://github.com/python/cpython/blob/2.7/Objects/exceptions.c#L364

and it is string, not function. you can confirm it by:

```
In [1]: try:
   ...:     1/0
   ...: except Exception as e:
   ...:     print(e.message)
   ...:     
integer division or modulo by zero
```

this patch changes message member's behavior to be the same as python standard exception:

```
In [1]: try:
   ...:     robot.SetDOFValues([])
   ...: except OpenRAVEException as e:
   ...:     print(e.message)
   ...:     
openrave (InvalidArguments): [void openravepy::PyKinBody::SetDOFValues(boost::python::api::object):3481] Passed in values to SetDOFValues have 0 elements, but robot has 5 dof
```

except the patch, the output becomes this:

```
<bound method _OpenRAVEException.message of <OpenRAVEException('openrave (InvalidArguments): [void openravepy::PyKinBody::SetDOFValues(boost::python::api::object):3481] Passed in values to SetDOFValues have 0 elements, but robot has 5 dof','InvalidArguments')>>
```

- Merge EITHER #1028 OR #1031. Do not merge both.
- Merging this causes all our codebase incompatibility including planningcommon.


rdiankov:
thanks~

cielavenir:
what was lucky is that I only need to fix planningcommon

PR #1032, opened 2021-11-22
fixed pyann and convexdecompositionpy as well

cielavenir:


cielavenir:
@rdiankov actually std::runtime_error is affected, could you work on this?

/cc @KazutoMurase

rdiankov:
thanks

PR #1033, opened 2021-12-03
Reduce ignore robot link names20211203

ziyan:


rdiankov:
@ziyan feels like if the body is disabled, we shouldn't event compute `_mapLinkIsNonColliding`...
Only problem is when a user enables/disables a body while being grabbed, then e we would have to change the grabbed info state ;0/

rdiankov:
nice catch~

PR #1034, opened 2021-12-03
fix GetKeywords/SetKeywords

yoshikikanemoto:
`env.GetKeywords` and `env.SetKeywords` were not working


rdiankov:
thanks~

python/bindings/openravepy_int.cpp
```diff
@@ -2764,16 +2764,16 @@ py::object PyEnvironmentBase::GetDescription() const
     return ConvertStringToUnicode(_penv->GetDescription());
 }
 
-void PyEnvironmentBase::SetKeywords(const std::vector<std::string>& sceneKeywords)
+void PyEnvironmentBase::SetKeywords(py::object oSceneKeywords)
```
cielavenir, line 2767: @yoshikikanemoto not sure, was this one required? only GetKeywords fix was required for me.

(note: I can test only pybind11 now)

yoshikikanemoto: yes, it was required

PR #1036, opened 2021-12-06
optimize grabbed object collision check

kanbouchou:
Summary
===============

- this MR improves self collision check time, especially when many (disabled) objects are grabbed
- Improves collision check time from 70 us to 23 us, when grabbing 36 disabled objects.

Optimizations
===============
- downcast from ``UserDataPtr`` to ``GrabbedPtr`` was expensive, so ``_vGrabbedBodies``  now holds ``GrabbedPtr``.
- minimize expensive ``_pgrabbedbody.lock()`` cache the result (``vLockedGrabbedBodiesCache``)
- check collision once per pair of grabbed bodies, previously we were checking twice (obj_b against obj_a, obj_a against obj_b)
- minimize shared pointer dereferencing
- Prune collision check early if grabbed objects are disabled (@ziyan )

Ideas for future work
================
- Stop using weak pointer for ``_pgrabbedbody`` and other things because ``lock`` is expensive
- have better data structure for ``_listNonCollidingLinks`` which is faster to search / lookup.
- further reduce use of shared pointer.

cc @ziyan @Puttichai


rdiankov:
@kanbouchou is this merge ready for review?

kanbouchou:
> @kanbouchou is this merge ready for review?

Yes, it's ready

rdiankov:
thanks~

PR #1037, opened 2021-12-15
Allow update from info to set empty URIs

ziyan:


CMakeLists.txt
```diff
@@ -4,8 +4,8 @@ set( CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS TRUE )
 
 # Define here the needed parameters
 set (OPENRAVE_VERSION_MAJOR 0)
-set (OPENRAVE_VERSION_MINOR 86)
-set (OPENRAVE_VERSION_PATCH 1)
+set (OPENRAVE_VERSION_MINOR 87)
```
rdiankov, line 7: @ziyan perhaps don't need minor change, only patch change

ziyan: Done.

rdiankov:
thanks~

PR #1038, opened 2021-12-16
Draft: Lazy Computation of List Non-Colliding Links

Puttichai:
### Motivation

Whenever a robot grabs an object (via `robot.Grab(target, ...)` command), a `Grabbed` object is created to store grab-related information. One thing that gets computed right away is `_listNonCollidingLinks`, which stores information of which robot links do not collide with the object when grabbing takes place. This `_listNonCollidingLinks` will then be used later on when `robot.CheckSelfCollision` is called to check if there are new robot links that previously did not collide with the grabbed object but now do so.

One of the problems with this is that `_listNonCollidingLinks` computation may be expensive, for example when the robot has many links, as it requires many check-collision calls. Furthermore, `_listNonCollidingLinks` might not even be used at all if, for example, `robot.CheckSelfCollision` never takes place or the grabbed object is never enabled.

Another problem is due to the current implementation. By definition, an element of `_listNonCollidingLinks` is a link `L` that _touches_ the grabbed object at the time of grabbing. Whether or not the link `L` touches the object **does not** depend on the enable state of `L`. However, in the current implementation, extra computation of `_listNonCollidingLinks` is always triggered whenever the enable state of the target changes from disabled to enabled.

This pull request aims to clean up the implementation related to `Grabbed` class and `Grab`-related functions in order to ensure consistency as well as potentially improve performance.

#### Class `Grabbed`
- **Member Variable Names**
  There are changes in member variable names in order to make their meanings clearer:
  - `_pgrabbedbody` -> `_pGrabbedBody`
  - `_plinkrobot` -> `_pGrabbingLink`
  - `_troot` -> `_tRelative`
  - `_listNonCollidingLinks` -> `_listNonCollidingLinksWhenGrabbed`
  - `_setRobotLinksToIgnore` -> `_setGrabberLinksToIgnore`
  - `_vattachedlinks` -> `_vAttachedToGrabbingLink`
  -  `_enablecallback` and `_mapLinkIsNonColliding` were removed since no longer used.

- **Computation of `_listNonCollidingLinksWhenGrabbed`**
  Now we switch to using a lazy approach. That is, we will only compute `_listNonCollidingLinksWhenGrabbed` when it is needed (i.e. for `CheckSelfCollision`). Initialization of a `Grabbed` object now does not trigger computation of `_listNonCollidingLinksWhenGrabbed`.

  In order to be able to perform the computation correctly at a later time, it becomes necessary that we save the states of the grabber and the grabbed object at the moment of `Grab` call. This saved states will then be restored temporarily for `_listNonCollidingLinksWhenGrabbed` computation (in `ComputeListNonCollidingLinks` function).

  And since the result of computation of `_listNonCollidingLinksWhenGrabbed` does not depend on any link's enable states, `_enablecallback` is removed.

#### `Grab` functions

- There are three signatures of `Grab` functions. Two of them implement their own grabbing procedure that are, however, slightly different. (The difference is how they handle multiple, consecutive `Grab` call with the same object to be grabbed.)

  In this pull request, the implementation is now unified in `bool Grab(KinBodyPtr, LinkPtr, const std::set<int>&)`. The other two functions will just call this with correct arguments.

### Notes

- Information of `_listNonCollidingLinksWhenGrabbed` is inevitably lost  during a conversion from `Grabbed` to `GrabbedInfo`. Therefore, it is important that when initializing a grabbed state of a robot from` GrabbedInfo`, the robot must be first set to some proper configuration so that `_listNonCollidingLinksWhenGrabbed`, which might be computed later, does not exclude the robot links that should not be ignored during collision checking.


rdiankov:
thanks~

PR #1039, opened 2021-12-17
Bullseye

ziyan:


cielavenir:
@rdiankov actually merging https://github.com/rdiankov/openrave/pull/1019 into production as well is enough.

ziyan:
> @rdiankov actually merging #1019 into production as well is enough.

Are these changes needed? I saw you are using them in the branch file @cielavenir

cielavenir:
not needed but keeping production being superset of master is important

rdiankov:
thanks~

PR #1044, opened 2022-01-26
[quick] handle loading of GT_None

kanbouchou:
Issue this PR is fixing
======================
When we saved a scene with ``GT_None`` geom type to a file, the saved file could not be opened by openrave.

Resolution
==============
``GT_None`` is serialized to empty string by ``_GetGeometryTypeString`` via ``KinBody::GeometryInfo::SerializeJSON``, however, ``KinBody::GeometryInfo::DeserializeJSON`` was throwing exception on empty string.

In this PR, empty string for geom type is parsed as ``GT_None`` so that serialization and deserialization behave in consistent manner.


rdiankov:
thanks

PR #1045, opened 2022-01-26
Propagate boost::filesystem::error properly

cielavenir:
my previous patch caused this `runtime_error: <unprintable runtime_error object>
runtime_error()`

(would you mind tell me if my boost::python exception handling is correct?)


rdiankov:
sure, thanks~

PR #1050, opened 2022-02-03
fixed unexpected invalidargument error from env.CheckCollision(link1, link2, report=collisionReport)

MiraiHattori:
Fixed `bool PyEnvironmentBase::CheckCollision(object o1, object o2, PyCollisionReportPtr pReport)` python bindings when checking collision between links.

Before this change, `env.CheckCollision(link1, link2, report=report)` fails with `invalid argument 1` message.
After this change, it is expected to return True/False without failure.


cielavenir (approved):

rdiankov:
thanks~

PR #1051, opened 2022-02-03
fix not updating FCLCollisionManagerInstance properly on body removal

kanbouchou:
Background
============

- This MR fixes a bug related to update of ``FCLCollisionManagerInstance``  cache which can cause wrong result being returned from ``CheckCollision`` when robot grabs and releases different bodies in env while env body indices are changed.
- unit test for this is submitted separately

Overview
==========
- Before https://github.com/rdiankov/openrave/pull/954, environment body index (so called env body id back then) was monotonically increasing, but after pr #954 it can be recycled to allow constant time access to containers of kin bodies.
- fclrave stores identity of kin body by environment body index, but this was not updated properly
 - for ``_envmanagers``, ``body`` was removed from the value of map, but key (vector of env body indices), corresponding to the removed body was not updated
 - for ``_bodymanagers``, entry corresponding to ``body`` was removed, but ``body`` was not removed from ``FCLCollisionManagerInstance`` of other entries.
 - These changes are in ``FCLCollisionChecker::RemoveKinBody``

Others
========
- improved log printing logic so that there is no false positive warnings.
- improved log messages to be more informative.


rdiankov:
nice catch!

PR #1052, opened 2022-02-06
Use py::list for vLinkInfo etc

cielavenir:
- KinBody::LinkInfoPtr cannot be converted to Python type
- Even if I fix the type (see 374f88a6d), `bodyInfo._vLinkInfos.append()` will not work. But fixing in stl.h way (using PYBIND11_MAKE_OPAQUE / py::bind_vector, see 6cdd5eeb3 and 5b2eb55a8) now breaks `bodyInfo._vLinkInfos=[]` (both see RealTimeRobotTask3.UpdateObjects).

We have to use py::object (or py::list).

test_yxx was failing due to this issue.


rdiankov:
thanks

PR #1053, opened 2022-02-08
disabled isBranching condition

Puttichai:
### Summary

The function `_OptimizePathSingleGroupShift` in `linearsmoother.cpp` is a function specifically written to work with robots with certain kinematics structure. 

It seems that originally the cases in which this function should be used were detected by the condition `_nUseSingleDOFSmoothing == 3`. Then `isBranching` was added later. However, as it turns out, `isBranching` is not a sufficient condition to decide whether `_OptimizePathSingleGroupShift` should be called. This new variable may lead to unexpected calls to `_OptimizePathSingleGroupShift`, which in the end may result in a problematic path.

I think it is best to disable the use of `isBranching` for now.

cc. @AntoineRyu @kanbouchou @eisoku9618

### DIFF
--- a/plugins/rplanners/linearsmoother.cpp
+++ b/plugins/rplanners/linearsmoother.cpp
@@ -128,17 +128,21 @@ public:
             dReal totaldist = 0;
 
             bool isBranching = false;
-            if(!!_probot && _nUseSingleDOFSmoothing == 1) {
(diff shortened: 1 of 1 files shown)


rdiankov:
thanks~

PR #1056, opened 2022-02-14
add c++17 compatibility

cielavenir:
(maybe because gcc11 defaults to gnu++17) log4cxx is std::shared_mutex-enabled, so we need to compile openrave under c++17 (currently only when specified cmake -DCMAKE_CXX_STANDARD=17). It caused some incompatibility but I fixed them.

Note: c++17+clang combination requires Python3 because Python2 header has incompatible `register` keyword (on gcc it is warning).


rdiankov:
thanks

cielavenir:
by the way

> "we need to compile openrave under c++17"

seems this restriction was lifted since https://github.com/apache/logging-log4cxx/commit/426de9be50dd73ec251da7029b62145e55c2b593 (log4cxx 1.0.0)

cielavenir:
https://github.com/rdiankov/openrave/pull/1420#issuecomment-2292524887

PR #1060, opened 2022-02-28
WIP: robotconnectedbody: use at() to access vector elements

cielavenir:
to throw exception rather than segmentation fault.

background: in mujinsimulator we want to remove env-lock from robot.GetConnectedBodyActiveStates(). But (although the codepath should not cause segmentation fault) it could cause segmentation fault. In such case, with this MR we will just get exception. for detail see mujinsimulator MR 968.

test in autotester pipelineid=240397

### DIFF
--- a/src/libopenrave/robotconnectedbody.cpp
+++ b/src/libopenrave/robotconnectedbody.cpp
@@ -334,7 +334,7 @@ void RobotBase::ConnectedBody::GetResolvedLinks(std::vector<KinBody::LinkPtr>& l
     RobotBasePtr pattachedrobot = _pattachedrobot.lock();
     if( !!pattachedrobot ) {
         for(size_t ilink = 0; ilink < _vResolvedLinkNames.size(); ++ilink) {
-            links[ilink] = pattachedrobot->GetLink(_vResolvedLinkNames[ilink].first);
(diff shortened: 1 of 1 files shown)


src/libopenrave/robotconnectedbody.cpp
```diff
@@ -334,7 +334,7 @@ void RobotBase::ConnectedBody::GetResolvedLinks(std::vector<KinBody::LinkPtr>& l
     RobotBasePtr pattachedrobot = _pattachedrobot.lock();
     if( !!pattachedrobot ) {
         for(size_t ilink = 0; ilink < _vResolvedLinkNames.size(); ++ilink) {
-            links[ilink] = pattachedrobot->GetLink(_vResolvedLinkNames[ilink].first);
+            links.at(ilink) = pattachedrobot->GetLink(_vResolvedLinkNames.at(ilink).first);
```
rdiankov, line 337: @cielavenir Since we're iterating with `vResolvedLinkNames.size()`, then `_vResolvedLinkNames[ilink]` should always be valid, and there's no reason to use `at`. Can you clean up this MR? thanks

cielavenir:
mujinsimulator new grabber is merged, closing this, thanks~

PR #1061, opened 2022-03-01
fix wrong description

eisoku9618:
`OpenRAVE::Link::GetTransform()` has this description,

```cpp
        /// \brief Return the current transformation of the link in the world coordinate system.
        inline const Transform& GetTransform() const {
            return _info._t;
        }
```

while `OpenRAVE::LinkInfo::_t` has this description.

```cpp
        Transform _t; ///< the current transformation of the link with respect to the body coordinate system
```

Based on the current implementation, the former description is correct.

```python
body.SetTransform([1, 0, 0, 0, 0, 0, 0])
print(body.GetLinks()[0].GetInfo().SerializeJSON()['transform'])
# [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]

body.SetTransform([1, 0, 0, 0, 0, 0, 1]);
print(body.GetLinks()[0].GetInfo().SerializeJSON()['transform'])
# [1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0]
```

`_t` is not with respect to the body frame but with respect to the world frame.

### DIFF
--- a/include/openrave/kinbody.h
+++ b/include/openrave/kinbody.h
@@ -900,7 +900,7 @@ public:
         }
 
 private:
-        Transform _t; ///< the current transformation of the link with respect to the body coordinate system
+        Transform _t; ///< the current transformation of the link with respect to the world coordinate system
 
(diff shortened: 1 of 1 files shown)


yoshikikanemoto:
looks good to me

rdiankov:
thanks

PR #1062, opened 2022-03-03
WIP: compute a bounding box which is aligned with aribitrary orientation

eisoku9618:
In current KinBody AABB computation,
1. compute geometry bounding box aligned with the link orientation for each geometry
2. combine multiple geometry bounding boxes of a link for each link (blue box in the figure)
3. combine multiple link bounding boxes of a kinbody (big black box)

Screenshot from 2022-03-03 10-57-57

- kinematics tree
   - kinbody1
      - link1-1
         - geometry1-1-1
         - geometry1-1-2
      - link1-2
         - geometry1-2-1

So because of No. 1, when `GeometryInfo::_t`, which is `Local transformation of the geom primitive with respect to the link's coordinate system`, is not identity rotation matrix like the above figure, computed AABB becomes big.

I would like to get the red box instead, so I add `ComputeAABBOnTransform` function, which allows user to specify which axis bounding box should be aligned with. (Implementing OBB is also ok, but just spefiying aligned axis is enough for my case.)


eisoku9618:
```xml
<KinBody name="TestChain">
  <Body name="Body1" type="dynamic">
    <Translation>0 0 0</Translation>
    <RotationAxis>1 0 0 0</RotationAxis>
    <Geom type="box">
      <Extents>1 1 1</Extents>
      <Translation>1 0 0</Translation>
      <RotationAxis>0 1 0 -45</RotationAxis>
    </Geom>
  </Body>
</KinBody>
```

For example,

```python
referenceGeom = body.GetLinks()[0].GetGeometries()[0]
geomTransformInWorld = poseMult(body.GetLinks()[0].GetTransform(), referenceGeom.GetTransform())
body.ComputeAABBFromTransform(geomTransformInWorld).extents() # array([2, 1, 2])
body.ComputeAABBOnTransform(geomTransformInWorld).extents()  # array([1, 1, 1])
```

rdiankov:
@eisoku9618 you are only using tReference's rotation to get the axes to align to right? And the AABB position returned is still in the world coordinate system right? In that case, it is confusing to pass in a `Transform` since it also has position, but that position is not used. Instead can you specify only a quaternion to define the coordinate system? Perhaps even call it:

```
OrientedBox<dReal> ComputeOBBOnAxes(const Vector& quat, ...)
```

That way you can return the oriented bounding box (which is a transform and extents) and the function will be non-ambiguous.

eisoku9618:
> And the AABB position returned is still in the world coordinate system right?

No, the returned position (and extents) are w.r.t. `tReference`. In order to do so, I need to pass position too.

- `ComputeLocalAABB`: in the body frame
- `ComputeAABB`: in the world frame
- `ComputeAABBOnTransform`: in the specified frame `tReference`

rdiankov:
err... then why not just do `ComputeAABBFromTransform(tReference.inverse())`?

rdiankov:
I see that `KinBody::ComputeAABBFromTransform` is not doing the tightest thing possible, it should also be calling link's `KinBody::ComputeAABBFromTransform` to get the aabb and combine. So I would focus your pull request on fixing it rather than creating a new function.
thanks

eisoku9618:
@rdiankov I thought so at the beginning, but I think it is impossible to change `KinBody::ComputeAABBFromTransform` like that without breaking API. It is because `KinBody::ComputeAABBFromTransform` (and `ComputeLocalAABB` and `ComputeAABB`) returns position and extention information only. It doesn't return any information about reference frame's rotation, and actually the reference frame is world and the extentions are aligned with world.

What I need with this MR is a bounding box which is not aligned with the world coordinate system, like the following red box.

image

rdiankov:
@eisoku9618 actually it works just fine. please see my changes (try using `ComputeOBBOnAxes(quat)`). i also took this time to cleanup the class names in geometry.h, which break some of the planning compatibility stuff, so i'll just merge it for now. 
If somehow it still doesn't satisfy what you want, let me know.

example of ComputeOBBOnAxes:
image

thanks,

eisoku9618:
@rdiankov Thank you, it worked for me with https://github.com/rdiankov/openrave/pull/1063

PR #1063, opened 2022-03-07
return relative pose

eisoku9618:
related to https://github.com/rdiankov/openrave/pull/1062

I think we want to return relative pose to kinbody


src/libopenrave/kinbody.cpp
```diff
@@ -1791,12 +1791,13 @@ AABB KinBody::ComputeAABBFromTransform(const Transform& tBody, bool bEnabledOnly
 
 OrientedBox KinBody::ComputeOBBOnAxes(const Vector& quat, bool bEnabledOnlyLinks) const
 {
-    Transform tinv; tinv.rot = quatInverse(quat);
-    AABB ab = ComputeAABBFromTransform(tinv, bEnabledOnlyLinks);
+    const Vector relativeQuat = quatMultiply(quatInverse(GetTransform().rot), quat); // relative quat from this kinbody to the given quat w.r.t. world
```
rdiankov, line 1794: That changes the function signature quite significantly, making it hard to use and understand. If you really want to take into account the quat that is relative to the current transform why not do `body.ComputeOBBOnAxes(quatMultiply(myrelquat, body.GetTransform().rot)`?

eisoku9618: @rdiankov sorry my lack of explanation, but current production branch `ComputeOBBOnAxes` has 2 issues.

### issues
1. returned OrientedBox's position is relative to this kinbody while returned OrientedBox's rotation is **not** relative to this kinbody
2. `ComputeAABBFromTransform` called inside `ComputeOBBOnAxes` depends on this kinbody's rotation while `tinv` in `ComputeOBBOnAxes` doesn't take it into consideration

Both issues are not revealed when kinbody's rotation is identity, and this is an example of these.

```xml
<KinBody name="myBody">
  <Body name="1stBody" type="dynamic">
    <Translation>0 0 1</Translation>
    <RotationAxis>0 0 1 45</RotationAxis>
    <Geom type="box">
      <Extents>0.4714 0.2357 0.2</Extents>
      <transparency>1</transparency>
    </Geom>
  </Body>
  <Body name="2ndBody" type="dynamic">
    <Translation>0 0 1</Translation>
    <Geom type="box">
      <Extents>0.5 1 0.2</Extents>
      <diffuseColor>0 0 1</diffuseColor>
      <transparency>0.5</transparency>
    </Geom>
  </Body>
</KinBody>
```

image

The bounding box which I want to get with `ComputeOBBOnAxes` is the red one, but we cannot get this red extents (`0.5`x`1.0`) with any input `quat` argument.

image


With this MR,
1. input `quat` argument: absolute orientation of the desired bounding box
2. output `OrientedBox`'s pose is relative to this kinbody
3. output `OrientedBox`'s extents is w.r.t. `OrientedBox`'s pose frame (i.e. in local frame)

rdiankov:
@eisoku9618 You are right, there was a bug in the implementation. I think you were on the right track, except I didn't like the double inverse in your implementation, so I just committed a simpler version. It's already pushed to `production` branch, so please take a look.

eisoku9618:
@rdiankov Thank you. I confirm that It works for my examples. (Your change is not only removing double inverse. The output OrientedBox's pose has been changed from relative to kinbody to relative to world.)

rdiankov:
That's correct, it should be returning the box in the world with the orientation specified by quatWorld

eisoku9618:
@rdiankov Thank you!

PR #1067, opened 2022-03-16
Remove GetEndEffectorTransform api

kanbouchou:
Background
==========

``GeEndeffectorTransform`` returns manip's transform instead of endeffector's. This is confusing and I spent some time debugging a code because of my assumption that ``GeEndeffectorTransform`` returns endeffector's position. To avoid causing same confusion in the future to other people, I am proposing to remove the api.


rdiankov:
thanks~

PR #1069, opened 2022-03-28
fix: HUD text offset and control button avoidance

Hsifnus:
This PR fixes a bug where the HUD text in `osgviewerwidget` will render underneath the control buttons that can be optionally added via `QtOSGViewer::_CreateControlButtons`, obstructing some of the displayed text.

**Changes**

* Maintain a 2d vector offset property within `osgviewerwidget` that can be access and modified via a getter/setter pattern.
  * This property, `_vecTextScreenOffset`, affects the position of the upper left corner of the HUD text area (which will subsequently adjust its width as to cover available space going towards the right side).
* Use the getter/setter of  `_vecTextScreenOffset` to offset the HUD text away from the control buttons when `QtOSGViewer::_CreateControlButtons` is called.
* Expose `SetHUDTextOffset` as a viewer command for testing or external application use.

### DIFF
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -170,6 +170,9 @@ QtOSGViewer::QtOSGViewer(EnvironmentBasePtr penv, std::istream& sinput) : QMainW
     RegisterCommand("RotateCameraYDirection", boost::bind(&QtOSGViewer::_RotateCameraYDirectionCommand, this, _1, _2),
     "Rotates the camera around the current focal point in the direction of the screen y vector (in world coordinates). The argument thetaY is in radians -pi < dy < pi.");
 
+    RegisterCommand("SetHUDTextOffset", boost::bind(&QtOSGViewer::_SetHUDTextOffsetCommand, this, _1, _2),
+    "Sets the screen offset of the HUD's text. Default value when no control buttons are present is (10.0, 0.0).");
+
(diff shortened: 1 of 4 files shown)


plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -863,6 +866,10 @@ void QtOSGViewer::_CreateControlButtons()
     qvBoxLayout->setAlignment(Qt::AlignTop);
     qvBoxLayout->heightForWidth(40);
 
+    //  Move HUD text out of way of control buttons if needed
+    osg::Vec2 hudTextOffset = _posgWidget->GetHUDTextOffset();
+    _posgWidget->SetHUDTextOffset(std::max(60.0f, hudTextOffset.x()), hudTextOffset.y());
```
braineo, line 871: can we move it to where `_CreateControlButtons` is called?

plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -1318,6 +1330,18 @@ bool QtOSGViewer::_PanCameraYDirectionCommand(ostream& sout, istream& sinput)
     return true;
 }
 
+bool QtOSGViewer::_SetHUDTextOffsetCommand(ostream& sout, istream& sinput)
```
braineo, line 1333: not sure why we still want this command

braineo (changes requested):

braineo (approved):

braineo:
thanks

PR #1077, opened 2022-04-14
WIP: remove user data on BodyRemoved

eisoku9618:
`ODESpace::KinBodyInfo::vlinks` contains `KinBody::LinkWeakPtr` and it is used in `ODESpace::_Synchronize`. When a kinbody in openrave env is changed (e.g. Remove -> Clone -> Add), a `KinBody::LinkWeakPtr` stored in `ODESpace::KinBodyInfo::vlinks` becomes a null pointer so assertion happens in `ODESpace::_Synchronize`. We need to remove user data in that situation so that `ODESpace::GetCreateInfo` calls `InitKinBody` and update `ODESpace::KinBodyInfo::vlinks`. We do the same thing in fclspace.

## note to self
When we sync environment A to environment B and both environment have a same name kinbody but they have different KinematicsGeometryHash, we remove the kinbody from B and then clone the kinbody in A to the kinbody in B and then add the kinbody to B. So kinbody's user data is not cleared and kept. However `ODESpace::GetCreateInfo` depends user data to judge whether we use cached `ODESpace::KinBodyInfo` or we update `ODESpace::KinBodyInfo`, so we needed to clear user data.

### DIFF
--- a/plugins/oderave/odespace.h
+++ b/plugins/oderave/odespace.h
@@ -200,6 +200,7 @@ public:
 
             _geometrycallback.reset();
             _staticcallback.reset();
+            _bodyremovedcallback.reset();
         }
 
         KinBodyPtr GetBody() {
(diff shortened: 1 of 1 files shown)


eisoku9618:
I will add a test for this

rdiankov:
thanks

PR #1078, opened 2022-04-15
check and normalize direction vector length

kanbouchou:
Background
============

- This MR check that direction vector of some IKParameter is of unit length. Without this, non-unit length direction vector can cause problems in unexpected ways later, so better to catch here.
- First check if direction vector is close to unit length, and if far, assert. This logic is taken from quaternion check in ``geometry.h``

### DIFF
--- a/include/openrave/openrave.h
+++ b/include/openrave/openrave.h
@@ -1172,9 +1172,15 @@ public:
     }
     inline void SetDirection3D(const Vector& dir) {
         _type = IKP_Direction3D; _transform.rot = dir;
+        dReal length2 = _transform.rot.lengthsqr3();
+        MATH_ASSERT(length2 > 0.99 && length2 < 1.01); // make sure it is at least close
+        _transform.rot.normalize4();
     }
     inline void SetRay4D(const RAY& ray) {
         _type = IKP_Ray4D; _transform.trans = ray.pos; _transform.rot = ray.dir;
+        dReal length2 = _transform.rot.lengthsqr3();
+        MATH_ASSERT(length2 > 0.99 && length2 < 1.01); // make sure it is at least close
+        _transform.rot.normalize4();
     }
     inline void SetLookat3D(const Vector& trans) {
         _type = IKP_Lookat3D; _transform.trans = trans;
@@ -1182,6 +1188,9 @@ public:
     /// \brief the ray direction is not used for IK, however it is needed in order to compute the error
     inline void SetLookat3D(const RAY& ray) {
         _type = IKP_Lookat3D; _transform.trans = ray.pos; _transform.rot = ray.dir;
+        dReal length2 = _transform.rot.lengthsqr3();
+        MATH_ASSERT(length2 > 0.99 && length2 < 1.01); // make sure it is at least close
+        _transform.rot.normalize4();
     }
     inline void SetTranslationDirection5D(const RAY& ray) {
         _type = IKP_TranslationDirection5D; _transform.trans = ray.pos; _transform.rot = ray.dir;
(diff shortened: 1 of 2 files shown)


rdiankov:
@kanbouchou it should be `normalize3`

kanbouchou:
> @kanbouchou it should be `normalize3`

right, let me fix that.

Puttichai:
@kanbouchou What aboud `SetValues` function?

kanbouchou:
> @kanbouchou What about `SetValues` function?

oh I missed it. Let me fix that as well.

include/openrave/openrave.h
```diff
@@ -1076,7 +1076,15 @@ inline T NormalizeCircularAnglePrivate(T theta, T min, T max)
...
     return theta;
 }
-
+/// \brief ensures first three elements of vector forms a unit length vector
+/// first checks if input direction vector is close to unit length, and raise if not because it is most likely a bug on application side
+/// \param direction[inout] direction vector to be checked, if far from being unit length, most likely a bug on application side so assert. Otherwise, normalize and return.
+inline void Ensure3DVecUnitLength(Vector& direction)
+{
+    const dReal length2 = direction.lengthsqr3();
+    MATH_ASSERT(length2 > 0.99 && length2 < 1.01); // make sure it is at least close
```
rdiankov, line 1085: @kanbouchou problem with this is that you don't get the stack of which function called it. Doing a MACRO would be better...

kanbouchou: yeah, I thought about it but decided to go with inline function thinking most env has only one or two ik parameterization types so should be relatively easy to guess the stack.

Sure, I'll make this a macro and update this MR.

include/openrave/openrave.h
```diff
@@ -1172,20 +1190,24 @@ class OPENRAVE_API IkParameterization
     }
     inline void SetDirection3D(const Vector& dir) {
         _type = IKP_Direction3D; _transform.rot = dir;
-    }
+        ENSURE_3DVEC_UNIT_LENGTH(_transform.rot);
+     }
     inline void SetRay4D(const RAY& ray) {
         _type = IKP_Ray4D; _transform.trans = ray.pos; _transform.rot = ray.dir;
-    }
+        ENSURE_3DVEC_UNIT_LENGTH(_transform.rot);
```
Puttichai, line 1197: @kanbouchou How about we put this check in the constructor of `class ray` too?

Puttichai:
@kanbouchou I think this might be a good time to also update descriptions of `IkParameterizationType` so that it explicitly states if for a certain ik type, something to be unit length, etc.

include/openrave/openrave.h
kanbouchou: Thanks, I added a check similar to quaternion class, but not exactly ``ENSURE_3DVEC_UNIT_LENGTH`` because ``ray`` class is templated and ``ENSURE_3DVEC_UNIT_LENGTH`` could not be applied.

kanbouchou:
> @kanbouchou I think this might be a good time to also update descriptions of `IkParameterizationType` so that it explicitly states if for a certain ik type, something to be unit length, etc.

I added a comment as follows to ``IkParameterization`` class
```
    ///< for IKP_Direction3D, IKP_Ray4D, IKP_TranslationDirection5D rot is a unit length 3d direction vector
    ///< for IKP_Transform6D, rot is a unit length quaternion
    Transform _transform;
```

include/openrave/openrave.h
kanbouchou: @rdiankov please review again, I made checker function to a macro (added similar one for quaternion). locally tested that stack (asserted location) is actually the caller of the macro.

kanbouchou:
@rdiankov please check again when you have time.

rdiankov:
thanks~

PR #1079, opened 2022-04-18
update plugins/qtcoinrave/item.cpp

lazydroid:
This is a fix for the shadowed variables bugs that arise when QTCOIN is used during the compilation. Unfortunately, this did not came up earlier in our previous tests since there are not so many users of QTCOIN,


rdiankov:
thanks~

yoshikikanemoto:
@lazydroid are you sure that your MR is to master, not production?

lazydroid:
@yoshikikanemoto right, I shall make another PR for the production branch, sorry for missing that.

PR #1081, opened 2022-04-20
Fix scaling when deserializing readable interfaces

Puttichai:
### Summary

When serializing readable interfaces in `KinBodyInfo::SerializeJSON`, scaling is applied. However, scaling is missing when deserializing readable interfaces. Scaling is already applied correctly when deserializing other fields such as `transform`.

This pull request modifies `KinBodyInfo::_DeserializeReadableInterface` so that it can use an appropriate scaling factor when calling `DeserializeJSON`.


ziyan:
@Puttichai Since you changed header, you might need to bump minor version.

ziyan (approved):

PR #1089, opened 2022-04-27
Use pickle protocol 2

cielavenir:
because it is more efficient

### DIFF
--- a/python/databases/__init__.py
+++ b/python/databases/__init__.py
@@ -110,7 +110,7 @@ class DatabaseGenerator(metaclass.AutoReloader):
         except OSError:
             pass
         with open(filename, 'wb') as f:
-            pickle.dump((self.getversion(),params), f)
+            pickle.dump((self.getversion(),params), f, 2)
     def generate(self):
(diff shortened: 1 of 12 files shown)


rdiankov:
@cielavenir let's also update the versions of the databases so we force re-generation

cielavenir:
https://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support

cielavenir:
ok let me update

cielavenir:
well is https://github.com/rdiankov/openrave/pull/1089/commits/67343eea4d8489d6383627afaaeb9fb6e1ee2ec0 correct?

rdiankov:
@cielavenir what about linkstatistics?
```
    def getversion(self):
        return 6
```

cielavenir:
@rdiankov done

well should I revert 67343eea4 ?

rdiankov:
@cielavenir we also have convexdecomposition.py

rdiankov:
thanks~

PR #1092, opened 2022-05-02
Fix IPython startup, otherwise matplotlib GUI does not work well

cielavenir:
(loadscene.py / runbinpickingsim.py etc has to be updated as well)

### DIFF
--- a/python/openrave.py.in
+++ b/python/openrave.py.in
@@ -247,7 +247,7 @@ if __name__ == "__main__":
                             from IPython import embed
                             embed()
                         else: # later versions of ipython 1.1+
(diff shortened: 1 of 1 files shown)


Puttichai (approved):
Looks good to me.

cielavenir:
@rdiankov

- `embed.InteractiveShellEmbed.instance` exists even on IPython 1.x (cf https://github.com/ipython/ipython/blob/1.x/IPython/terminal/embed.py), so the change will be least
- alternatively you could do `embed(local_ns=locals())`

rdiankov:
sure... thanks

PR #1096, opened 2022-05-09
Rename to jointControlInfo(RobotController|IO|ExternalDevice), basica…

ziyan:
…lly "jointControlInfo" + controlMode.


felixvd:
This has my support

braineo:
Looks much better IMO

snozawa:
nice!

how about variable names?

if this PR is intended to change minimally, e.g. only change the naming exposed to the users, i think it's fine.
Thanks!

ziyan:
> nice!
> 
> how about variable names?
> 
> if this PR is intended to change minimally, e.g. only change the naming exposed to the users, i think it's fine. Thanks!

The variable names are fine as-is. They can change over time, and compilation will check for correctness. The JSON names, however, are contracts that will be much much harder to change.

snozawa:
> The variable names are fine as-is. They can change over time, and compilation will check for correctness. The JSON names, however, are contracts that will be much much harder to change.

i see, make sense!

PR #1105, opened 2022-05-17
Add control of whether or not to use _MergeConsecutiveSegments in parabolicsmoother2

Puttichai:
### Summary

With the changes in this pull request, users will be able to use an existing planning parameter `maxmergeiterations` to control `parabolicsmoother2` whether or not to use `_MergeConsecutiveSegments`, which takes place before the shortcutting process.


rdiankov:
thanks

PR #1106, opened 2022-05-18
plugins/basesamplers: Split header-only implementation into source file.

undisputed-seraphim:


rdiankov:
any advantages to this?

undisputed-seraphim:
I did it as part of my goal of making the compiler shut up about boost bind warnings. I split it into a CPP file so that I can limit the spilling of a `using namespace boost::placeholders` in a header file.

(Well, yeah, there's still `using namespace OpenRAVE;` but that will be a pull request for another time)

PR #1109, opened 2022-05-18
Cleanup cmake more

undisputed-seraphim:
- Remove `STDC_LIBRARY` because we should let CMake decide the std library to use for us instead of forcing a particular value. When cross compiling, force-overriding a stdlib leads to compile errors because the toolchain is supposed to handle this for us.

- Remove `EXTRA_COMPILE_FLAGS`. Previously this was used to set the `-fPIC` flag, but CMake has a canonical way of setting `-fPIC` for all projects and modules called `CMAKE_POSITION_INDEPENDENT_CODE`, which is automatically disabled for platforms that don't support the flag.

- Change `NATIVE_COMPILE_FLAGS` to call into a check for `OPT_NATIVE` (newly introduced here)
This change is OK because internally we use `NATIVE_COMPILE_FLAGS` on the jhbuild command line to set `-march=native -mtune=native`, which I am now going to use the correct way, setting `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`.

n.b. `-march=native` implies `-mtune=native`

Autotester ID: 254606


CMakeLists.txt
```diff
@@ -16,9 +16,10 @@ cmake_policy(SET CMP0002 NEW)
 # http://www.cmake.org/cmake/help/cmake-2.6.html#policy:CMP0003
 cmake_policy(SET CMP0003 NEW)
 
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14) # C++14 for std::shared_timed_mutex
-endif()
+set(CMAKE_CXX_STANDARD 14) # C++14 for std::shared_timed_mutex
```
cielavenir, line 19: log4cxx provided by apt on Ubuntu Jammy (22.04) cannot be compiled with C++14.

this change will break https://github.com/cielavenir/mujin_recruiting/commit/8a1613e4b5913eee0a4344ce0bbb6a514e8e353c#diff-3920b095bdcd12cd3fd5bd8934327a1371372fc1fc7f5d5213601cfbc1a38d00R208 .

cielavenir:
(Unlike dev build) Docker build does not use native, please double check with @ziyan for the correct direction

CMakeLists.txt
undisputed-seraphim: OK, but libopenrave-core is already using `std::shared_timed_mutex` so we have to enable C++14 whether we like it or not

undisputed-seraphim:
Not a problem, I can just add a check that disables native if the variable is not set or empty.

CMakeLists.txt
undisputed-seraphim: [Is it related to this?](https://logging.apache.org/log4cxx/latest_stable/dependencies.html)

`As of version 0.12.0, Log4cxx requires a minimum C++ version of C++11. If C++17 is not available, then Log4cxx requires Boost Thread in order to build, which in turn requires chrono and date_time.`

cielavenir: Yes I meant C++17/20/23/bla

cielavenir: https://github.com/apache/logging-log4cxx/blob/master/src/cmake/boost-fallback/boost-std-configuration.h.cmake

the config.h of this file specifies std::shared_mutex mode because gcc11 defaults to C++17. And C++14 cannot include it.

cielavenir: actually it might be lucky that gcc10 defaults to c++14. 17 has some incompatibilities even on openrave...

PR #1112, opened 2022-05-19
Undefined behaviour in RobotBase::Manipulator

undisputed-seraphim:
There are places where `RobotBase::Manipulator` is being constructed with a `new` (such as [robotconnectedbody.cpp#L802](https://github.com/rdiankov/openrave/blob/production/src/libopenrave/robotconnectedbody.cpp#L802).

However, it turns out that the constructors of `RobotBase::Manipulator` is private. Doing a `new` construction while the constructors are in private scope is undefined behaviour. I would like to change the constructor to public. (The private constructor also prevents usage of `make_shared`, an upcoming optimization.)


cielavenir (approved):

rdiankov:
need to update minor version since ABI changes

rdiankov:
thanks

PR #1113, opened 2022-05-19
Better C++ conformance with Clang

undisputed-seraphim:
Some of these warnings were issued when I tried to compile this against Clang, using its default set of warnings:

-Winconsistent-missing-override - Improperly labeled methods that override methods from base class

-Wno-non-virtual-dtor - Not defining a destructor to a derived class with virtual functions is actually undefined behaviour; gcc is lax in this but consulting with the standard shows that Clang is correct in issuing the warning.

-26db3539a: Self assignment

`string` was not prefixed with `std::` in a header file - not sure how this even compiled before, but Clang gives a compile error.

One unused variable was removed and one logging bug was discovered.

### DIFF
--- a/include/openrave/collisionchecker.h
+++ b/include/openrave/collisionchecker.h
@@ -75,6 +75,8 @@ public:
         Reset();
     }
 
+    virtual ~CollisionReport() noexcept {}
+
     /// \brief resets the report structure for the next collision call
     ///
     /// depending on nKeepPrevious will keep previous data.
--- a/include/openrave/openrave.h
+++ b/include/openrave/openrave.h
@@ -1003,7 +1003,7 @@ protected:
         \param[out] useddofindices a vector of unique DOF indices targetted for the body
         \param[out] usedconfigindices for every used index, returns the first configuration space index it came from
      */
-    void ExtractUsedIndices(const char* pBodyName, int nBodyNameLength, int timederivative, std::vector<int>& useddofindices, std::vector<int>& usedconfigindices) const;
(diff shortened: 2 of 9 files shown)


undisputed-seraphim:
Autotester ID: 254883

include/openrave/collisionchecker.h
```diff
@@ -73,6 +73,8 @@ class OPENRAVE_API CollisionReport
         Reset();
     }
 
+    virtual ~CollisionReport() noexcept {}
```
rdiankov, line 76: @undisputed-seraphim just curious, what is the purpose of adding this here?

undisputed-seraphim: A class that has methods marked `virtual` should have also have a `virtual` destructor. [See this.](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-virtual-class-destructor.html) This is so that so class instances can be destroyed properly when accessing them through pointers.

It's just a warning, but since this is in a header file, the warning is emitted once every time the header is encountered, which is many times, and adds a lot of noise to compilation.

undisputed-seraphim:
Latest pipeline ID: 258501 / 341825

include/openrave/openrave.h
```diff
@@ -1003,7 +1003,7 @@ class OPENRAVE_API ConfigurationSpecification
         \param[out] useddofindices a vector of unique DOF indices targetted for the body
         \param[out] usedconfigindices for every used index, returns the first configuration space index it came from
      */
-    void ExtractUsedIndices(const char* pBodyName, int nBodyNameLength, int timederivative, std::vector<int>& useddofindices, std::vector<int>& usedconfigindices) const;
```
rdiankov, line 1006: @undisputed-seraphim why make the change to `size_t`?

PR #1114, opened 2022-05-19
Restore environment name in JSON

ziyan:
`name` attribute in environment JSON is needed by the registration cell. Registration cell creates a scene file first using OpenRAVE, then upload that file to WebStack. `name` needs to survive in the file. We allow customers to set a custom name for work pieces using this attribute as well.

This is a partial revert of 083d804ba523cd163f2c0852feaa68d7548075a2


include/openrave/environment.h
```diff
@@ -808,14 +803,19 @@ class OPENRAVE_API EnvironmentBase : public boost::enable_shared_from_this<Envir
         /// \param vInputToBodyInfoMapping maps indices into rEnvInfo["bodies"] into indices of _vBodyInfos: rEnvInfo["bodies"][i] -> _vBodyInfos[vInputToBodyInfoMapping[i]]. This forces certain _vBodyInfos to get updated with specific input. Use -1 for no mapping
         void DeserializeJSONWithMapping(const rapidjson::Value& rEnvInfo, dReal fUnitScale, int options, const std::vector<int>& vInputToBodyInfoMapping);
 
+        /// \name Properties loaded from environment file and will be kept as-is
```
rdiankov, line 806: @ziyan  what does that mean "as-is"?

ziyan: Meaning `env.Load('{"name": "customer product name"}')` will not interfere with `env.GetName()` / `env.GetIdName()` which you already repurposed for debugging purpose. We want `env.WriteToMemory('json')` to give back `{"name": "customer product name"}'`

ziyan: Ping @rdiankov

rdiankov: err..... that's confusing. We need a better name than `name`.
Might as well call the parameter `description`

ziyan: @rdiankov Well, name was repurposed and broken by you and Kei. Description already exists, it's used for notes about the part on registration ui.

rdiankov: Nevertheless, `name` is not appropriate.
Some suggestions for this use case: `productName`, `shortDescription`, `title`

I think `productName` would be fitting

rdiankov: @ziyan you have no preference?

ziyan: @rdiankov This implies we have to change all existing data in the database, create db migration, change API contracts, change client code etc.

I'd much rather such work be minimized in the future. When you re-purpose existing fields, please consider the breakage.

For now, I am going to put this as low priority and work on it later. Or find alternative to set name from the UI. Maybe change `createpickupbox.py` to serialize to json then modify the name inside.

rdiankov: "name" was always a backhanded way of getting some other functionality complete in a quick way, and it never had the proper design discussions.
Changes are not that much and not so bad.
I can talk to you offline about this.

ziyan: @rdiankov Please involve @braineo for field name change

ziyan: And perhaps @felixvd for API doc changes

ziyan: @rdiankov How about change the `_name` to `_internalName` or `_immutableName` or `_debugName`?

braineo: I think we need major migration if we change the name attribute, since that field is already widely used in production for registration onsite. The field will be lost after `Load()` and `Save()`. Can we have a new field for debug purpose and restore the original `_name` in the info?

PR #1116, opened 2022-05-23
Static initialization in config specification

undisputed-seraphim:
Now that we target C++14, brace init of a set is possible and makes for more concise code.


rdiankov:
sure, thanks~

PR #1117, opened 2022-05-23
Fix retiming dofindices

Puttichai:
### Summary

This pull request removes the assumption in `parabolicretimer` and `parabolicretimer2` that the manip's arm DOF and trajectory DOF are equal.

The assumption is not necessary for the computation. It also does not hold if the trajectory also contains manip's gripper DOF.


rdiankov:
thanks~

PR #1118, opened 2022-05-25
Substitute mutexes, locks, and threads from boost to std

undisputed-seraphim:
Note that this pull request does NOT merge into production or master; it merges into a "master dev branch" I call openrave-static. I'm doing it like this so that I can make small merge requests that are easy to manage and comprehend, topic by topic, without affecting production branch immediately.

This merge request is for the conversion of mutexes, locks, condition_variables, and threads from `boost` namespace to `std`. This is done in preparation for supporting platforms that have different implementations of threads and mutexes that are not necessarily fully posix compatible, but provide a C++ standard library.

Note that `call_once` has not been converted yet. I'm still testing those, because the syntax change to `std` is not as straightforward as I thought. That will be a separate merge request.

### DIFF
--- a/include/openrave/environment.h
+++ b/include/openrave/environment.h
@@ -24,7 +24,7 @@
 
 namespace OpenRAVE {
 
-typedef boost::recursive_try_mutex EnvironmentMutex;
+typedef std::recursive_mutex EnvironmentMutex;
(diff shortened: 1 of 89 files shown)


undisputed-seraphim:
Closed to rework some changes

PR #1120, opened 2022-05-25
Change synchronization primitives from boost to std

undisputed-seraphim:
This merge request is for the conversion of mutexes, locks, condition_variables, and threads from `boost` namespace to `std`. This is done in preparation for supporting platforms that have different implementations of threads and mutexes that are not necessarily fully posix compatible, but provide a C++ standard library.

Summary of changes:
- synchronization primitives changed from boost to std.
- Added option to CMake `OPT_ENVIRONMENT_RECURSIVE_LOCK` to toggle between typedefs in file environment.h. Defaults to ON (recursive mutex enabled).
- Everywhere else in the code (besides the typedef in environment.h), `EnvironmentMutex::scoped_lock` is changed to `EnvironmentLock`.
- Headers added where necessary.

If C++17 is not enabled (I understand we're only supporting C++14 at the moment, so this is optional) but recursive locks are requested, then compilation will fall back to the boost implementation (in other words, current implementation)


rdiankov:
Is there any advantage of supporting boost synchronization primitives? If not, just convert the synchronization primitives all to std (but make sure to do it with typedefs to allow for easier changes later.

For now, please keep other boost usages like boost::shared_ptr as is

undisputed-seraphim:
> Is there any advantage of supporting boost synchronization primitives? If not, just convert the synchronization primitives all to std (but make sure to do it with typedefs to allow for easier changes later.
> 
> For now, please keep other boost usages like boost::shared_ptr as is

There really is no advantage. There may even be a disadvantage, e.g. stdlibs of various platforms may implement optimizations specific to that platform, whereas boost has to use generic implementations (otherwise it will have to include every possible platform and the code size will bloat).

I have kept other boost usage as-is.

undisputed-seraphim:
Put on hold for a bit, while I add the typedefs

undisputed-seraphim:
@rdiankov Kindly review

rdiankov:
thanks~

include/openrave/environment.h
```diff
@@ -24,8 +24,25 @@
 
 namespace OpenRAVE {
 
-typedef boost::recursive_try_mutex EnvironmentMutex;
-typedef EnvironmentMutex::scoped_lock EnvironmentLock;
+#if OPENRAVE_ENVIRONMENT_RECURSIVE_LOCK
+#if __cplusplus >= 201703L
```
cielavenir, line 28: Actually this `__cplusplus >= 201703L` is confusing. This path compiles in C++11.

I don't see why limited to C++17, maybe @undisputed-seraphim was confused with std::scoped_lock?

Actually std::scoped_lock cannot be used here.

Now I removed `__cplusplus >` from environment.h and the condition is unclear.

cc @felixvd @hemangandhi @yoshikikanemoto

PR #1121, opened 2022-05-26
remove pybind11 extract warning, is more spamming than expected

cielavenir:
autotester log are often flooded


cielavenir:
actually this should be draft

I even expect someone submits better idea

cielavenir:
@undisputed-seraphim so you said that you would utilize shared_ptr, but that shared_ptr cannot be casted to different type. I wonder how you were going to utilize it.

cielavenir:
so, `extract_<T>::check()` sometimes used for positive reasons (other than sanity check), where I think warning should not be printed.

PR #1122, opened 2022-05-31
Split plugin database into header and cpp file

undisputed-seraphim:
Part of the changes I'm making to implement a statically linked plugin database.

In this change, the original plugin database implementation is split into a compilation unit (as opposed to the previous header-only implementation) to facilitate the next set of changes. This is done so that each step of changes is easy to understand and review. Aside from a bit of rearrangement of code, there are no logic changes.

pipeline ID: 257006

### DIFF
--- a/src/libopenrave/libopenrave.cpp
+++ b/src/libopenrave/libopenrave.cpp
@@ -1272,22 +1272,22 @@ InterfaceBasePtr RaveCreateInterface(EnvironmentBasePtr penv, InterfaceType type
 
 RobotBasePtr RaveCreateRobot(EnvironmentBasePtr penv, const std::string& name)
 {
-    return RaveInterfaceCast<RobotBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_Robot, name));
+    return boost::static_pointer_cast<RobotBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_Robot, name));
 }
 
 PlannerBasePtr RaveCreatePlanner(EnvironmentBasePtr penv, const std::string& name)
 {
-    return RaveInterfaceCast<PlannerBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_Planner, name));
+    return boost::static_pointer_cast<PlannerBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_Planner, name));
 }
 
 SensorSystemBasePtr RaveCreateSensorSystem(EnvironmentBasePtr penv, const std::string& name)
 {
-    return RaveInterfaceCast<SensorSystemBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_SensorSystem, name));
+    return boost::static_pointer_cast<SensorSystemBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_SensorSystem, name));
 }
 
 ControllerBasePtr RaveCreateController(EnvironmentBasePtr penv, const std::string& name)
 {
-    return RaveInterfaceCast<ControllerBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_Controller, name));
+    return boost::static_pointer_cast<ControllerBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_Controller, name));
 }
 
 MultiControllerBasePtr RaveCreateMultiController(EnvironmentBasePtr env, const std::string& rawname)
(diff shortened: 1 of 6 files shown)


src/libopenrave/libopenrave.cpp
```diff
@@ -791,7 +791,7 @@ class RaveGlobal : private boost::noncopyable, public boost::enable_shared_from_
         if( !_pdefaultsampler ) {
             boost::mutex::scoped_lock lock(_mutexinternal);
             BOOST_ASSERT( _mapenvironments.size() > 0 );
-            _pdefaultsampler = GetDatabase()->CreateSpaceSampler(_mapenvironments.begin()->second->shared_from_this(),"MT19937");
+            _pdefaultsampler = RaveCreateSpaceSampler(_mapenvironments.begin()->second->shared_from_this(), "MT19937");
```
rdiankov, line 794: @undisputed-seraphim umm... why make this change?

rdiankov: @undisputed-seraphim ping

undisputed-seraphim: I removed the callee `CreateSpaceSampler`. Now I realize that `RaveCreateSpaceSampler` creates it under the global OpenRAVE instance which may not necessarily be correct.

I will do it like this:
```
InterfaceBasePtr ifBasePtr = GetDatabase()->Create(_mapenvironments.begin()->second->shared_from_this(), PT_SpaceSampler, "MT19937");
_pdefaultsampler = RaveInterfaceCast<SpaceSamplerBase>(ifBasePtr);
```

undisputed-seraphim:
@rdiankov Kindly review

src/libopenrave/libopenrave.cpp
```diff
@@ -1271,22 +1272,22 @@ InterfaceBasePtr RaveCreateInterface(EnvironmentBasePtr penv, InterfaceType type
 
 RobotBasePtr RaveCreateRobot(EnvironmentBasePtr penv, const std::string& name)
 {
-    return RaveGlobal::instance()->GetDatabase()->CreateRobot(penv,name);
+    return RaveInterfaceCast<RobotBase>(RaveCreateInterface(penv, PT_Robot, name));
```
rdiankov, line 1275: @undisputed-seraphim why not use `RaveGlobal::instance()->GetDatabase()->Create` directly?

undisputed-seraphim: Sure I can, but there's no difference between those two ways, the compiler will inline all of them during optimization.

I went with this way because it's fewer characters for the human programmer to parse, and the defintion of `RaveCreateInterface` is right above.

rdiankov: actually there is, please use `RaveGlobal::instance()->GetDatabase()-`

undisputed-seraphim:
Pipeline #345768

src/libopenrave/libopenrave.cpp
```diff
@@ -1310,62 +1311,62 @@ MultiControllerBasePtr RaveCreateMultiController(EnvironmentBasePtr env, const s
 
 ModuleBasePtr RaveCreateModule(EnvironmentBasePtr penv, const std::string& name)
 {
-    return RaveGlobal::instance()->GetDatabase()->CreateModule(penv, name);
+    return RaveInterfaceCast<ModuleBase>(RaveGlobal::instance()->GetDatabase()->Create(penv, PT_Module, name));
```
rdiankov, line 1314: @undisputed-seraphim I think GetDatabase() ensures that interfaces returned from plugins have GetInterfaceType() match the expected type (If not, that's a bug then). Perhaps you can put the asserts in the database, and just use `boost::static_pointer_cast` directly here.

undisputed-seraphim: @rdiankov  It turns out that the checks in `RaveInterfaceCast` [returns a null shared_ptr if the types don't match](https://github.com/rdiankov/openrave/blob/master/include/openrave/openrave.h#L2601), and it looks like some of the tests in autotester has come to rely on this behaviour (using an assert causes those tests to fail). I shall simulate that behaviour to maintain compatibility.

undisputed-seraphim: Upon closer examination, the reason for that is because it's not a fatal condition to look for an interface but it can't be found, and `Create` just returns a null (shared) pointer and `RaveInterfaceCast` just passes the null pointer along.

rdiankov:
woot, congrats~

PR #1123, opened 2022-06-03
Save environment intParameters in JSON

ziyan:
I chose the name `intParameters` to be more consistent with what's in joint. There is no "unsigned" integer in JSON.


rdiankov:
@ziyan rapidjson has uint64 support, and i think it is misleading if we save them as `intParameters`. I guess something we can do is convert to `std::map<std::string, int64_t>_nInt64Parameters` and just have the `SetUInt64Parameter` cast to that.

PR #1124, opened 2022-06-07
reimplement pybind11 extract_ so that try-catch is not required

cielavenir:
closes #1121


cielavenir:
```
robot.GetActiveManipulator().GetIkParameterization(openravepy.IkParameterizationType.Transform6D)
```

this gives warn if unpatched, does not give warn if patched

rdiankov:
thanks~

PR #625, opened 2018-12-16
Migrate from boost::numeric to boost:numpy for Boost 1.6x

morxa:
Adapt the code base to Boost 1.6x:
* Replace every `boost::numeric::array` by `boost::numpy::array`
* Use `boost::numpy::ndarray` instead of `PyArray`
* Adapt and rewrite parts of `PyTriMesh` constructor and its `GetTriMesh` function
* Remove no longer needed `AutoPyArrayObjectDereferencer`

This fixes #551.

PR #1125, opened 2022-06-08
WIP: Gain compatibility with the stock Pybind11 (2.9.2)

cielavenir:
Historically, we were discussing boost-1.65 compatibility in https://github.com/rdiankov/openrave/pull/625 but we forgot it after pybind11 version. So openrave cannot be built on boost::python on latest Debian even if we are able to get py2 binary.

The thing is that `py::object operator[]` does not accept int not size_t in the stock pybind11. So we have to use patched pybind11, which is not user-friendly.
I personally wrote the build guide to make pybind11 version somewhat user-friendly, but it caused some internal problems.

Now I tried to build openrave using the stock pybind11. It is like that casting all integer indexing by py::to_object.

I did test the compilation on both bullseye (pybind11) and stretch (boost::python), so it works at least.

However, it surely costs the readability a lot. Please reconsider about the guide availability. Removing the guide is not good idea at all.


cielavenir:
tried to add this but operator[] cannot be global...

```
pybind11::detail::item_accessor operator[](const pybind11::object &o, int key) {
    return o[pybind11::to_object(key)];
}

pybind11::detail::item_accessor operator[](const pybind11::object &o, size_t key) {
    return o[pybind11::to_object(key)];
}
```

cielavenir:
There seems like a pybind11 issue open for years https://github.com/pybind/pybind11/issues/1309

cielavenir:
sorry guys I'm tired, about discussion about this matter, please don't come to my desk directly and please check with @lazydroid first.

cielavenir:
let @woodychow and @tgn3000 know this matter

cielavenir:
just to mention that https://github.com/pybind/pybind11/pull/4004 will deprecate this pull request (clarify: I want to close this #1125 without merging)

rdiankov:
thanks, for now, let's gain compatibility with stock.

PR #1126, opened 2022-06-13
add __hash__ for all classes comparing pointer. fixed the hash type to long.

cielavenir:
some tasks uses env as dict key, so env should implement `__hash__`. also, openravepy_robotbase.h uses long as `__hash__` return type and others should use long as well.

----

pipelineid=259984


rdiankov:
sure, thanks~

PR #1127, opened 2022-06-13
Introduce CollisionReportInfo

kanbouchou:
Background
==============

- This PR provides a new class ``CollisionReportInfo`` which contains equivalent information to ``CollisionReport`` in a way that is independent from OpenRAVE environement.
- Intended use is for serializing & deserializing collision report information without requiring openrave environment.


rdiankov:
thanks~

PR #1129, opened 2022-06-14
Add macro to check for required json keys

DevwratJoshi:
1. ```LoadJsonValueByKey``` now returns a bool. If the key does not exist in the json document, the function will return false, else will return true.     
2. A macro (```ORJSON_LOAD_REQUIRED_JSON_VALUE_BY_KEY```) is added, wrapping ```LoadJsonValueByKey```.  If the keys are not in the document, an exception is thrown.


rdiankov:
thanks~

PR #1133, opened 2022-06-22
Propagate geometry name when the robot.xml is loaded

snozawa:
- geometry name in `robot.xml` is not propagated and resultant geometry name is always empty. this PR fixes it.
- not fully sure whether it needs to set the names to all of `listGeometryInfos`, or it's ok to set `listGeometryInfo.front()`. for now, i set the name to `front`.


rdiankov:
sure, thanks~

PR #1135, opened 2022-06-23
bug fix computing self collision among grabbed items with same grabbing link

kanbouchou:
Background
===========

- This MR fixes a false positive self collision between two grabbed items grabbed by the same link.
- We should not check collision between two bodies grabbed by same robot link (and the rigidly attached links).

Detail
======
 - ``vAttachedToOtherGrabbed`` was empty, so ``bSameLink && std::find(vAttachedToOtherGrabbed.begin(), vAttachedToOtherGrabbed.end(), *itOtherGrabbedLink) != vAttachedToOtherGrabbed.end()`` always evaluates to false.
- As a result, false positive detection of two grabbed items who share same grabbing robot link could happen.


rdiankov:
thanks~

PR #1136, opened 2022-06-28
add IKRA_RejectDistanceCheck

cielavenir:
In teachworker ikfastcpp, this 0x401 is hardcoded with todo comment.

This is used for angle threshold check.

### DIFF
--- a/include/openrave/iksolver.h
+++ b/include/openrave/iksolver.h
@@ -53,6 +53,7 @@ enum IkReturnAction
     IKRA_RejectEnvCollision = (IKRA_Reject|0x00000040),
     IKRA_RejectJointLimits = (IKRA_Reject|0x00000100),
     IKRA_RejectKinematicsPrecision = (IKRA_Reject|0x00000200),
+    IKRA_RejectDistanceCheck = (IKRA_Reject|0x00000400),
     IKRA_RejectCustomFilter = (IKRA_Reject|0x00008000), // the reason should be set in the upper 16 bits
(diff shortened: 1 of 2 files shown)


rdiankov:
@cielavenir if not used in the openrave core, why not add it to all the other custom IKRA enums in `controllercommon/src/include/mujinopenravecommon/mujinopenravecommon.h`?

cielavenir:
Thank you. I checked the file. Unfortunately all upper 16bits were occupied.

@kanbouchou Now I need your opinion and you would need to fix webstack2_itlcache2_buildme/ikfastcpp/src/orp_ikfastcpp/ikfastcppsolver_solve.cpp#L300 TODO comment. /cc @lazydroid

kanbouchou:
> Thank you. I checked the file. Unfortunately all upper 16bits were occupied.
> 
> @kanbouchou Now I need your opinion and you would need to fix webstack2_itlcache2_buildme/ikfastcpp/src/orp_ikfastcpp/ikfastcppsolver_solve.cpp#L300 TODO comment. /cc @lazydroid

@rdiankov What about making ``enum IkReturnAction : uint64_t`` so that we have more upper bits available for ``IKRA_RejectDistanceCheck``?

rdiankov:
@kanbouchou yes, i think we should be moving to 64bit since we have too many return codes.

cielavenir:
https://github.com/rdiankov/openrave/pull/1159

PR #1138, opened 2022-07-01
Fix SamplePointsSameDeltaTime does not respect ensureLastPoint

AntoineRyu:
While using `SamplePointsSameDeltaTime` with `ensureLastPoint == true`, it is possible to miss the last point if it is within `g_fEpsilon` of a multiple of `deltatime`. Most interpolation methods will be negligibly close, since the error is < `g_fEpsilon`, however, we can completely miss it if the interpolation method is "previous". Therefore, always copy the trajectory's last point into the returned data's last point for groups with interpolation "previous" if `ensureLastPoint == true`.


rdiankov:
@AntoineRyu in general, why not just copy the whole last point rather than checking for prevoius?

rdiankov:
@AntoineRyu I just pushed my recommended change, please review it. thanks

AntoineRyu:
Looks good. I had already created a new test for it and it passes with your proposed changes.

rdiankov:
thanks

PR #1141, opened 2022-07-12
link pthread explicitly

cielavenir:
currently dlopen is resolved from libxml2 and pthread is resolved from libboost_thread, which could break later.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,6 +27,9 @@ endif()
 if( CLOCK_GETTIME_FOUND )
   set(openrave_libraries ${openrave_libraries} rt)
 endif()
+if( CMAKE_THREAD_LIBS_INIT )
+  set(openrave_libraries ${openrave_libraries} ${CMAKE_THREAD_LIBS_INIT})
+endif()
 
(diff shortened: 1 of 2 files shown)


cielavenir:
pipelineid=265994

undisputed-seraphim:
I think the CMake values `CMAKE_THREAD_LIBS_INIT` are not set unless you do `find_package(Threads)` first.

Additionally, l think the value used in `target_link_libraries` should be `Threads::Threads` so that it is portable (e.g. the system I am porting OpenRAVE to does not have `pthreads`, it provides a different thread library)

cielavenir:
- right find_package(Threads) is required, CMake implicitly loads Threads when loads MKL, but we should not rely on that behavior.
    - This one is pushed

----

- for Threads::Threads, it is CMake 3.1+, so not sure if it is ok
    - maybe ok as you (@undisputed-seraphim) introduced OpenGL::OpenGL and CMake 3.8 hard-requirement...?
    - by the way CMAKE_THREAD_LIBS_INIT is empty string on non-posix environments.

rdiankov:
sure guys, thanks

PR #1142, opened 2022-07-12
optimize PyTrajectoryBase

kanbouchou:
Brief
=====
- This MR aims to speed up PyTrajectoryBase
- For common input data size (1000 waypoints, 13 dof), ``Insert`` in new implementation is 4 to 10 times faster.

Optimizaions
===============

- When input is numpy array of float data type, use ``memcpy`` to copy the memory value as chunk. Fallback to ``ExtractArray`` if input is of other types.
- Have two std vector caches, ``_vtimesCache`` and ``_vdataCache`` to minimize memory allocation.


rdiankov:
@kanbouchou While we are on this topic, let's add a `Insert(size_t index, const dReal* pdata, int nDataElements)` implementation so we can avoid the unnecessary copying.

kanbouchou:
> @kanbouchou While we are on this topic, let's add a `Insert(size_t index, const dReal* pdata, int nDataElements)` implementation so we can avoid the unnecessary copying.

I implemented this. In addition
- optimized ``ConfigurationSpecification::GetDOF()`` a bit. Further improvement would require caching of dof value, but didn't do it for now because ``_vgroups`` is public and can be modified from outside easily.
- bumped minor version for new api's.

Please review again when you have time.

PR #1143, opened 2022-07-19
prohibit missing return

kanbouchou:
Overview
===============

- This MR adds ``-Werror=return-type`` to compiler option to fail to build functions without proper return.

Background
===============

- I ran into an mysterious segfault with function missing proper return.
- I think there was a behavior change in g++ around how it treats following code, somewhere between version 7.5.0 and 8.4.0.

```c++
#include <iostream>

bool SomeFunc(int& i)
{
    std::cout << i << std::endl;
    if (i > 0) {
        i--;
        SomeFunc(i);
    }
    //return true;
}

int main()
{
    int i = 1;
    SomeFunc(i);
    return 0;
}
```
without return, 
  - g++ 8.4 without optimization -> fine
  - g++ 8.4 with ``O2`` -> segfault
  - g++ 7.5 with ``O2`` -> fine

### DIFF
--- a/3rdparty/convexdecomposition/NvHashMap.h
+++ b/3rdparty/convexdecomposition/NvHashMap.h
@@ -1351,7 +1351,7 @@ namespace CONVEX_DECOMPOSITION
 	to choose from.  I only looked at a billion or so.
 	--------------------------------------------------------------------
 	*/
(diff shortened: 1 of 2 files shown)


cielavenir:
It is also related to controllercommon fe3f14adf05d6c08242207351767afa8b8644099

cielavenir (approved):

rdiankov:
sure, thanks~

PR #1145, opened 2022-08-05
fix basestring

cielavenir:
autotester pipelineid=272814

### DIFF
--- a/python/misc.py
+++ b/python/misc.py
@@ -19,6 +19,7 @@ import os.path
 from sys import platform as sysplatformname
 from sys import stdout
 import numpy
+import six
 try:
     from itertools import izip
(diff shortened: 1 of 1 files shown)


rdiankov:
@cielavenir this is py2 compatible right?

cielavenir:
yes it is compatible

- it is tested under autotester
- it is tuple of basestring on py2, so the same meaning under isinstance

```
mujin@desktop:~ $ python2 -c 'import six;print(six.string_types)'
(<type 'basestring'>,)
mujin@desktop:~ $ python3 -c 'import six;print(six.string_types)'
(<class 'str'>,)
```

rdiankov:
thanks

PR #1149, opened 2022-08-17
Prefix joint names in gripper info with connected body name

strmio:
Link names and tool names in a Gripper Info in a Connected Body are prefixed with the connected body name when attached to a robot. Likewise, joint names also should be prefixed with a connected body name.

In the current implementation, only `gripperJointNames` are properly prefixed with a connected body name after deserialised into `GripperInfo::gripperJointNames`. This patch adds a logic to prefix all the parameters containing joint names in `GripperInfo::_docGripperInfo` with a connected body name.


rdiankov:
thanks~

Issue #123, opened 2012-09-15
divide global openrave functions and enviornment


rdiankov:
register all environments somewhere that all nodes can access? (ie, necessary for synchronization, RedirectController?)

PR #1150, opened 2022-08-27
Check parentFilename before constructing an invalid uri out of it.

ziyan:
Prevent the creation of uri such as `file:#123` that breaks many systems.

`_filename` is not always set depending on how the environment is loaded. If the scene didn't come from a file probably the correct thing to do is preserve its original uri.


rdiankov:
thanks

PR #1153, opened 2022-10-06
add gemoetric function of OrientedBoxFromAABB

YutaKojio:
In some cases, `OrientedBox` is more useful than `OBB` since it is easy to transfom.

### DIFF
--- a/include/openrave/geometry.h
+++ b/include/openrave/geometry.h
@@ -1913,10 +1913,10 @@ inline obb<T> OBBFromAABB(const RaveAxisAlignedBox<T>& ab, const RaveTransform<T
 /// \ingroup geometric_primitives
 /// \param[in] t transformation used to set the coordinate system of ab.
 template <typename T>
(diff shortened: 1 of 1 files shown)


include/openrave/geometry.h
```diff
@@ -1908,6 +1908,30 @@ inline obb<T> OBBFromAABB(const RaveAxisAlignedBox<T>& ab, const RaveTransform<T
...
+}
+
+/// \brief Transform an axis aligned bounding box to an oriented bounding box expressed in transform.
+///
+/// \ingroup geometric_primitives
+/// \param[in] t transformation used to set the coordinate system of ab.
+template <typename T>
+inline RaveOrientedBox<T> OrientedBoxFromAABB(const RaveAxisAlignedBox<T>& ab, const RaveTransform<T>& t)
+{
+    return OrientedBoxFromAABB(ab,RaveTransformMatrix<T>(t));
```
rdiankov, line 1932: @YutaKojio Basically this will transform to matrix and back to quat again, that's pretty silly...

YutaKojio: oh, right.. I fixed it not to convert wastefully.

rdiankov:
thanks

PR #1156, opened 2022-10-13
Fix self collision check edge case

kanbouchou:
Overview
=========

Under following scenario, self collision check between grabbed part and robot does not work properly.

0. Initially, robot has no self collision checker
1. robot grabs a body
2. set self collision checker to robot
3. self collision check does not work properly for grabbed part

Problem
=========
Following code in ``KinBody::Grab`` is responsible for initializing self collision checker's info for grabbed body.
```
    if( !!_selfcollisionchecker && _selfcollisionchecker != GetEnv()->GetCollisionChecker() ) {
        // collision checking will not be automatically updated with environment calls, so need to do this manually
        _selfcollisionchecker->InitKinBody(pGrabbedBody);
    }
```
However, when self collision checker is set after grab, ``InitKinBody`` for the grabbed body does not happen and self collision check does not work properly.

### DIFF
--- a/plugins/fclrave/fclspace.h
+++ b/plugins/fclrave/fclspace.h
@@ -619,7 +619,7 @@ public:
         if( !!pinfo ) {
             return GetLinkBV(*pinfo, index);
         } else {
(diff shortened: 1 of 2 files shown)


rdiankov:
thanks~

PR #1162, opened 2022-11-01
Fix compile errors

undisputed-seraphim:
92bd5bd62bfc6523adc86d82246b7b85a3f060c2: Missing dummy function leads to compilation failure when Curl is enabled but MsgPack is disabled.

36d32763eec10ac2715f32845cf537d49e85c87f: Incorrectly spelled variable name, missing include (Can't find OPENRAVE_EXCEPTION_FORMAT otherwise).

763c0243b3b71e565b9ed6d14aadeea95ca5bce9: Under certain flag combinations, the boost filesystem library is not linked properly.

ccea434d5b91b0aad58baeca29dcdcc07e6c7b00: The pointer-based API in msgpack [is](https://github.com/mujin/msgpack-c/blob/master/include/msgpack/v1/unpack.hpp#L1569) [obsolete](https://github.com/mujin/msgpack-c/blob/master/include/msgpack/v1/object.hpp#L1084) and is removed in a future version.

### DIFF
--- a/src/libopenrave-core/jsonparser/jsonreader.cpp
+++ b/src/libopenrave-core/jsonparser/jsonreader.cpp
@@ -23,6 +23,7 @@
 #include <openrave/openravejson.h>
 #include <openrave/openravemsgpack.h>
 #include <openrave/openrave.h>
+#include <openrave/openraveexception.h>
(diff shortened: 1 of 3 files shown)


rdiankov:
thanks

PR #1165, opened 2022-11-09
fix memory leak by not decrementing ref count

kanbouchou:
Background
==============
- Memory leak was introduced https://github.com/rdiankov/openrave/pull/1142
- Problem was that reference count was not getting decremented in the destructor of ``AutoPyArrayObjectDereferencer``.
-  ``AutoPyArrayObjectDereferencer psaver(openravepy::AutoPyArrayObjectDereferencer(arrPtr));`` is actually a function declaration according to c++ spec so destructor was not being called.
    - https://en.wikipedia.org/wiki/Most_vexing_parse#Example_with_classes


rdiankov:
oops, nice catch. let's be careful in the future...

PR #1166, opened 2022-11-09
check the existance of grabbing link in current robot

eisoku9618:
improve CBAS support on state saver

### situation
- cloneenv
   - robot grabs an item
   - the grab link index is greater than the number of links of origrobot
- origenv
   - robot does not grab anything
   - different CBAS from cloneenv

```python
cbas_A = [0, 1]
cbas_B = [1, 0]

env.Remove(robot)
robot.SetConnectedBodyActiveStates(cbas_A)
env.Add(robot)
robot.Grab(workpiece, robot.GetLink('CBAS_A_link')) # this link index is greater than the number of robot links in cbas_B

with robot.CreateRobotStateSaver(KinBody.SaveParameters.GrabbedBodies):
    env.Remove(robot)
    robot.SetConnectedBodyActiveStates(cbas_B)
    env.Add(robot)
```

We get `vector::_M_range_check: __n >= this->size()` exception from `at` access in `GetLinkVelocity` of genericphysicsengine.cpp because stored `Grabbed` information contains information from `cbas_A`, which might not exist in `cbas_B`.

- `KinBody::KinBodyStateSaver::Restore(boost::shared_ptr<KinBody> body)` -> `KinBody::KinBodyStateSaver::_RestoreKinBody(boost::shared_ptr<KinBody> pbody)` -> `KinBody::_UpdateGrabbedBodies()` -> `KinBody::Link::GetVelocity(Vector& linearvel, Vector& angularvel)` -> `GetLinkVelocity(KinBody::LinkConstPtr plink, Vector& linearvel, Vector& angularvel)`

I think the expected behavior is not restoring grabbed body which is grabbed by previous CBAS.


src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -92,6 +92,10 @@ void KinBody::KinBodyStateSaver::_RestoreKinBody(boost::shared_ptr<KinBody> pbod
         OPENRAVE_ASSERT_OP(pbody->_vGrabbedBodies.size(),==,0);
         for (const GrabbedPtr& pGrabbed : _vGrabbedBodies) {
             KinBodyPtr pGrabbedBody = pGrabbed->_pGrabbedBody.lock();
+            if( !pbody->GetLink(pGrabbed->_pGrabbingLink->GetName()) ) {
```
kanbouchou, line 95: We should do this inside ``if( !!pGrabbedBody ) {``, because ``pGrabbedBody`` could be a null pointer.

Also please check if ``pGrabbed->_pGrabbingLink`` is not null.

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -92,6 +92,10 @@ void KinBody::KinBodyStateSaver::_RestoreKinBody(boost::shared_ptr<KinBody> pbod
...
+                RAVELOG_WARN_FORMAT("env=%s, could not restore grabbing link '%s' for grabbed body '%s' since %s does not have the link (body num links is %d)", pbody->GetEnv()->GetNameId()%pGrabbed->_pGrabbingLink->GetName()%pGrabbedBody->GetName()%pbody->GetName()%pbody->GetLinks().size());
```
kanbouchou, line 96: Minor, but we are not really trying to restore grabbing link here, so the message should be changed a bit.

kanbouchou (changes requested):

rdiankov:
To tell you the truth, I think the correct behavior is to raise an exception because essentially the user is losing data: the grabbed object, which is dangerous. I don't see why we would want to encourage this.

src/libopenrave/kinbodystatesaver.cpp
eisoku9618: thank you

src/libopenrave/kinbodystatesaver.cpp
eisoku9618: updated the message in throw.

eisoku9618:
@kanbouchou @rdiankov I updated this PR to raise an exception and change the caller side to release if necessary before using kinbodystatesaver and tested.

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -93,6 +93,12 @@ void KinBody::KinBodyStateSaver::_RestoreKinBody(boost::shared_ptr<KinBody> pbod
         for (const GrabbedPtr& pGrabbed : _vGrabbedBodies) {
             KinBodyPtr pGrabbedBody = pGrabbed->_pGrabbedBody.lock();
             if( !!pGrabbedBody ) {
+                KinBody::LinkPtr pGrabbingLink(pGrabbed->_pGrabbingLink);
+                if( !!pGrabbingLink && !pbody->GetLink(pGrabbingLink->GetName()) ) {
```
kanbouchou, line 97: Shouldn't we throw exception if either ``pGrabbingLink`` or ``pbody->GetLink(pGrabbingLink->GetName())`` is nullptr?

rdiankov: i would do `!pGrabbingLink || !pbody->GetLink(pGrabbingLink->GetName()`

eisoku9618: @kanbouchou @rdiankov thank you, I updated the code to check `!pGrabbingLink` and remove old exception checks for warnings.

rdiankov:
thanks

PR #1167, opened 2022-11-10
Draft: Make chucking direction integer

kanbouchou:


kanbouchou:
same as https://github.com/rdiankov/openrave/pull/1168

PR #1168, opened 2022-11-11
make chucking direction int typed

kanbouchou:
Background
=============
- chucking direction is currently stored as double, but I believe the possible values are ``-1``, ``0``, and ``1``.
- In our application, there are many places where we try to parse chucking direction from json data, and result in exception. Fixing  the type on  application side is tedious, so I propose to change the internal representation int.


rdiankov:
thanks~

PR #1169, opened 2022-11-14
Fix self collision checking for grabbed item in cloned env

kanbouchou:
Problem Description
==================

0. In ``env1``, robot is grabbing a body
1. When doing ``env2 = env1.CloneSelf()``, robot in ``env2`` does call ``_selfcollisionchecker->InitKinBody`` in ``KinBody::Clone``, but grabbed body is not initialized.
2. in ``env2``, self collision checker does not consider the grabbed body.

Proposed Fix
============

- Initialize fclspace in ``KinBodyStateSaver`` who is in charge of making sure cloned robot ends up grabbing the body. I am proposing this fix because ``KinBodyStateSaver::Restore`` essentially has the same functionality as ``KinBody::Grab`` who initializes fcl space for grabbed item. Downside of this is that ``KinBodyStateSaver::Restore`` becomes slower.

Alternative Fix 1
================

- In ``Environment::_Clone``, after cloning of bodies and restoring of grabbed bodies are finished, perform ``_selfcollisionchecker->InitKinBody`` on grabbed bodies, but it felt messy.

Alternative Fix 2
================

- In the scope of ``KinBody::Clone`` called from ``Environment::_Clone``, robot is not grabbing, so we cannot ``InitKinBody`` on grabbed item like https://github.com/rdiankov/openrave/pull/1156.
```
                        // Should ignore grabbed bodies since the grabbed states will be updated later (via state saver's Restore) below.
                        pnewbody->Clone(pbody, options|Clone_IgnoreGrabbedBodies);
```
We could change the way ``Environment::_Clone`` works so that ``KinBody::Clone`` is called with body being grabbed, but that is a lot of change which may break other cases.


rdiankov:
thanks~

PR #1170, opened 2022-11-14
Add "files" to KinBody to retain files after opening and saving a scene

woswos:
Currently, "files" only exist inside `KinBodyInfo` class and thus opening and saving a scene doesn't retain the "files" section. This change fixes the issue.


rdiankov:
thanks~

PR #1171, opened 2022-11-16
fix ode collision checker and qtcoin renderer didn't consider CalibrationBoard geom type

yoshikikanemoto:
fyi @Puttichai


yoshikikanemoto:
Mark as draft since there seems to be another issue related to GT_CalibrationBoard. Geom for collision checking might be off from visualization.

yoshikikanemoto:
Also fixed discrepancy between qtosg visualization and fcl collision checking and added qtcoin rendering support

rdiankov:
thanks~

PR #1173, opened 2022-11-19
Update attached sensor geometry recursively

justism:
Recursively update attached sensor geometry to handle inner documents such as `instrinsics`


rdiankov:
thanks~

PR #1176, opened 2022-12-06
Remove normals from assimp

cielavenir:
https://www.dropbox.com/s/7f0g474xf9nb5wm/teapot.stl

without my patch:

```
openrave.py --viewer= -p 'mesh=env.GetBodies()[0].GetLinks()[0].GetGeometries()[0].GetCollisionMesh(); print("triangles=%d, vertices=%d"%(len(mesh.indices),len(mesh.vertices))); exit()' ~/Downloads/teapot.stl 
triangles=894, vertices=2310
```

with my patch:

```
openrave.py --viewer= -p 'mesh=env.GetBodies()[0].GetLinks()[0].GetGeometries()[0].GetCollisionMesh(); print("triangles=%d, vertices=%d"%(len(mesh.indices),len(mesh.vertices))); exit()' ~/Downloads/teapot.stl
triangles=894, vertices=480
```

this causes two issues:

1. larger msgpack size
2. potential performance issue

The core issue is that aiProcess_JoinIdenticalVertices does not work well if the stl has normals because JoinIdenticalVertices considers the normals value as vertex normals while assimp STLLoader loads triangle normals. According to https://github.com/assimp/assimp/issues/407 , we need to remove normals component during import.


cielavenir:
tested on:

- assimp 3.0.1 / openrave Python2
- assimp 5.2.5 / openrave Python3

PR #1177, opened 2022-12-07
Prohibit return local reference

cielavenir:
this week we had a trouble that one returned a reference to local std::string and it caused segfault.

for detail, please see test-townsquare!1754

autotester pipelineid=324937

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,12 @@ if( CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX OR COMPILER_IS_CLANG)
   add_definitions("-Werror=return-type")
 endif()
 
+if( CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX )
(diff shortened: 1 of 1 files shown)


cielavenir:
related to https://github.com/rdiankov/openrave/pull/1143

lazydroid (approved):
looks good, wish it was added earlier

rdiankov:
besides adding the error flag, is there a fix somewhere for something that was returning a local variable inside openrave?

undisputed-seraphim:
Did we give up supporting Clang? :laughing:

undisputed-seraphim:
Yeah I feel that this should be accompanied by a relevant fix to a piece of code that _is_ returning a local reference somewhere.

Otherwise we are just adding another warning to the pile of warnings that we are already ignoring.

cielavenir:
@undisputed-seraphim @rdiankov It was found in visioncommon and I don't see other parts returning local reference, but we cannot guarantee that the same failure would happen in openrave as well.

rdiankov:
sure, thanks

PR #1178, opened 2022-12-14
Fix lost non-trimesh geometry by zero padding

kanbouchou:
Problem
=========

- If a geometry is not trimesh type and zero clearance is specified, ``ConvexDecompositionModel`` generates no geometry for it. This is dangerous because those geometries are not collision-checked.

Proposal
==========

- Make sure ``self.linkgeometry`` contains the same number of elements as the total number of geometries of the robot.
- Bumped ``getversion`` return value to 4 to invalidate previously generated cache which may have above described problem.


rdiankov:
thanks~

PR #1181, opened 2022-12-20
Let the toolchain choose the name of standard C++ library to link against

undisputed-seraphim:
When cross-compiling, affixing the name of the standard library to link against causes problems such as the linker linking against the wrong library.

Cross compilers have a toolchain file that helps it determine the paths and names of libraries. Let those toolchain files do their job instead of meddling in it.

Autotester pipeline #460831


rdiankov:
sure, let's try this, thanks

PR #1182, opened 2022-12-20
Use CMAKE_SYSTEM_PROCESSOR

undisputed-seraphim:
cmake: Use CMAKE_SYSTEM_PROCESSOR as a cross-compilation-compatible way to detect target architecture. Current cmake code breaks when trying to do cross-compilation.

(Not even sure if it is necessary, I grepped the code and no file uses the `__x86_64__` define, and one commented-out line uses the `IS_X86_64` cmake variable. I secondarily propose that we remove it completely)

UPDATE:
Check for x86 has been removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -530,10 +530,6 @@ if( ASSIMP_FOUND )
   endif()
 endif()
 
-if (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64" OR ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "AMD64" )
-  add_definitions("-D__x86_64__")
-endif()
-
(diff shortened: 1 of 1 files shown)


cielavenir:
Actually even it is not required

In my hobby project there is a variable like this. This works without config, which means they are defined by compiler itself.

```
        const char *arch=
#if defined(__x86_64__)
        #if defined(__ILP32__)
                "x32"
        #else
                "amd64"
        #endif
#elif defined(__i386__)
        "i386"
#elif defined(__aarch64__)
        "arm64"
#elif defined(__arm__)
        "arm"
#elif defined(__sparc__)
        "sparc"
#else
        "other"
#endif
        ;
```

undisputed-seraphim:
Yeah I think I will add a commit to remove it completely instead.

rdiankov:
thanks

PR #1183, opened 2022-12-27
Add PyCollisionReport::Reset

cielavenir:
In airgripper simulator, we are now checking collision in a specific loop. There we want to reuse the same CollisionReport instance, so Resetting is better.


lazydroid (approved):
thanks! this will simplify and make things a bit faster, LGTM!

rdiankov:
thanks~

PR #1187, opened 2023-01-10
adding robotControllerAxis variables to JointControlInfo_RobotController

MiraiHattori:
Some intParameters and floatParameters in robotmotionparameters would be better in each jointInfo, so in preparation to the change I added  `robotControllerAxis{Mult|Offset|ProductCode}` in `JointControlInfo_RobotController`.


MiraiHattori:
@kanbouchou could you review this PR?

kanbouchou:
@MiraiHattori please make sure to bump version.

python/bindings/openravepy_kinbody.cpp
```diff
@@ -654,25 +654,60 @@ PyJointControlInfo_RobotController::PyJointControlInfo_RobotController(const Joi
 {
     controllerType = jci.controllerType;
 
-    py::list _robotControllerAxisIndex;
+    py::list _robotControllerAxisProductCode, _robotControllerAxisIndex, _robotControllerAxisMult, _robotControllerAxisOffset;
```
kanbouchou, line 657: I think you are following the naming pattern of existing code, however we should not start local variable with ``_``.  please name to something else.

src/libopenrave-core/colladaparser/colladareader.cpp
```diff
@@ -1750,13 +1750,34 @@ class ColladaReader : public daeErrorHandler
                                         if( pchild->getElementName() == std::string("controllerType") ) {
                                             jci.controllerType = pchild->getCharData();
                                         }
+                                        else if( pchild->getElementName() == std::string("robotControllerAxisProductCode") ) {
+                                            int ijointaxis = boost::lexical_cast<int>(pchild->getAttribute("axis"));
```
kanbouchou, line 1754: this is a repeated pattern, let's cleanup.

python/bindings/openravepy_kinbody.cpp
MiraiHattori: renamed them into `axis{Index|Mult|Offset|ProductCode}`.

src/libopenrave-core/colladaparser/colladareader.cpp
MiraiHattori: change the code to have ijointaxis parsed everytime getElementName() result is not `controllerType`.

MiraiHattori:
@kanbouchou 
sorry for not bumping version. just bumped.
I changed the `robotControllerAxis*` variable order in the code since I've noticed it was not constant in the code.

kanbouchou:
@MiraiHattori thanks, looks good to me.

@rdiankov can you please check?

include/openrave/kinbody.h
```diff
@@ -142,6 +142,9 @@ class OPENRAVE_API JointControlInfo_RobotController : public InfoBase
     std::string controllerType; ///< the type of the controller used to control this joint
     //int robotId = -1; ///< index of the robot within that controller?
     boost::array<int16_t, 3> robotControllerAxisIndex = {-1, -1, -1}; ///< indicates which DOF in the robot controller controls which joint axis (up to the DOF of the joint). -1 if not specified/not valid.
+    boost::array<dReal, 3> robotControllerAxisMult = {1.0, 1.0, 1.0}; ///< indicates the multiplier to convert the joint values from openrave to the robot controller, per joint axis.
+    boost::array<dReal, 3> robotControllerAxisOffset = {0.0, 0.0, 0.0}; ///< indicates the offset, in the robot controller units, that should be applied to the joint values, per joint axis.
+    boost::array<std::string, 3> robotControllerAxisProductCode = {"", "", ""}; ///< indicates the product code of the robot controller used to control each joint axis.
```
rdiankov, line 147: @MiraiHattori @kanbouchou can you explain why you need a "productCode" per axis? What are some examples of product codes? thanks

MiraiHattori: @rdiankov product code is an array of strings to identify which servo device the robot controller has to connect to, like a serial number of each device.
usually each axis is controlled by one servo controller device, so it's stored per axis.
it can be `{ "001", "002", ...}` or `{"Com001", "Com002", ...}`, depending on the user of the `robotControllerAxisProductCode` field.

rdiankov: @MiraiHattori @kanbouchou How is that different from "ElectricMotorActuatorInfo::model_type"?

rdiankov: @MiraiHattori Got it, please also add this to the comment since the current comment is confusing. thanks

MiraiHattori: @rdiankov Sure, I added comments about it. I also updated the comment of `robotControllerAxisMult` about units.

include/openrave/kinbody.h
```diff
@@ -142,6 +142,9 @@ class OPENRAVE_API JointControlInfo_RobotController : public InfoBase
     std::string controllerType; ///< the type of the controller used to control this joint
     //int robotId = -1; ///< index of the robot within that controller?
     boost::array<int16_t, 3> robotControllerAxisIndex = {-1, -1, -1}; ///< indicates which DOF in the robot controller controls which joint axis (up to the DOF of the joint). -1 if not specified/not valid.
+    boost::array<dReal, 3> robotControllerAxisMult = {1.0, 1.0, 1.0}; ///< indicates the multiplier to convert the joint values from the environment units to the robot controller units, per joint axis. (valueInRobotControllerUnits - robotControllerAxisOffset) / robotControllerAxisMult = valueInEnvUnits
```
kanbouchou, line 145: @rdiankov Could you check this comment? 

Current definition is, joint value in orenv unit (usually rad or m) is given by (raw_pulse - ``robotControllerAxisOffset``) / ``robotControllerAxisMult``

Even though this is just a comment and openrave doesn't use ``robotControllerAxisMult`` and ``robotControllerAxisOffset``, our applications is to follow this definition, so quite important.

rdiankov:
thanks~

PR #1188, opened 2023-01-11
allow name mismatch when IAM_AllowRenaming is used

eisoku9618:
```json
{
  "name": "hello world file name",
  "keywords": [
    "part"
  ],
  "bodies": [
    {
      "name": "hello world body name",
      "links": [
        {
          "name": "hello world link name",
          "geometries": [
            {
              "type": "box",
              "halfExtents": [
                0.05,
                0.05,
                0.05
              ]
            }
          ]
        }
      ],
      "isRobot": false
    }
  ]
}
```

When I load this file to openrave environment, I get this error.

```
openravepy._openravepy_0_105.openravepy_int._OpenRAVEException: ('openrave (InvalidArguments): [virtual void Environment::UpdateFromInfo(const OpenRAVE::EnvironmentBase::EnvironmentBaseInfo&, std::vector<boost::shared_ptr<OpenRAVE::KinBody> >&, std::vector<boost::shared_ptr<OpenRAVE::KinBody> >&, std::vector<boost::shared_ptr<OpenRAVE::KinBody> >&, OpenRAVE::UpdateFromInfoMode):2987] pInitBody->GetName() == pKinBodyInfo->_name, (eval hello_world_body_name == hello world body name) names should be matching', 'InvalidArguments')
```

This is because when we add a kinbody in `UpdateFromInfo()` of `environment-core.h`, we use `IAM_AllowRenaming` for new body or robot, so name can be changed e.g. ` ` (white space) -> `_` (underscore) but we compare the original name and the renamed name. For new body or robot, we use `IAM_AllowRenaming` so we don't need to check name matching in that case.

### DIFF
--- a/src/libopenrave-core/environment-core.h
+++ b/src/libopenrave-core/environment-core.h
@@ -2940,7 +2940,7 @@ public:
                 // for new body or robot
                 KinBodyPtr pNewBody;
                 if (pKinBodyInfo->_isRobot) {
-                    RAVELOG_VERBOSE_FORMAT("add new robot %s", pKinBodyInfo->_id);
+                    RAVELOG_VERBOSE_FORMAT("add new robot id=%s, name=%s", pKinBodyInfo->_id%pKinBodyInfo->_name);
                     RobotBasePtr pRobot = RaveCreateRobot(shared_from_this(), pKinBodyInfo->_interfaceType);
                     if( !pRobot ) {
                         pRobot = RaveCreateRobot(shared_from_this(), "");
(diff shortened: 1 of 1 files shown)


rdiankov:
not sure we should be supporting this case....
how about calling `utils::ConvertToOpenRAVEName` in `UpdateFromInfo` ?

eisoku9618:
@rdiankov  Thank you for checking.

> not sure we should be supporting this case....

There was a problem that a user used webapi to register an item with white spaces in the name and our system does not start properly with that `names should be matching` exception. If openrave should not support this case, I think we should reject such a query in webapi side and change `IAM_AllowRenaming` to `IAM_StrictNameChecking` in `UpdateFromInfo`. Personally I prefer this (not supporting white spaces in openrave) because our registraion UI does not allow a user to input white spaces in `bodies/name`.

cc @woswos

PR #1191, opened 2023-01-19
Statically linked plugins for OpenRAVE

undisputed-seraphim:
This patch allows OpenRAVE to be fully statically linked with a select set of plugins.

The modules are:
- Base robots
- Base controllers
- Base samplers
- Base sensors
- FCL

This feature can be toggled by selecting `OPT_STATIC_PLUGINS=ON` in CMake.

Note that selecting `OPT_STATIC_PLUGINS=ON` will also turn off `OPT_PLUGINS`; they are conflicting flags. Therefore, selecting `OPT_STATIC_PLUGINS=ON` will disable the plugin loading functionality.

You must still toggle `OPT_STATIC=ON` if you wish to obtain a `libopenrave.a` archive instead of a `libopenrave.so` shared object.


rdiankov:
thanks

Issue #1189, opened 2023-01-13
Error when building code: 'error: ‘directory_options’ is not a member of ‘fs’; did you mean ‘directory_file’'


jstmn:
Hi, 

I'm getting the following error when trying to build this library:
```
[ 52%] Building CXX object src/libopenrave/CMakeFiles/libopenrave.dir/trajectory.cpp.o
/home/jstm/repos/openrave/src/libopenrave/planningutils.cpp: In member function ‘virtual int OpenRAVE::planningutils::DynamicsCollisionConstraint::Check(const std::vector<double>&, const std::vector<double>&, const std::vector<double>&, const std::vector<double>&, OpenRAVE::dReal, OpenRAVE::IntervalType, int, OpenRAVE::ConstraintFilterReturnPtr)’:
/home/jstm/repos/openrave/src/libopenrave/planningutils.cpp:2694:23: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
 2694 |         for (i = 0; i < params->GetDOF(); i++,itres++) {
      |                     ~~^~~~~~~~~~~~~~~~~~
/home/jstm/repos/openrave/src/libopenrave/planningutils.cpp:2737:23: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
 2737 |         for (i = 0; i < params->GetDOF(); i++,itres++) {
      |                     ~~^~~~~~~~~~~~~~~~~~
/home/jstm/repos/openrave/src/libopenrave/planningutils.cpp:2802:19: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
 2802 |     for (i = 0; i < params->GetDOF(); i++) {
      |                 ~~^~~~~~~~~~~~~~~~~~
/home/jstm/repos/openrave/src/libopenrave/planningutils.cpp:2910:35: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
 2910 |                     for( i = 0; i < numroots; ++i) {
      |                                 ~~^~~~~~~~~~
[...]

PR #1194, opened 2023-01-28
Fix regression in master branch (v0.105.0)

PeterBowman:
This patchset targets:
* https://github.com/rdiankov/openrave/commit/d6469af022b0c44037df69d11de66a7ac4d14b38: affects Boost <1.72.0 (Ubuntu 20.04 installs 1.71.0, [example](https://github.com/roboticslab-uc3m/openrave-yarp-plugins/actions/runs/4020262953/jobs/6907987221#step:17:364)), cc @undisputed-seraphim
  * fixes https://github.com/rdiankov/openrave/issues/1189
* a long-standing missing header inclusion that for some reason began causing trouble on Ubuntu 20.04/22.04 after the recent merge to master ([example](https://github.com/roboticslab-uc3m/openrave-yarp-plugins/actions/runs/4020262953/jobs/6907987788#step:17:352))
* incidentally, a few usual build-related directories missing in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,9 @@ msvc_soqt/
 src/models/
 
 myrelpath.pyc
+
+/build*/
+/.vs/
+/.vscode/
+/__pycache__/
+/sympy/
--- a/include/openrave/logging.h
+++ b/include/openrave/logging.h
@@ -20,6 +20,7 @@
 #ifndef OPENRAVE_LOGGING_H
 #define OPENRAVE_LOGGING_H
 
+#include <string>
 #include <vector>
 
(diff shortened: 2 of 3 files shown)


cielavenir:
should we include <string> as well using this oppotunity?

PeterBowman:
> should we include `<string>` as well using this oppotunity?

Sure, it's done now. I thought about that, but then I noticed it is probably not the only other header missing and eventually left it aside since the code compiled fine for me after adding the vector thing. Anyway, I have added that header, too.

cielavenir (approved):

rdiankov:
thanks guys

PR #1196, opened 2023-01-30
Fix inconsistent override keyword, shadowing variables, fix iterating with type that requires copying

yoshikikanemoto:
detected by clang


yoshikikanemoto:
test: Pipeline #479417

rdiankov:
thanks~

PR #1197, opened 2023-01-30
Silence warnings for boost::bind headers, unused deprecated headers, unused variables, and uninitialized variables

undisputed-seraphim:
## Foreword

OpenRAVE generates _a lot_ of warnings during compilation. Many times I need to track down a single error buried underneath all the warnings, and while the warnings are harmless, they get in the way of debugging and troubleshooting. Therefore I would like to silence these warnings as much as I can.

BEFORE: clean build generates 4399 lines of output text.
AFTER: clean build generates 3258 lines of output text.
__25.94%__ improvement!

## What is changed
- Starting from boost 1.73, `boost/bind.hpp` header emits a warning if you try to use placeholders without `using namespace boost::placeholders;`. The new behaviour has been available since before boost 1.50, so it is safe to update code to use the new behaviour.
- `boost/iterator.hpp` is deprecated at boost 1.74 and it emits a warning for every file that includes this header. It has been obsolete since boost 1.56 and is just an empty file that `#include <iterator>`, [which is itself obsolete](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0174r1.html#2.1).
- Silence warnings about variables that have not been initialized prior to usage.
- Silence warnings about unused variables, and variables that are set but not used.

## What is _not_ changed
- Numpy 1.7 API warnings - Cannot be fixed unless we migrate away from Python 2, I think
- Warnings about comparison of integer expressions of different signedness - Will deal with it another time, requires  discussion
- Warnings about deprecated functions from non-boost libraries - No direct migration route without changing more code than I'd like
- `unused-local-typedefs` - Comes from a set of old Boost headers vendored in `plugins/include`, but changing it requires careful testing because this affects some delicate calculations.
- Redefined macros - Mostly pertains to the `FOREACH`, `FOREACH_NOINC` macros, ideally I'd like to just get rid of all of them at once since we have range-based for-loops built into the language now.

## Bonus round
In commit [ee4a055](https://github.com/rdiankov/openrave/pull/1197/commits/ee4a0559dd7710612049bda31b58fe2ea20ead98), an assert compares two values erroneously, contrary to the comment that follows... that's a bug that would have been caught if we checked the warnings instead of ignoring them :wink: 

Pipeline #479554

### DIFF
--- a/plugins/rplanners/cubicretimer.cpp
+++ b/plugins/rplanners/cubicretimer.cpp
@@ -254,18 +254,18 @@ protected:
             _v0vel[i] = *(itdataprev+info->gvel.offset+i);
             _v1vel[i] = *(itdata+info->gvel.offset+i);
         }
-        dReal deltatime = *(itdata+_timeoffset);
-        dReal ideltatime = 1/deltatime;
-        dReal ideltatime2 = ideltatime*ideltatime;
-        for(int i=0; i < info->gvel.dof; ++i) {
-            dReal px = *(itdata+info->gpos.offset+i) - *(itdataprev+info->gpos.offset+i);
-            dReal v0 = *(itdataprev+info->gvel.offset+i);
-            dReal v1 = *(itdata+info->gvel.offset+i);
-            dReal c3 = (v1 + v0 - 2*px*ideltatime)*ideltatime2;
-            dReal c2 = (3*px*ideltatime - (2*v0 + v1))*ideltatime;
+//        dReal deltatime = *(itdata+_timeoffset);
+//        dReal ideltatime = 1/deltatime;
+//        dReal ideltatime2 = ideltatime*ideltatime;
+//        for(int i=0; i < info->gvel.dof; ++i) {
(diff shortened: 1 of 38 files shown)


plugins/rplanners/cubicretimer.cpp
```diff
@@ -254,18 +254,18 @@ class CubicTrajectoryRetimer : public TrajectoryRetimer
             _v0vel[i] = *(itdataprev+info->gvel.offset+i);
             _v1vel[i] = *(itdata+info->gvel.offset+i);
         }
-        dReal deltatime = *(itdata+_timeoffset);
```
rdiankov, line 257: @Puttichai can you check this out? Was this intentional to have dead code like this? or did you not finish something?

undisputed-seraphim: This line was last changed 10 years ago [here](https://github.com/rdiankov/openrave/commit/658a6e1b09b8c17d1005a0dce8231386aa6a0c9d#diff-379b23408e41198919e1c9a87048dec7bc2da2f13c06ee412cf13f3daa3052b8R235) if it helps

rdiankov:
thanks~

PR #1199, opened 2023-02-06
Support Links with Cyclic Hierarchy

justism:
Loading a scene file with 2 KinBody links that circularly reference each other as parent and child will get stuck in an infinite recursion (and eventually seg faults) when traversing the openscenegraph nodes.
 
This PR adds cache of seen nodes when traversing to prevent the infinite recursion.

Attached is a simple scene JSON file that will reproduce the issue.
[test.mujin.txt](https://github.com/rdiankov/openrave/files/10661291/test.mujin.txt) (renamed with .txt because .json not supported on github)


rdiankov:
thanks

PR #1202, opened 2023-02-10
pytest: Fix python tests involving RaveLoadPlugin

undisputed-seraphim:
Due to the change in the way plugins are loaded in `RaveLoadPlugins` (https://github.com/rdiankov/openrave/pull/1137): File names must be non-truncated (`myplugin` -> `libmyplugin.so`).

Pipeline #486583

### DIFF
--- a/docs/mainpage.dox
+++ b/docs/mainpage.dox
@@ -185,15 +185,15 @@ openrave --loadplugin ./libplugincpp.so
 - Another way is to load it from the C++/Python/APIs:<br/>
 \b C++
 \code
-RaveLoadPlugin(env,"plugincpp");
+RaveLoadPlugin(env,"libplugincpp.so");
 \endcode
 \b Python
 \verbatim
-RaveLoadPlugin('plugincpp')
+RaveLoadPlugin('libplugincpp.so')
 \endverbatim
 \b Octave
 \verbatim
-orEnvLoadPlugin('plugincpp');
+orEnvLoadPlugin('libplugincpp.so');
 \endverbatim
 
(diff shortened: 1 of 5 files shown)


rdiankov:
@undisputed-seraphim Just FYI, it looks like windows support is being dropped.

undisputed-seraphim:
> @undisputed-seraphim Just FYI, it looks like windows support is being dropped.

I can't tell if this is just a remark or if you'd like something to be changed :sweat_smile: 

But yeah, openrave isn't buildable on Windows since I joined, but if you'd like that to change, I can make the necessary changes.

rdiankov:
@undisputed-seraphim Not buildable on windows could just be a build problem. However, this change would make it incompatible from a design perspective, so at least we have to worry about that.

rdiankov:
@undisputed-seraphim let's try to figure out a design that doesn't involve users knowing the prefix and extensions of the plugins

undisputed-seraphim:
@rdiankov Okay, let me come up with another solution.

cielavenir:
Why don't you simply add PLUGIN_EXT when loading?

cielavenir:
@undisputed-seraphim By the way, after https://github.com/rdiankov/openrave/pull/1137 , OPENRAVE_PLUGINS='' has to be set explicitly, which does not seem good for plain openrave installation. /cc @lazydroid

cielavenir:
Those two issue are addressed by https://github.com/rdiankov/openrave/pull/1227

cielavenir:
will be replaced by #1227

PR #1203, opened 2023-02-13
python: Scripts installed into /usr/bin are now python version agnostic.

undisputed-seraphim:
Apologies - we missed out changing these 3 files. The value `PYTHON_EXECUTABLE_NAME` used to configure the files no longer exist (and aren't necessary anymore).

Instead a version agnostic python shebang will allow users to freely pick whichever Python version they like, since both versions exist now.


cielavenir (approved):
thanks a lot~

undisputed-seraphim:
Addendum: Added a commit that fixed some linking bugs that got left out in the python2+3 merge.

rdiankov:
sure, thanksl~

PR #1204, opened 2023-02-15
Unify InitGeometries behavior

eisoku9618:
There are 2 definitions of `InitGeometries` but the behavior for `bForceRecomputeMeshCollision` is different, and I think this is not intentional. `bForceRecomputeMeshCollision` was introduced by 55757fee6bf513d0397d868637cdd6099da5ab3a


rdiankov:
sure, thanks

PR #1206, opened 2023-02-19
Use Gary's ConvertStringToUnicode converting implementation

cielavenir:
Without `PyUnicode_Decode `, I found that `robot.GetName()` becomes bytes on Python3, and it is not the same as Python2.

Python2

```
In [1]: 'xyz' == 'xyz'
Out[1]: True

In [2]: b'xyz' == 'xyz'
Out[2]: True

In [3]: u'xyz' == 'xyz'
Out[3]: True
```

Python3

```
In [1]: 'xyz' == 'xyz'
Out[1]: True

In [2]: b'xyz' == 'xyz'
Out[2]: False

In [3]: u'xyz' == 'xyz'
Out[3]: True
```

I have also confirmed that stretch (boost::python)'s `robot.GetName()` type is unicode. so the behavior should be the same on pybind11.

----

pipelineid=338041 Pipeline #492326

### DIFF
--- a/python/bindings/include/openravepy/pybind11/pybind11_bindings.h
+++ b/python/bindings/include/openravepy/pybind11/pybind11_bindings.h
@@ -114,14 +114,8 @@ namespace py = pybind11;
 
 inline py::object ConvertStringToUnicode(const std::string& s)
 {
-    /*
-       TGN: Is the following an alternative?
-       ```
-       PyObject *pyo = PyUnicode_Decode(s.c_str(), s.size(), "utf-8", nullptr);
-       return py::cast<py::object>(pyo); // py::handle_to_object(pyo);
-       ```
-     */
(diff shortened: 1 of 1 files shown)


rdiankov:
@cielavenir Does python2 behavior change?

cielavenir:
robot.GetName() type will be changed from str to unicode on Python2, in this term the Python2 behavior will change, although the pipeline result looks reasonable. The thing is, boost::python edition was using unicode. So this can be said as a kind of regression when adapting pybind11, because I did not review Gary's code careful enough.

It is about one year after we migrated to pybind11. Maybe changing the spec now might be dangerous. If so, I can change so that Gary's "alternative method" is used only on Python3.

rdiankov:
I think okay, thanks!

PR #1208, opened 2023-03-03
Fix Python3 WriteToMemory msgpack; should return bytes

cielavenir:
if filetype == "msgpack", bytes have to be returned for sure. but not sure if whether bytes or str should be returned on json/collada.

testopenrave (non-legacy) **fully passed** locally on py3.


rdiankov:
@cielavenir I'm a little worried about returning different object types. How about we always return bytes for python3?

cielavenir:
@rdiankov sure, I updated and retested testopenrave.

I suppose no need to run /test as this does not affect existing code paths.

rdiankov:
thanks

PR #1210, opened 2023-03-15
lock env in TaskManipulation

cielavenir:
today we saw a deadlock

env.Add should lock env because Python world can be deadlocked due to GIL

by the way, should we lock env for other interfaces?


yoshikikanemoto (approved):

rdiankov:
thanks!

PR #1212, opened 2023-03-17
steal reference count when constructing pybind11::object for PyUnicode_Decode

cielavenir:
According to https://docs.python.org/2.7/c-api/unicode.html#c.PyUnicode_Decode , PyUnicode_Decode returns New reference. Meanwhile, pybind11::object internally increases the reference count when constructed. It seems like the latter causes double-increasing. To disable auto-increasing reference count, we should use py::reinterpret_steal.

This can also be said to handle_to_object, whose user is PyInt_FromLong.


cielavenir:
Pipeline #508207

pipelineid=343554

lazydroid (approved):
LGTM, thanks!

cielavenir:
appendix

- boost::python
    - object(handle<>(pyo)) no-increment
    - object(handle<>(borrowed(pyo))) increment
- pybind11
    - cast<object>(pyo) **increment, which is the issue**
    - reinterpret_steal<object>(pyo) no-increment

rdiankov:
thanks~

PR #1214, opened 2023-03-19
fixed python binding of CheckColliison for Ray was not working.

yoshikikanemoto:
overloads are resolved by the order they were registered. needed to be registered before overloads by "object". https://pybind11.readthedocs.io/en/stable/advanced/functions.html#overload-resolution-order


rdiankov:
thanks~

PR #1215, opened 2023-03-21
specify -Wno-shadow on collada-dom and qhull

cielavenir:
We install qhull from apt and build collada-dom via jhbuild, but in some cases people could want to build those from openrave/3rdparty. Then -Werror=shadow causes problems.


cielavenir:
Pipeline #510172

pipelineid=344241

rdiankov:
sure..

PR #1218, opened 2023-03-30
fcl: Split files into header+implementation

undisputed-seraphim:
Implementation code moved verbatim from headers to sources with no changes.

Motivation: I would like to test changes to fcl without waiting for a recompile for everything in the whole module. Additionally, navigating the code and seeing the relationships between the functions is also easier.
(also it looks nicer)

Pipeline #514930


rdiankov:
thanks~

PR #1220, opened 2023-04-06
GetCameraDistanceToFocus() and GetCameraTransform() can be accessed f…

ziyan:
…rom non-viewer thread.

Tested locally.

Test pipeline: 519430


rdiankov:
thanks~

PR #1225, opened 2023-04-18
Geometry info crop xyz20230413

LLipter:
Add `CropContainerMarginXYZ` field to `GeometryInfo` object, which is a list of 6 floating numbers.


include/openrave/kinbody.h
```diff
@@ -359,7 +359,13 @@ class OPENRAVE_API KinBody : public InterfaceBase
                    && _fTransparency == other._fTransparency
                    && _bVisible == other._bVisible
                    && _bModifiable == other._bModifiable
-                   && _calibrationBoardParameters == other._calibrationBoardParameters;
+                   && _calibrationBoardParameters == other._calibrationBoardParameters
+                   && _vCropContainerMarginXYZ[0] == other._vCropContainerMarginXYZ[0]
```
ziyan, line 363: boost array has no `operator==` overload?

include/openrave/kinbody.h
```diff
@@ -490,6 +496,7 @@ class OPENRAVE_API KinBody : public InterfaceBase
         float _fTransparency = 0; ///< value from 0-1 for the transparency of the rendered object, 0 is opaque
         bool _bVisible = true; ///< if true, geometry is visible as part of the 3d model (default is true)
         bool _bModifiable = true; ///< if true, object geometry can be dynamically modified (default is true)
+        boost::array<dReal, 6> _vCropContainerMarginXYZ = {0, 0, 0, 0, 0, 0};
```
ziyan, line 499: It's `XYZXYZ`, not `XYZ`

undisputed-seraphim:
Can/Should we put the operators in `geometry.h`?

Alternatively, should we also write fuller 6-dim vector implementations so that it looks more consistent with `RaveVector` in `geometry.h`? (Actually I think we should use `Eigen3` vectors and matrices everywhere so we don't have to keep reinventing the wheel each time we need a new N-dim vector)

include/openrave/kinbody.h
```diff
@@ -490,6 +517,7 @@ class OPENRAVE_API KinBody : public InterfaceBase
         float _fTransparency = 0; ///< value from 0-1 for the transparency of the rendered object, 0 is opaque
         bool _bVisible = true; ///< if true, geometry is visible as part of the 3d model (default is true)
         bool _bModifiable = true; ///< if true, object geometry can be dynamically modified (default is true)
+        boost::array<dReal, 6> _vCropContainerMarginsXYZXYZ = {0, 0, 0, 0, 0, 0};
```
rdiankov, line 520: I think the 6 dim array has been hard to understand in the system and will be even harder if introduced into openrave this way. How about `Vector negCropContainerMargins, posCropContainerMargins`?

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -1031,6 +1053,22 @@ void KinBody::GeometryInfo::DeserializeJSON(const rapidjson::Value &value, const
                 _meshcollision.Clear();
             }
         }
+        if (value.HasMember("negativeCropContainerMargins")) {
+            orjson::LoadJsonValueByKey(value, "negativeCropContainerMargins", vGeomDataTemp);
+            vGeomDataTemp *= fUnitScale;
+            if (vGeomDataTemp != _vNegativeCropContainerMargins) {
+                _vNegativeCropContainerMargins = vGeomDataTemp;
+                _meshcollision.Clear();
```
ziyan, line 1061: You do not need to clear mesh here since this parameter does not affect mesh generation.

LLipter: ok, they are removed.

rdiankov:
for now accepting, but would like visualizations.

PR #1226, opened 2023-04-26
Fix redundant end effector collision check

yoshikikanemoto:
1. end effector collision check functions seem to be computing collision twice for the same link.
2. removed unnecessary `TransformSaver`
3. small optimizations to fclrave

@eisoku9618

### DIFF
--- a/src/libopenrave/robotmanipulator.cpp
+++ b/src/libopenrave/robotmanipulator.cpp
@@ -854,12 +854,7 @@ bool RobotBase::Manipulator::IsChildLink(const KinBody::Link &link) const
             return false;
         }
     }
-    for(size_t ijoint = 0; ijoint < probot->GetJoints().size(); ++ijoint) {
-        if( probot->DoesAffect(ijoint,ilink) && !probot->DoesAffect(ijoint,iattlink) ) {
-            return true;
-        }
-    }
-    return false;
+    return true;
 }
 
 void RobotBase::Manipulator::GetIndependentLinks(std::vector<LinkPtr>& vlinks) const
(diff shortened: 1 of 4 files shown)


src/libopenrave/robotmanipulator.cpp
```diff
@@ -1125,22 +1031,24 @@ bool RobotBase::Manipulator::CheckEndEffectorSelfCollision(CollisionReportPtr re
...
-                if( probot->CheckLinkSelfCollision(ilink,report) ) {
-                    if( !bAllLinkCollisions ) { // if checking all collisions, have to continue
-                        return true;
+        // check all gripper links including
+        // 1. links that are rigidly attached to the end effector
+        // 2. links that are controlled by joints other than arm joints
+        // 3. links that are connected with passive but non-static joints
+        if( bIgnoreManipulatorLinks ) {
+            FOREACHC(itindependentlink,vindependentinks) {
+                if( *itlink != *itindependentlink && (*itindependentlink)->IsEnabled() ) {
```
kanbouchou, line 1040: Since result of ``CheckCollision(linkA, linkB)`` and  ``CheckCollision(linkB, linkA)`` should be the same, we can skip one of them by something like ``ilink < (**itindependentlink).GetIndex()`` instead of ``*itlink != *itindependentlink``?

src/libopenrave/robotmanipulator.cpp
```diff
@@ -854,12 +854,7 @@ bool RobotBase::Manipulator::IsChildLink(const KinBody::Link &link) const
             return false;
         }
     }
-    for(size_t ijoint = 0; ijoint < probot->GetJoints().size(); ++ijoint) {
-        if( probot->DoesAffect(ijoint,ilink) && !probot->DoesAffect(ijoint,iattlink) ) {
-            return true;
-        }
-    }
-    return false;
+    return true;
```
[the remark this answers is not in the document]
yoshikikanemoto: if `link` and `ee` are connected with static joint, `__pEffector->IsRigidlyAttached(link)` returns true, so this change doesn't matter. the change here is for this function to have the same behavior as other functions for `links that are connected with passive but non-static joints`

src/libopenrave/robotmanipulator.cpp
yoshikikanemoto: in my understanding from `GetIndependentLinks`, `independentlinks` should never include `*itlink` and the check here is "just in case". so i think we don't need to change the logic here.

kanbouchou: I see, they are mutually exclusive, then sure I think current logic is fine.

src/libopenrave/robotmanipulator.cpp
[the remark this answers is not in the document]
kanbouchou: Right, passive joints (not static).

If the behavior change is intended one, I think the function name ``IsChildLink`` is a bit strange, because ``link`` here is more like a sibling than a child of ``ee``. I'll leave to @rdiankov about the function name.

kanbouchou (approved):

rdiankov:
thanks~

PR #1227, opened 2023-04-28
Canonicalize libraryname

cielavenir:
This pull request tries to resolve two issues:

- Canonicalize LoadPlugin argument
- Allow OPENRAVE_PLUGINS not set, which is useful for plain openrave installation

I checked that test_global.py works.

### DIFF
--- a/src/libopenrave/plugindatabase.cpp
+++ b/src/libopenrave/plugindatabase.cpp
@@ -216,9 +216,11 @@ void DynamicRaveDatabase::ReloadPlugins()
 bool DynamicRaveDatabase::LoadPlugin(const std::string& libraryname)
 {
     std::string canonicalizedLibraryname = libraryname;
+#ifndef _WIN32
     if(canonicalizedLibraryname.substr(0, 3) != "lib") {
         canonicalizedLibraryname = "lib" + canonicalizedLibraryname;
     }
+#endif
     if(canonicalizedLibraryname.substr(canonicalizedLibraryname.size() - PLUGIN_EXT.size()) != PLUGIN_EXT) {
         canonicalizedLibraryname += PLUGIN_EXT;
     }
--- a/src/libopenrave/plugindatabase.cpp
+++ b/src/libopenrave/plugindatabase.cpp
@@ -133,21 +133,19 @@ DynamicRaveDatabase::~DynamicRaveDatabase()
 void DynamicRaveDatabase::Init()
 {
     const char* pOPENRAVE_PLUGINS = getenv("OPENRAVE_PLUGINS"); // getenv not thread-safe?
-    if (!pOPENRAVE_PLUGINS) {
-        RAVELOG_WARN("Failed to read environment variable OPENRAVE_PLUGINS");
-        return;
-    }
     std::vector<std::string> vplugindirs;
-    utils::TokenizeString(pOPENRAVE_PLUGINS, s_delimiter, vplugindirs);
-    for (int iplugindir = vplugindirs.size() - 1; iplugindir > 0; iplugindir--) {
-        int jplugindir = 0;
-        for(; jplugindir < iplugindir; jplugindir++) {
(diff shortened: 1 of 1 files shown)


cielavenir:
Pipeline #532669

src/libopenrave/plugindatabase.cpp
```diff
@@ -124,21 +133,19 @@ DynamicRaveDatabase::~DynamicRaveDatabase()
 void DynamicRaveDatabase::Init()
 {
     const char* pOPENRAVE_PLUGINS = getenv("OPENRAVE_PLUGINS"); // getenv not thread-safe?
-    if (!pOPENRAVE_PLUGINS) {
-        RAVELOG_WARN("Failed to read environment variable OPENRAVE_PLUGINS");
-        return;
```
lazydroid, line 129: why do we remove this warning?

src/libopenrave/plugindatabase.cpp
```diff
@@ -208,27 +215,25 @@ void DynamicRaveDatabase::ReloadPlugins()
 
 bool DynamicRaveDatabase::LoadPlugin(const std::string& libraryname)
 {
-    // If the libraryname matches any of the existing loaded libraries, then reload it
+    std::string canonicalizedLibraryname = libraryname;
+    if(canonicalizedLibraryname.substr(0, 3) != "lib") {
+        canonicalizedLibraryname = "lib" + canonicalizedLibraryname;
```
lazydroid, line 220: there's a selection of PLUGIN_EXT for different OS, do the libraries for WIN32 always supposed to have `lib` prefix?

src/libopenrave/plugindatabase.cpp
cielavenir: That warning was added by https://github.com/rdiankov/openrave/pull/1137 but it is wrong. If failed to read, it should be just ignored.

src/libopenrave/plugindatabase.cpp
cielavenir: added #ifndef _WIN32, thanks a lot!

src/libopenrave/plugindatabase.cpp
lazydroid: shall we just drop the `return;` statement and keep the warning in the logs?

cielavenir: `utils::TokenizeString(pOPENRAVE_PLUGINS` will cause segmentation fault then

src/libopenrave/plugindatabase.cpp
```diff
@@ -124,21 +133,19 @@ DynamicRaveDatabase::~DynamicRaveDatabase()
...
+                }
+            }
+            if (jplugindir < iplugindir) {
+                vplugindirs.erase(vplugindirs.begin()+iplugindir);
             }
-        }
-        if (jplugindir < iplugindir) {
-            vplugindirs.erase(vplugindirs.begin()+iplugindir);
         }
     }
```
cielavenir, line 150: @lazydroid (sorry I don't understand fully but) does this work for you?

```suggestion
    }
    else {
        RAVELOG_WARN("Failed to read environment variable OPENRAVE_PLUGINS");
    }
```

src/libopenrave/plugindatabase.cpp
lazydroid:
```
    if (!!pOPENRAVE_PLUGINS) {
        utils::TokenizeString(pOPENRAVE_PLUGINS, s_delimiter, vplugindirs);
        ...
    } else {
        RAVELOG_WARN("Failed to read environment variable OPENRAVE_PLUGINS");
    }
```
 ?

lazydroid: ah. noticed your comment below, thanks!

lazydroid (approved):
LGTM, thank you!

undisputed-seraphim (approved):
I see no issues :+1:

rdiankov:
thanks~

PR #1228, opened 2023-05-02
visualization on cropping margin

LLipter:
Screenshot from 2023-05-02 10-38-39
Screenshot from 2023-05-02 10-38-56

### DIFF
--- a/plugins/qtosgrave/osgrenderitem.cpp
+++ b/plugins/qtosgrave/osgrenderitem.cpp
@@ -502,6 +502,7 @@ void KinBodyItem::Load()
                     upperBound.z = extents.z;
                     upperBound -= positiveCropContainerMargins;
 
+                    // bottom 4 edges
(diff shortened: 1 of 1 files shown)


ziyan:
@rdiankov I am not sure if:

1. you need planes instead of lines, but it might be too intrusive
2. if you need the ability to turn off this visualization, as it is always on right now

And Ran was also asking if you need the same visualization for crop container empty margins as well, but that's even more clutter on the screen.

Another alternative approach would be to use the visualization manager route.

rdiankov:
please also add visualization for empty container margins.
red is not a good color, that's usually reserved for obstacles. perhaps you should use a shade of the geometry color

LLipter:
I  move this branch to rdiankov:oprenrave repo. See pull request https://github.com/rdiankov/openrave/pull/1231

PR #1230, opened 2023-05-04
install symlink to relative path

cielavenir:
so that the openrave installtion can be moved

/cc @felixvd


rdiankov:
sure, thanks

PR #1231, opened 2023-05-05
Container crop visualization20230428

LLipter:
`CropContainerMargins` are in a darker color.
`CropContainerEmptyMargins` are in a lighter color.

Screenshot from 2023-05-05 10-58-59
Screenshot from 2023-05-05 10-59-09


rdiankov:
thanks~ let's see if people think the crop regions are too annoying

PR #1232, opened 2023-05-05
make PyEnvironmentBase a public API

LLipter:
Ziyan asked me to write a unified `NotifyEnvironmentChanged` implementation. While working on it, I realize I need to have access to `PyEnvironmentBase` in order to make python binding work. Can we mark this class as a public openrave API?


rdiankov:
It's private for a reason, so no.
We have a c++ implementation of NotifyEnvironmentChanged, why not use that?

PR #1235, opened 2023-05-11
Slight improvements on parabolicsmoother2 and timebased checking

snozawa:
# Summary
- this contains slight improvements on parabolicsmoother2 and `DynamicsCollisionConstarint::Check`

# Contents
1. Check `dynamicsConstraintsType` before checking the dynamic limits or torque limits. https://github.com/rdiankov/openrave/commit/99721524d8ebdf11b2d5ef0383b261941343f34f
2. The initial value of `currentBestScore` is set to `1.0`. it looks too huge and most likely, we don't enter the score checking code path. i'm not sure it's intended behavior. but, if not intended, theoretically `0` should be the initial value. https://github.com/rdiankov/openrave/commit/4927dcdb9e54412c1e2c405804916a55637408d5
3. improve dump trajectory stuff for better debugging. mostly, share the single `fileIndex` so that we can track all of the logs from one `PlanPath` execution with the same `fileIndex`. it will be easier to search the relevant files. https://github.com/rdiankov/openrave/commit/958ecf544c02b6edb2a5f61e2f1c365ea36edb3d
4. others : improve the print messages.

### DIFF
--- a/plugins/rplanners/parabolicsmoother2.cpp
+++ b/plugins/rplanners/parabolicsmoother2.cpp
@@ -409,6 +409,13 @@ public:
         _uniformsampler->SetSeed(_parameters->_nRandomGeneratorSeed);
 
         _fileIndexMod = 10000; // for trajectory saving
+        if( !!_logginguniformsampler ) {
+            _fileIndex = _logginguniformsampler->SampleSequenceOneUInt32();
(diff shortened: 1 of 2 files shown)


rdiankov:
thanks!

PR #1236, opened 2023-05-15
Flag to ignore link from self collision checking

kanbouchou:
Background
=============

- With existing way, self collision check can be skipped per link pair via ``forcedAdjacentLinks``.
- Some links are to be excluded from self collision check against all other links, and having to specify ``forcedAdjacentLinks`` for every other link can be tedious.

Proposed change
================

- Introduce a new link property ``isSelfCollisionIgnored``. If true, this link is not checked in self collision computation at all.
- Only FCL and ODE checkers are supported.


include/openrave/kinbody.h
```diff
@@ -991,6 +993,12 @@ class OPENRAVE_API KinBody : public InterfaceBase
         /// \brief returns true if the link is enabled. \see Enable
         bool IsEnabled() const;
 
+        /// \brief Ignores a Link from self collision.
+        void IgnoreSelfCollision(bool bIgnore);
```
rdiankov, line 997: @kanbouchou let's call this `SetIgnoreSelfCollision`

kanbouchou: I changed the name from ``SetIgnoreSelfCollision``

PR #1237, opened 2023-05-16
Make a deep copy on connected body information

LLipter:
## Background

When openrave loads a changed environment JSON dictionary, openrave will first make a copy of the current environment and then apply the environment changes to the copy. Then openrave will compare the old environment and the new environment. If they're different, some components will be re-initialized. But when openrave copy connected body info, it makes a shallow copy. Then the environment change will be applied to both the original environment and the new environment because they point to the same underlying object. Later openrave will not re-initialize the environment because there's no diff. This behaviour will cause environment changes in connected body being not effective until restart the process.

## Proposed change

Make a deep copy.

### DIFF
--- a/src/libopenrave/robotconnectedbody.cpp
+++ b/src/libopenrave/robotconnectedbody.cpp
@@ -424,6 +424,36 @@ void RobotBase::ConnectedBody::ExtractInfo(RobotBase::ConnectedBodyInfo& info) c
 {
     // TODO: extract info from connectedbody to make extra geometry working
     info = _info;
+
+    info._vLinkInfos.resize(_vResolvedLinkNames.size());
+    for(size_t iLinkInfo = 0; iLinkInfo < _vResolvedLinkNames.size(); ++iLinkInfo) {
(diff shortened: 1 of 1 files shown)


src/libopenrave/robotconnectedbody.cpp
```diff
@@ -424,6 +424,35 @@ void RobotBase::ConnectedBody::ExtractInfo(RobotBase::ConnectedBodyInfo& info) c
 {
     // TODO: extract info from connectedbody to make extra geometry working
     info = _info;
+
+    for(size_t iLinkInfo = 0; iLinkInfo < _info._vLinkInfos.size(); ++iLinkInfo) {
+        info._vLinkInfos[iLinkInfo].reset(new KinBody::LinkInfo());
```
ziyan, line 429: `info._vLinkInfos[iLinkInfo] = boost::make_shared<KinBody::LinkInfo>()` to avoid memory leak when exception is thrown.

ziyan:
@LLipter Nice catch! Please update patch version as well.

rdiankov:
nice catch~

PR #1238, opened 2023-05-16
Fix drawtrimesh interface when compiled with pybind11

gywn:
The affected code section allows drawtrimesh to accept an array of color attributes that match the triangle array. This is more efficient than creating many draw handles because OpenGL can consume this array of color attributes directly. This functionality was impacted when using pybind11 to compile the Python interface.

### DIFF
--- a/python/bindings/openravepy_int.cpp
+++ b/python/bindings/openravepy_int.cpp
@@ -2612,7 +2612,7 @@ object PyEnvironmentBase::drawtrimesh(object opoints, object oindices, object oc
             return toPyGraphHandle(_penv->drawtrimesh(vpoints.data(),sizeof(float)*3,pindices,numTriangles,ExtractVector34(ocolors,1.0f)));
         }
         else {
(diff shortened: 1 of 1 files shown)


cielavenir:
By the way I'm a little unclear, do you mean ocolors is not GCed until PyGraphHandle is GCed and your patch fixes that?

gywn:
@cielavenir What failed in my build is `extract<boost::multi_array<float,2> >(ocolors)`. It seems `pybind11::object::cast` does not support `boost::multi_array`. I look through the code base and it seems this is the only place where such a conversion is considered. Runtime error message from pybind11:

```text
Exception RuntimeError: u'Unable to cast Python instance to C++ type (compile in debug mode for details)' in <finalize object at 0x7ff9cce81d00; dead> ignored
```

And sorry about the `auto`, I was unaware/ignorant about this coding style in Mujin/OpenRAVE.

rdiankov:
thanks~

PR #1240, opened 2023-05-26
Avoid clearing Kinematics Geometry hash when adding to environment

kanbouchou:
Background
==============

- Kinematics geometry hash is expensive to compute and we would like to avoid recomputing it unless necessary
- When we clone a kinbody and add to an environment, cloning is copying ``__hashkinematics``, but it was reset inside ``_ComputeInternalInformation`` and ``_PostprocessChangedParameters`` both called from ``Environment::Add``

Proposal
=============
1. Inside ``_ComputeInternalInformation``, do not invalidate kinematics geometry hash.
2. Do not pass ``KinBody::Prop_LinkDynamics`` to ``_PostprocessChangedParameters`` when calling ``Environment::Add`` as dynamics properties should not be changed during this process.

cc @yoshikikanemoto


kanbouchou:
> will make some changes based on offline discussions, setting to draft until completed.

@rdiankov please review again, the final change set is basically to make sure that in call path of adding to environment, make sure we don't invalidate hash. I didn't need to add logic to invalidate hash when related properties changes, it was already all taken care of.

kanbouchou:
@rdiankov changed variable name as discussed offline, from `__hashKinGeomDyn` to `__hashKinematicsGeometryDynamics`

rdiankov:
thanks

PR #1241, opened 2023-06-01
Fix not clearing hash in Destroy

kanbouchou:
In https://github.com/rdiankov/openrave/pull/1240, hash is not cleared any more in the call stack of adding to environment, so instead need to clear it in ``Destroy``


rdiankov:
Thanks

PR #1244, opened 2023-06-07
fixed out of range access in FindRampNDIndex

MiraiHattori:
the code path may access `*_rampnds.end()`, so fixing.

### DIFF
--- a/plugins/rplanners/rampoptimizer/ramp.cpp
+++ b/plugins/rplanners/rampoptimizer/ramp.cpp
@@ -334,7 +334,7 @@ void ParabolicCurve::FindRampIndex(dReal t, int& index, dReal& remainder) const
         int index_ = 0;
         dReal curTime = 0;
         std::vector<Ramp>::const_iterator itramp = _ramps.begin();
-        while ((itramp - 1) != _ramps.end() && t >= curTime) {
+        while( itramp != _ramps.end() && t >= curTime ) {
             curTime += itramp->duration;
             itramp++;
             index_++;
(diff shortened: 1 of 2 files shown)


MiraiHattori:
note: pipeline id is 558171

plugins/rplanners/rampoptimizer/ramp.cpp
```diff
@@ -989,7 +989,7 @@ void ParabolicPath::FindRampNDIndex(dReal t, int& index, dReal& remainder) const
         int index_ = 0;
         dReal curTime = 0;
         std::vector<RampND>::const_iterator itrampnd = _rampnds.begin();
-        while ((itrampnd - 1) != _rampnds.end() && t >= curTime) {
+        while( itrampnd != _rampnds.end() && t >= curTime ) {
```
kanbouchou, line 992: There is similar code in ``ParabolicCurve::FindRampIndex``, please check that as well.

MiraiHattori: @kanbouchou that's true. fixed `PiecewisePolynomial::FindPolynomialIndex` as well. could you review them?

rdiankov: @Puttichai  can you check?

Puttichai: Looks good to me.

yoshikikanemoto: looks good to me too

rdiankov:
thanks~

PR #1250, opened 2023-06-09
Introduce main py

cielavenir:
This moves openrave.py main functionality to `openravepy.__main__.py`, still keeping openrave.py existence.

Currently openrave.py is Python2 stuff, but this will enable `python2 -m openravepy` and `python3 -m openravepy`, namely you can explicitly set python version to use.

Our standard sphnix runs in python3, but openrave.py (in master) runs in python2, which will cause problems in running shell-block https://github.com/rdiankov/openrave/blob/master/docs/source/command_line_tools.rst . I suggest to change those (eg) `openrave.py --listdatabases` to `python3 -m openravepy --listdatabases`.

@felixvd


rdiankov:
openrave.py is installed in `bin`, shouldn't do already be in the python2/3 path?

cielavenir:
It depends on what `#!/usr/bin/env python` refers. Currently we cannot explicitly specify the python version.

rdiankov:
got it, in that case, let's try this. thanks

PR #1251, opened 2023-06-12
Add API for setting visibility of crop margins, render crop margins as boxes, fix transparency rendering

seanzhang15:
This PR: 
- adds a command to selectively visualize crop container margins
- renders crop margins as colored boxes instead of line meshes to better visualize the remaining volume
- adds labels to indicate the type of crop margin. 
- switches to the proper logic for rendering transparency after tests have shown no difference in performance. 

Crop-margin visualization:
image


cielavenir:
It causes meshes to show the wireframes, which makes harder to view the scene.

Please make mesh behavior as before.

Screenshot from 2023-06-12 12-58-31

cielavenir:
It is like this in production branch:

Screenshot from 2023-06-12 13-10-15

cielavenir:
(I will restart the performance test after that issue is corrected)

cielavenir (changes requested):
It causes meshes to show the wireframes

seanzhang15:
New image:

image

The outline is still displayed, but the ones in the back are properly covered by the robot, which is semi-transparent here. 

cc @braineo

cielavenir (approved):

cielavenir:
now it would be ok

braineo:
transparent box inside another transparent box

image

rdiankov:
thanks~

PR #1252, opened 2023-06-12
Compile error when trying to index a py::object with size_t key

undisputed-seraphim:
When compiling natively with certain combinations of flags:

```
cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo \
    -DODE_USE_MULTITHREAD=ON \
    -DOPT_IKFAST_FLOAT32=OFF \
    -DOPENRAVE_MSGPACK=OFF  \
    -DOPT_LOG4CXX=ON \
    -DBoost_NO_BOOST_CMAKE=OFF \
    -DUSE_PYBIND11_PYTHON_BINDINGS=ON \
    -DOPT_INSTALL_3DMODELDATA=OFF \
    -DOPT_MSGPACK=OFF \
    -DOPT_CURL=OFF \
    -DOPT_OCTAVE=OFF \
    -DOPT_COLLADA=OFF \
    -DOPT_MATLAB=OFF \
    -DOPT_STATIC=OFF \
    -DOPT_PYTHON=OFF  \
    -DCMAKE_CXX_FLAGS="-Wno-shadow"
```

I ran into a compilation error:

```
/home/user/Code/openrave/python/bindings/openravepy_global.cpp:355:75: error: invalid conversion from ‘size_t’ {aka ‘long unsigned int’} to ‘const char*’ [-fpermissive]
  355 |         vOrientedBox[iOrientedBox] = ExtractOrientedBox(pyOrientedBoxList[iOrientedBox]);
      |                                                                           ^~~~~~~~~~~~
      |                                                                           |
      |                                                                           size_t {aka long unsigned int}
```

The reason is because `py::object::operator[]` takes a string (`const char*`) instead of an integral value like `size_t`.

I imagine that what we really want is to cast the `py::object` into a `py::list` before trying to index it, so I applied a fix that does so accordingly.

I'm not sure why this error is not encountered in other builds. Perhaps it's due to very old versions of pybind11?

----

Edit by ciel: This is a new instances of https://github.com/rdiankov/openrave/pull/1125 .

### DIFF
--- a/plugins/rplanners/parabolicsmoother2.cpp
+++ b/plugins/rplanners/parabolicsmoother2.cpp
@@ -12,6 +12,7 @@
 // You should have received a copy of the GNU Lesser General Public License along with this program.
 // If not, see <http://www.gnu.org/licenses/>.
 #include "openraveplugindefs.h"
+#include <cfloat>
 #include <fstream>
 #include <openrave/planningutils.h>
 
--- a/python/bindings/openravepy_global.cpp
+++ b/python/bindings/openravepy_global.cpp
@@ -349,10 +349,11 @@ std::vector<OrientedBox> ExtractOrientedBoxArray(py::object pyOrientedBoxList)
     if( IS_PYTHONOBJECT_NONE(pyOrientedBoxList) ) {
         return {};
     }
-    const size_t arraySize = len(pyOrientedBoxList);
+    py::list pyOrientedBoxArray(pyOrientedBoxList);
+    const size_t arraySize = pyOrientedBoxArray.size();
     std::vector<OrientedBox> vOrientedBox(arraySize);
     for(size_t iOrientedBox = 0; iOrientedBox < arraySize; ++iOrientedBox) {
-        vOrientedBox[iOrientedBox] = ExtractOrientedBox(pyOrientedBoxList[iOrientedBox]);
+        vOrientedBox[iOrientedBox] = ExtractOrientedBox(pyOrientedBoxArray[iOrientedBox]);
     }
     return vOrientedBox;
 }
--- a/python/bindings/openravepy_global.cpp
+++ b/python/bindings/openravepy_global.cpp
@@ -349,11 +349,10 @@ std::vector<OrientedBox> ExtractOrientedBoxArray(py::object pyOrientedBoxList)
     if( IS_PYTHONOBJECT_NONE(pyOrientedBoxList) ) {
         return {};
     }
-    py::list pyOrientedBoxArray(pyOrientedBoxList);
-    const size_t arraySize = pyOrientedBoxArray.size();
+    const size_t arraySize = len(pyOrientedBoxList);
     std::vector<OrientedBox> vOrientedBox(arraySize);
     for(size_t iOrientedBox = 0; iOrientedBox < arraySize; ++iOrientedBox) {
-        vOrientedBox[iOrientedBox] = ExtractOrientedBox(pyOrientedBoxArray[iOrientedBox]);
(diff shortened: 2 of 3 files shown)


cielavenir:
@undisputed-seraphim what exact version of pybind11 did you use? (v2.9.2 / v2.9_ty / mujin_v2.9.2 / other)

cielavenir:
if you use STOCK v2.9.2, what you need is INSTEAD:

```diff
-vOrientedBox[iOrientedBox] = ExtractOrientedBox(pyOrientedBoxList[iOrientedBox]);
+vOrientedBox[iOrientedBox] = ExtractOrientedBox(pyOrientedBoxList[py::to_object(iOrientedBox)]);
```

This is described in https://github.com/rdiankov/openrave/pull/1125 .

undisputed-seraphim:
> @undisputed-seraphim what exact version of pybind11 did you use? (v2.9.2 / v2.9_ty / mujin_v2.9.2 / other)

The version I am using is 2.9.2 stock.

cielavenir:
Please try py::to_object(iOrientedBox) then

undisputed-seraphim:
Pipeline #561845

cielavenir:
@undisputed-seraphim Thank you! Are there any other places?

undisputed-seraphim:
I didn't encounter any more compile error so I think that's all

cielavenir:
thank you for checking!

cielavenir (approved):

plugins/rplanners/parabolicsmoother2.cpp
```diff
@@ -12,6 +12,7 @@
 // You should have received a copy of the GNU Lesser General Public License along with this program.
 // If not, see <http://www.gnu.org/licenses/>.
 #include "openraveplugindefs.h"
+#include <cfloat>
```
rdiankov, line 15: @undisputed-seraphim why is cfloat necessary?

undisputed-seraphim: I'm using gcc. I was trying to build OpenRAVE in native Debian 11 (without Docker).

undisputed-seraphim: @rdiankov Compiling with the cmake flags above results in an error from gcc stating that DBL_MAX was not defined anywhere.

cielavenir: let me check it thanks

rdiankov:
got it, thanks

plugins/rplanners/parabolicsmoother2.cpp
cielavenir: There is an include chain:

- plugins/include/openraveplugindefs.h
- include/openrave/openrave.h
- include/openrave/collisionchecker.h
- include/openrave/openravejson.h
- /usr/include/boost/lexical_cast.hpp
- /usr/include/boost/math/special_functions/sign.hpp
- /usr/include/boost/math/tools/config.hpp
- cfloat
- float.h

So it depends on Boost version.

@undisputed-seraphim For reference, may I ask what boost version you used? Note that Debian 11's stock is 1.74.

cielavenir: I mean, DBL_MAX issue does not reproduce by using Boost 1.74.

cielavenir: > So it depends on Boost version.

and adding cfloat include is correct, I approve. I clarify that

cielavenir: He used Debian 11 stock, have no idea :joy:

PR #1256, opened 2023-06-15
clear kinematics geometry hash when connected body active state changes

kanbouchou:
Background
==============

- https://github.com/rdiankov/openrave/pull/1240 minimized clearing of kinematics geometry hash, and https://github.com/rdiankov/openrave/pull/1241 fixed missing clearing of kinematics geometry hash.
- A few more locations were needed to clear this cache, so this MR fixes it.

Note
=========
- `__hashKinematicsGeometryDynamics` was private, but now protected because ``RobotBase`` class need to access it.
- Bumping version from 116 to 118 because https://github.com/rdiankov/openrave/pull/1251 updated changelog to 117 but probably forgot to change version in cmake.


include/openrave/kinbody.h
```diff
@@ -3561,9 +3561,9 @@ class OPENRAVE_API KinBody : public InterfaceBase
     boost::shared_ptr<rapidjson::Document> _prAssociatedFileEntries; ///< files tag maintaining entries of data files associated with this object
     Transform _baseLinkInBodyTransform; ///< the transform of the base link in the body coordinate frame. The body transform returned is baselink->GetTransform() * _baseLinkInBodyTransform.inverse(). When setting a transform, the base link transform becomes body->GetTransform() * _baseLinkInBodyTransform
     Transform _invBaseLinkInBodyTransform; ///< _baseLinkInBodyTransform.inverse() for speedup
+    mutable std::string __hashKinematicsGeometryDynamics; ///< hash serializing kinematics, dynamics and geometry properties of the KinBody
```
rdiankov, line 3564: @kanbouchou why mutable? doesn't look like you are changing it in a const function

kanbouchou: Actually, ``OpenRAVE::KinBody::GetKinematicsGeometryHash() const`` could modify this, so I think ``mutable`` is still needed.

rdiankov:
thanks~

PR #1259, opened 2023-06-17
fix: ConfigurationJitter should not permanently change robot's activeDOFIndices

osbertngok:
The side effect of changing robot's active dof indices seem to be unintended in `UpdateLimits` function in ConfigurationJitter / WorkspaceConfigurationJitter, and might lead to some unexpected consequences such as `Robot::CheckSelfCollision() const` might accidentally change active dof indices, should there be configuration jitterer having registered callbacks on robot changes.

Changes proposed by @yoshikikanemoto / @Puttichai


rdiankov:
already merged

PR #1262, opened 2023-06-21
GPG-based encryption

undisputed-seraphim:
# This pull request

This pull request seeks to merge the ability to _read_ GPG-encrypted files.

There are some preamble commits.
First, the functions used to query file extensions are simplified. `_Is***URI()` functions are decoupled from also querying file extensions, so that fewer slow regex operations are used in the initial file-or-URI loading process.
We test if a string is a URI only once; subsequently we only need to check the extensions. I have also streamlined some of the logic used in `ReadInterfaceURI()`.
I have also added a `stringutils.h` to deduplicate the multiple suffix matching functions, then added prefix matching for symmetry.

The new feature incurs a dependency on `libgpgme` and `libgpgmepp`. These are the [official C and C++ interfaces to GPG](https://gnupg.org/software/gpgme/index.html). The implementation simply adds an additional matching test for `.gpg` and `.pgp` file extensions and calls `_ParseEncryptedDocument()` in `jsondownloader.cpp`, which is where the meat of the implementation is.

To leverage this feature, simply install the correct secret key into the OS. If a matching key to decrypt a document exists, then it is automatically used; notice that the code does not need to do any key-finding, it simply loads the data into a structure and then calls `gpgCtx->decrypt()`. The resulting plaintext document is then stored back into the same context buffer, and then recurses into `_ParseDocument()`.

# Further work
This implementation only supports reading. Write-and-encrypting documents will be a subsequent pull request.


cielavenir:
@undisputed-seraphim May I have a very simple test case (or )? Also, I remember that gpgme requires gpg-agent running in background; Is gpg-agent started automatically?

cielavenir:
@undisputed-seraphim Seems I need the private key for dev/webstack/-/blob/encryptEnvironmentForTesting20230620/gpgpubkey.asc (if I need to test this)

src/libopenrave-core/stringutils.h
```diff
@@ -0,0 +1,95 @@
...
+#if  __cplusplus >= 201703L
+#define OPENRAVE_NODISCARD [[nodiscard]]
+#else
+#define OPENRAVE_NODISCARD __attribute__((warn_unused_result))
+#endif
+
+namespace OpenRAVE {
+
+// Returns true if a string begins with a matching prefix.
+OPENRAVE_NODISCARD inline bool StringStartsWith(const std::string& input, const std::string& prefix, bool ignoreCase = true) {
```
rdiankov, line 32: @undisputed-seraphim these are not very efficient implementations, so can you just put the header file close to the new code you write that will use them? or you can make them more efficient by reducing memory allocation (ie creating new std::strings)

undisputed-seraphim: As discussed I will change it to `StringStartsWith(const char* inputStr, size_t inputSz, const char* prefixStr, size_t prefixSz, bool ignoreCase)`.

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -1340,15 +1283,65 @@ bool RaveParseMsgPackData(EnvironmentBasePtr penv, KinBodyPtr& ppbody, const std
...
+    if (reader.IsDownloadingFromRemote()) {
+        reader.OpenRemoteDocument(uri, rEnvInfo);
+    } else
+#endif
+    {
+        std::string fullFilename = ResolveURI(uri, std::string(), reader.GetOpenRAVESchemeAliases());
+        if (fullFilename.size() == 0 ) {
+            return false;
+        }
+        OpenRapidJsonDocument(fullFilename, rEnvInfo);
```
cielavenir, line 1311: I'm afraid but using OpenRapidJsonDocument in RaveParseEncryptedURI is wrong.

src/libopenrave-core/jsonparser/jsondownloader.cpp
```diff
@@ -49,21 +72,59 @@ static void _ParseURI(const std::string& uri, std::string& scheme, std::string&
...
-{
-    try {
-        MsgPack::ParseMsgPack(doc, buffer.GetString(), buffer.GetSize());
+// Need to forward declare this
+static void _ParseDocument(OpenRAVE::JSONDownloadContextPtr pContext);
+
+static bool _ParseEncryptedDocument(OpenRAVE::JSONDownloadContextPtr pContext) {
+#if OPENRAVE_ENCRYPTION
+    // gpgme_check_version must be executed at least once before we create a context.
+    RAVELOG_DEBUG("GPG version: %s\n", gpgme_check_version(NULL));
```
cielavenir, line 81: RAVELOG_DEBUG shortcuts and we have to ensure gpgme_check_version.

```suggestion
    {
        // gpgme_check_version must be executed at least once before we create a context.
        const char *gpgmeVersion = gpgme_check_version(NULL);
        RAVELOG_DEBUG("GPG version: %s\n", gpgmeVersion);
    }
```

cielavenir:
Test procedure:

- machine 1

1. gpg --import < private.key
1. gpg --encrypt --recipient 0123456789ABCDEF < motoman-gp7.mujin.json > motoman-gp7.mujin.json.gpg
2. copy motoman-gp7.mujin.json.gpg and private.key

- machine 2

1. gpg --import < private.key
2. killall gpg-agent  # to ensure that gpgme wakes gpg-agent up automatically
3. `python -m openravepy -i --viewer= -p "RaveSetDebugLevel(100); env.LoadURI('mujin:/motoman-gp7.mujin.json.gpg',{'openravescheme':'mujin','remoteurl':'file:///home'})"`

It was loaded correctly :tada:

Issues:

1. RaveSetDebugLevel needs to be set, cf https://github.com/rdiankov/openrave/pull/1262#discussion_r1238116412
2. Without tini, gpg zombie processes are created.
3. env.Load() does not work. For testing I used the fact that curl can download file:/// URI, but then it is incompatible with how the scene is loaded in the controller. Actually what is the intended use? If the gpg file is used inside the controller, we would need two key pairs, one for env and one for downloaddebug, because env one's private key needs to be installed to the controller. Also, we need some protection mechanism or unencrypted files can be downloaded easily.

src/libopenrave-core/jsonparser/jsondownloader.cpp
undisputed-seraphim: I will note this fix, but I will do it in a different way as I am preparing some new code.

undisputed-seraphim:
In `jsoncommon.h`, most of the `RaveParse{JSON|MsgPack}{URI|File|Data}` and `RaveWritee{JSON|MsgPack}{File|Stream|Memory}` are highly redundant, and I think we should see if those code can be deduplicated.

src/libopenrave-core/jsonparser/jsonreader.cpp
undisputed-seraphim: This is now implemented

cielavenir:
> Also, we need some protection mechanism or unencrypted files can be downloaded easily.

I will check this part with @woswos

cielavenir:
@undisputed-seraphim is it possible to set the homedir? It seems possible by set_engine_info (or Context::setEngineHomeDirectory).

src/libopenrave-core/jsonparser/jsonwriter.cpp
```diff
@@ -374,23 +289,101 @@ void RaveWriteMsgPackMemory(EnvironmentBasePtr penv, std::vector<char>& output,
...
+    vectorstream vs(std::ios::binary);
+    RaveWriteEncryptedStream(listbodies, vs, atts, alloc, mimeType);
+    vs.swap_vector(output);
+}
+
+void RaveWriteEncryptedStream(EnvironmentBasePtr penv, std::ostream& os, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc, MimeType mimeType)
+{
+    std::string keyName;
+    for (const std::pair<std::string, std::string>& attribute : atts) {
+        if (attribute.first == "openravekey") {
```
ziyan, line 333: Can we name it `gpgkeyid`? perhaps also have `gpghomedir`?

ziyan: Actually, these key ids should be repeatable? Since we should be able to encrypt for multiple recipients.

rdiankov:
@ziyan is this ready?

superfashi:
> @ziyan is this ready?

I will look at the code and have this ready.

src/libopenrave-core/jsonparser/gpgutils.cpp
```diff
@@ -0,0 +1,215 @@
+// -*- coding: utf-8 -*-
+// Copyright (C) 2023 Tan Li Boon <liboon.tan@mujin.co.jp>
```
cielavenir, line 2:
```suggestion
// Copyright (C) 2023 Tan Li Boon <undisputed.seraphim@gmail.com>
```

ridiculously minor but which should be better?

src/libopenrave-core/jsonparser/jsonwriter.cpp
superfashi: done

superfashi:
Should be cleaned up @cielavenir do you want to test again?

src/libopenrave-core/jsonparser/jsonwriter.cpp
cielavenir: How about gpghomedir?

superfashi: Isn't that by environment variable GNUPGHOME as you have tested?

cielavenir: maybe let's avoid requiring GNUPGHOME envvar

superfashi: It's slightly hard to inject the options into `GpgDecrypt` as it is used inside `JSONDownloaderScope` which doesn't really accept a list of options/attributes.

cielavenir: @ziyan for gpghomedir, is it ok to rely on GNUPGHOME envvar instead? you asked for it, so we need your reply.

ziyan: It's best to have it instead of caller setting environment vars before calling openrave. Imagine your `penv->Load` code calling `setenv` first in your program.

superfashi: I know the design-wise it's good to have an option, but currently I don't have a good idea on how to inject the attributes into `JSONDownloaderScope` to be used for gpg decryption functions

ziyan:
@superfashi Please add unit test to `testopenrave`

src/libopenrave-core/environment-core.h
```diff
@@ -830,12 +846,7 @@ class Environment : public EnvironmentBase
         }
         }
 
-        if( listbodies.size() == 1 ) {
```
ziyan, line 833: Why is this check removed?

src/libopenrave-core/environment-core.h
```diff
@@ -4179,91 +4157,51 @@ class Environment : public EnvironmentBase
...
     }
 
     static bool _IsMsgPackData(const std::string& data)
     {
         return data.size() > 0 && !std::isprint(data[0]);
     }
 
+    static bool _IsEncryptedFile(const std::string& filename)
+    {
+        return StringEndsWith(filename, ".pgp") || StringEndsWith(filename, ".gpg");
```
ziyan, line 4196: Looks like the code above says we only support `.gpg`, so let's remove `.pgp` here

src/libopenrave-core/jsonparser/jsonwriter.cpp
ziyan: Ok, I see.

Just curious, how would you set home dir with gpgme?

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -120,21 +112,17 @@ static void OpenRapidJsonDocument(const std::string& filename, rapidjson::Docume
     }
 }
 
-/// \brief get the scheme of the uri, e.g. file: or openrave:
-static void ParseURI(const std::string& uri, std::string& scheme, std::string& path, std::string& fragment)
+/// \brief open and cache an encrypted document
+static void OpenEncryptedDocument(const std::string& filename, rapidjson::Document& doc)
```
ziyan, line 116: I don't get this logic here. If we open gpg file, how do we know it's json inside? There is no context about json vs msgpack here.

src/libopenrave-core/environment-core.h
superfashi: Liboon made the change to get rid of a single body condition. I think if you follow the code and get to `JsonWriter::Write` it is the same.

https://github.com/rdiankov/openrave/blob/b6c8feaacd4216f3de180c755d4acfb10b0b50b1/src/libopenrave-core/jsonparser/jsonwriter.cpp#L57-L65

src/libopenrave-core/jsonparser/jsonwriter.cpp
cielavenir: GNUPGHOME envvar (as we discussed)

https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration.html

src/libopenrave-core/jsonparser/jsoncommon.h
```diff
@@ -48,30 +53,39 @@ bool RaveParseMsgPackFile(EnvironmentBasePtr penv, RobotBasePtr& pprobot, const
...
-void RaveWriteMsgPackStream(const std::list<KinBodyPtr>& listbodies, const ostream& os, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+void RaveWriteMsgPackStream(EnvironmentBasePtr penv, std::ostream& os, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+void RaveWriteMsgPackStream(const std::list<KinBodyPtr>& listbodies, std::ostream& os, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
 void RaveWriteMsgPackMemory(EnvironmentBasePtr penv, std::vector<char>& output, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
-void RaveWriteMsgPackMemory(KinBodyPtr pbody, std::vector<char>& output, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
 void RaveWriteMsgPackMemory(const std::list<KinBodyPtr>& listbodies, std::vector<char>& output, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
 
+void RaveWriteEncryptedFile(EnvironmentBasePtr penv, const std::string& filename, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc, MimeType);
```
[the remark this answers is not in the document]
superfashi: I think there is just no need to copy & paste the same function everywhere. In fact I think we could refactor all other functions to use the same MimeType as well, given they are largely the same.

src/libopenrave-core/jsonparser/jsonwriter.cpp
superfashi: Should be `GpgME::Context::setEngineHomeDirectory`

src/libopenrave-core/jsonparser/jsonreader.cpp
superfashi: Yeah missed that. Added logic for using filename to detect.

src/libopenrave-core/jsonparser/jsoncommon.h
```diff
@@ -48,30 +53,39 @@ bool RaveParseMsgPackFile(EnvironmentBasePtr penv, RobotBasePtr& pprobot, const
 bool RaveParseMsgPackData(EnvironmentBasePtr penv, const std::string& data, UpdateFromInfoMode updateMode, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
 bool RaveParseMsgPackData(EnvironmentBasePtr penv, KinBodyPtr& ppbody, const std::string& data, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
 bool RaveParseMsgPackData(EnvironmentBasePtr penv, RobotBasePtr& pprobot, const std::string& data, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+bool RaveParseEncryptedFile(EnvironmentBasePtr penv, const std::string& filename, UpdateFromInfoMode updateMode, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
```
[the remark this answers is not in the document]
superfashi: See `OpenEncryptedDocument`

src/libopenrave-core/jsonparser/jsonwriter.cpp
cielavenir: @superfashi I'm confused. You said that passing the GpgME::Context::setEngineHomeDirectory argument from env.Load() is difficult.

superfashi: yeah it is difficult, unless you want to inject the logic into Load

src/libopenrave-core/jsonparser/jsoncommon.h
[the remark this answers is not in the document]
ziyan: But that's breaking the current convention. It's like adding dae support in jsonreader based on filename extensions

src/libopenrave-core/jsonparser/jsoncommon.h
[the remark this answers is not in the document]
ziyan: This is also breaking the convention.

src/libopenrave-core/jsonparser/jsoncommon.h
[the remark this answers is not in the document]
superfashi: We are always using extensions to detect file types (dae as fallback). I don't see how that breaks the convention here.  The data was processed before jsonreader.

src/libopenrave-core/jsonparser/jsoncommon.h
```diff
@@ -48,30 +49,46 @@ bool RaveParseMsgPackFile(EnvironmentBasePtr penv, RobotBasePtr& pprobot, const
 bool RaveParseMsgPackData(EnvironmentBasePtr penv, const std::string& data, UpdateFromInfoMode updateMode, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
 bool RaveParseMsgPackData(EnvironmentBasePtr penv, KinBodyPtr& ppbody, const std::string& data, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
 bool RaveParseMsgPackData(EnvironmentBasePtr penv, RobotBasePtr& pprobot, const std::string& data, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
```
ziyan, line 51: Where are `RaveParseEncryptedMsgPackData` and `RaveParseEncryptedJSONData`?

src/libopenrave-core/jsonparser/jsoncommon.h
```diff
@@ -48,30 +49,46 @@ bool RaveParseMsgPackFile(EnvironmentBasePtr penv, RobotBasePtr& pprobot, const
...
+
+void RaveWriteEncryptedMsgPackFile(EnvironmentBasePtr penv, const std::string& filename, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+void RaveWriteEncryptedMsgPackFile(const std::list<KinBodyPtr>& listbodies, const std::string& filename, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+void RaveWriteEncryptedMsgPackMemory(EnvironmentBasePtr penv, std::vector<char>& output, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+void RaveWriteEncryptedMsgPackMemory(const std::list<KinBodyPtr>& listbodies, std::vector<char>& output, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+void RaveWriteEncryptedMsgPackStream(EnvironmentBasePtr penv, std::ostream& os, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+void RaveWriteEncryptedMsgPackStream(const std::list<KinBodyPtr>& listbodies, std::ostream& os, const AttributesList& atts, rapidjson::Document::AllocatorType& alloc);
+
+bool GpgDecrypt(std::istream& inputStream, std::ostream& outputData);
```
ziyan, line 90: Do you need to expose these here?

src/libopenrave-core/jsonparser/jsoncommon.h
superfashi: Added

rdiankov:
is this ready?

rdiankov:
Anyone actively working on this?

ziyan:
> Anyone actively working on this?

@superfashi Please resolve conflicts and I will take another pass.

superfashi:
> > Anyone actively working on this?
> 
> 
> 
> @superfashi Please resolve conflicts and I will take another pass.

I will merge master by the end of day. Very sorry for the delayed response as I'm currently out of office.

src/libopenrave-core/jsonparser/jsoncommon.h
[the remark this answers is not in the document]
ziyan: This needs to be addressed still.

src/libopenrave-core/environment-core.h
```diff
@@ -586,17 +585,26 @@ class Environment : public EnvironmentBase
...
             return RaveParseJSONURI(shared_from_this(), uri, UFIM_Exact, atts, *_prLoadEnvAlloc);
         }
-        else if ( _IsMsgPackURI(uri) ) {
+        else if (_IsMsgPackFile(path)) {
             _ClearRapidJsonBuffer();
             return RaveParseMsgPackURI(shared_from_this(), uri, UFIM_Exact, atts, *_prLoadEnvAlloc);
         }
+        else if (_IsEncryptedFile(path)) {
+            _ClearRapidJsonBuffer();
+            return RaveParseEncryptedURI(shared_from_this(), uri, UFIM_Exact, atts, *_prLoadEnvAlloc);
```
ziyan, line 606: `RaveParseEncryptedMsgPackURI` and `RaveParseEncrypteJSONURI`

src/libopenrave-core/environment-core.h
```diff
@@ -631,6 +642,12 @@ class Environment : public EnvironmentBase
                 return true;
             }
         }
+        else if (_IsEncryptedFile(path)) {
+            _ClearRapidJsonBuffer();
+            if( RaveParseEncryptedFile(shared_from_this(), filename, UFIM_Exact, atts, *_prLoadEnvAlloc) ) {
```
ziyan, line 647: `RaveParseEncryptedMsgPackFile` and ``RaveParseEncryptedJSONFile`

src/libopenrave-core/jsonparser/jsonreader.cpp
ziyan: The point is: according to the design right now, I don't want to allow json parsing to switch to msgpack parsing in the middle of some reference chain. If top level is json, it has to be json all the way. If top leve is msgpack, it has to be msgpack all the way.

superfashi:
> is this ready?

It is now ready as reviewed by Ziyan, tested by me.

rdiankov:
thanks~

PR #1265, opened 2023-06-23
support std::vector<OpenRAVE::RaveVector<T>> serialization

eisoku9618:
need to move the definition of `SaveJsonValue` for `RaveVector` in front of the definition of `SaveJsonValue` for `std::vector` to support `std::vector<OpenRAVE::RaveVector<T>>` serialization.


yoshikikanemoto (approved):
looks good to me

rdiankov:
Thanks~

PR #1268, opened 2023-06-26
Support loading kinbody data with a references chain of 3 layers or more

LLipter:
When using `ReadKinBodyData()` to load JSON data in OpenRAVE, the reference resolution system does not work for a chain of 3 layers or more.  It only updates the kinbody information from the bottom layer of referencing hierarchy. This pull request fixes it.

See [!547](https://git.mujin.co.jp/backend/backend/-/issues/547)


ziyan:
Pipeline ID: 571234

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -708,47 +708,47 @@ class JSONReader
...
             bool isNewRobot = orjson::GetJsonValueByKey<bool>(rRefKinBodyInfo, "isRobot", false);
             if( isNewRobot ) {
                 pNewKinBodyInfo.reset(new RobotBase::RobotBaseInfo());
             }
             else {
                 pNewKinBodyInfo.reset(new KinBody::KinBodyInfo());
             }
-            pNewKinBodyInfo->DeserializeJSON(rRefKinBodyInfo, fRefUnitScale, _deserializeOptions);
+        }
+        pNewKinBodyInfo->DeserializeJSON(rRefKinBodyInfo, fRefUnitScale, _deserializeOptions);
```
rdiankov, line 724: there are code paths above that do:
```
envInfo._vBodyInfos.at(insertIndex)->DeserializeJSON(rRefKinBodyInfo, fUnitScale, _deserializeOptions);
```

and 
```
pExistingBodyInfo->DeserializeJSON(rRefKinBodyInfo, fRefUnitScale, _deserializeOptions);
```

So why aren't those code paths being hit? Potentially your change can lead to double deserialization.

ziyan: It's because `insertIndex` was kept at `-1` when the uri is local `#something`, since we didn't insert anything into `envInfo._vBodyInfos`

Because of `insertIndex = -1`, even though we found the `rRefKinBodyInfo`, we didn't `DeserializeJSON` with it.

LLipter:
```
envInfo._vBodyInfos.at(insertIndex)->DeserializeJSON(rRefKinBodyInfo, fUnitScale, _deserializeOptions);
```
will not be hit because it's under this branch `else if (!scheme.empty() && !path.empty()) {...}`, and the uri just has `#somebodyID`

LLipter:
```
pExistingBodyInfo->DeserializeJSON(rRefKinBodyInfo, fRefUnitScale, _deserializeOptions);
```
will not be hit because `insertIndex` is 0 at this time.

rdiankov: @ziyan my issue is when `insertIndex >= 0` and hits that code path, then there is a possibility of it deserializing twice.  When `insertIndex=-1`, there's no change in behavior.
So why when `insertIndex=0` did we not hit the above code paths that deseriaize?

ziyan: @rdiankov @LLipter see my latest commit

LLipter: Pipeline #571680 testing

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -697,7 +697,6 @@ class JSONReader
 
             const bool isPartial = orjson::GetJsonValueByKey<bool>(rRefKinBodyInfo, "__isPartial__", true);
             RAVELOG_VERBOSE_FORMAT("Deserializing rRefKinBodyInfo=%s into insertIndex=%d, isPartial=%d", orjson::DumpJson(rRefKinBodyInfo)%insertIndex%isPartial);
-            pNewKinBodyInfo->Reset();
```
LLipter: mhshp 6/7 and 7/7 failed. 

when we have body1 -> body2 -> body3, this line will wipe out data in body3, potentially making kinbody incomplete.

Pipeline #571837

ziyan: @LLipter can you try again with my latest change?

rdiankov:
please do careful testing before we declare this ready

rdiankov:
thanks, let's try it~

PR #1269, opened 2023-06-26
Add excludeBodyId option to JSON reader

superfashi:
This PR adds support of `excludeBodyId` attributes to `Env::LoadJSON`, which indicates what body ids are excluded from loading into the environment.


rdiankov:
changelog and patch update please

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -310,6 +313,10 @@ class JSONReader
                     throw OPENRAVE_EXCEPTION_FORMAT("Has body with no name in file '%s', so cannot load the scene: %s", _filename%orjson::DumpJson(rTempKinBodyInfo), ORE_InvalidArguments);
                 }
             }
+            if ( _excludeBodyIds.count(pKinBodyInfo->_id) ) {
+                itBodyInfo = envInfo._vBodyInfos.erase(itBodyInfo);
```
ziyan, line 317: @superfashi does this work? what if the bodyinfo is used later by a reference?

superfashi: Looks to me that `_ProcessEnvInfoBodies` which uses `_ExpandRapidJSON` already resolves the `referenceUri`

superfashi:
> changelog and patch update please

Sorry about that. Updated.

src/libopenrave-core/jsonparser/jsonreader.cpp
rdiankov: use `std::find` instead of `count`. thanks

superfashi: Can we use `set::find`?

cielavenir: .find is better; std::find is O(N) for (unordered_)set

cielavenir: By the way I'm personally surprised about this comment as I thought .count was (kind of) Mujin standard, eg https://github.com/rdiankov/openrave/blob/production/src/libopenrave/kinbodyjoint.cpp#L2417

rdiankov: Well... i'm personally surprised that code made it into openrave

rdiankov:
thanks

PR #1270, opened 2023-06-27
cmake: Use find_package for msgpack

undisputed-seraphim:
I'd like to use a newer version of msgpack (6.0.0, latest as of writing). Recent versions of msgpack has dropped support for finding the library via pkgconfig, so we should change CMakeLists to find msgpack via find_package instead.

----

- pipelineid=364920
- Pipeline #572059


cielavenir (approved):

cielavenir:
just note: please merge openrave (this one) first then jhbuild. merging jhbuild single will cause build error.

cielavenir:
- pipelineid=364920
- Pipeline #572059

rdiankov:
thanks~

PR #1271, opened 2023-06-28
Replace Unit with UnitInfo object

LLipter:
This pull request replaces `unit` in the environment with `UnitInfo` object. It also updates all related python bindings.


include/openrave/units.h
```diff
@@ -37,11 +37,13 @@ enum LengthUnit : int8_t
     LU_Nanometer = -9,
     LU_Inch = 0x10,
     LU_Foot = 0x11,
+    LU_TenthMillimeter = 0x12,
```
rdiankov, line 40: i would rather call it "100Micrometer", and the value should be -4

rdiankov: why do we need this by the way?

python/bindings/include/openravepy/openravepy_environmentbase.h
```diff
@@ -292,9 +292,9 @@ class PyEnvironmentBase : public OPENRAVE_ENABLE_SHARED_FROM_THIS<PyEnvironmentB
     void SetUserData(object o);
     object GetUserData() const;
 
-    void SetUnit(std::string unitname, dReal unitmult);
+    py::dict GetUnitInfo() const;
```
rdiankov, line 295: You can't just kill the old `SetUnit`, too many places use it.

python/bindings/openravepy_int.cpp
```diff
@@ -3791,6 +3801,172 @@ Because race conditions can pop up when trying to lock the openrave environment
         env.attr("TriangulateOptions") = selectionoptions;
     }
 
+#ifdef USE_PYBIND11_PYTHON_BINDINGS
+    object lengthUnit = enum_<LengthUnit>(m, "LengthUnit" DOXY_ENUM(LengthUnit))
+#else
+    object lengthUnit = enum_<LengthUnit>("LengthUnit" DOXY_ENUM(LengthUnit))
+#endif
+                          .value("Meter", LU_Meter)
```
rdiankov, line 3809: please use GetLengthUnitString() for the strings rather than hardcoding them like this

python/bindings/include/openravepy/openravepy_environmentbase.h
rdiankov: You can always set it as deprecated so that c++ generates a compile-time warning.

src/libopenrave/environment.cpp
```diff
@@ -120,8 +121,8 @@ void EnvironmentBase::EnvironmentBaseInfo::DeserializeJSONWithMapping(const rapi
         orjson::LoadJsonValueByKey(rEnvInfo, "revision", _revision);
     }
 
-    if (rEnvInfo.HasMember("unit")) {
-        orjson::LoadJsonValueByKey(rEnvInfo, "unit", _unit);
+    if (rEnvInfo.HasMember("unitInfo")) {
```
rdiankov, line 124: You have to continue to support the old "unit" since there are too many environments out there that use it.

include/openrave/units.h
LLipter: We have a python version unit conversion function (`GetUnitConversionScale`), which supports `0.1mm`. Do we want to support it as well here?

python/bindings/openravepy_int.cpp
LLipter: If so, the unit enum name will be `openravepy.LengthUnit.m` when we use it in python. I feel like `openravepy.LengthUnit.Meter` is more readable.

python/bindings/include/openravepy/openravepy_environmentbase.h
LLipter: Sure. I add them back.

src/libopenrave/environment.cpp
LLipter: sure. I add it back

rdiankov:
@LLipter there was a point in UpdateFromInfo that was assigning _unit in production, which I just changed to check and throw an exception. can you do the same in this branch and test?
Thanks

LLipter:
@rdiankov Sure, I also update my MR on webstack to throw an error in this situation

rdiankov:
@LLipter did you actually test, this simple scene fails:
```
{
  "bodies":[{
    "name":"C",
    "links":[{
      "name":"L",
      "geometries":[{
        "name":"G",
        "type":"cylinder",
        "radius":1,
        "height":1
      }]
    }]
  }],
  "unitInfo": {
    "lengthUnit":"mm"
  }
}
```

rdiankov:
thanks~

PR #1272, opened 2023-06-28
Add a Last Modified member to KinBody

undisputed-seraphim:
Prerequisite: [msgpack_v6 branch needs to be merged](https://github.com/rdiankov/openrave/pull/1270), as it leverages a new extension in the msgpack spec.

Adds a _lastModified time stamp to the environment. Note that this is not updated even if something changed in the environment upon saving.


src/libopenrave/environment.cpp
```diff
@@ -116,6 +116,17 @@ void EnvironmentBase::EnvironmentBaseInfo::DeserializeJSONWithMapping(const rapi
     // for DeserializeJSON, there are two possibilities: 1. full json passed in 2. diff json passed in
     // for example, do not clear _vBodyInfos.clear(), since we could be dealing with partial json
 
+    if (rEnvInfo.HasMember("modifiedAt")) {
+        std::string lastModifiedStr;
+        orjson::LoadJsonValueByKey(rEnvInfo, "modifiedAt", lastModifiedStr);
+        std::tm ctm;
+        ::strptime(&lastModifiedStr[0], "%FT%T%z", &ctm);
```
rdiankov, line 123: this won't do microseconds, can you use this instead:
```
int64_t stampus = (boost::posix_time::from_iso_extended_string(pIsoFormatDateTime) - boost::posix_time::from_time_t(0)).total_microseconds();
```

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -35,6 +35,39 @@
 
 namespace OpenRAVE {
 
+static int64_t ConvertIsoFormatDateTimeToLinuxTimeUS(const char* pIsoFormatDateTime)
+{
+    // RFC 3339 Nano format (2006-01-02T15:04:05.999999999Z07:00)
+    struct tm datetime;
+    const char *remain = strptime(pIsoFormatDateTime, "%FT%T", &datetime);
```
rdiankov, line 42: please handle the case where `remain` is NULL.

```
RETURN VALUE
       The return value of the function is a pointer to the first character not processed in this function call.  In case the input string contains more  characters  than  required  by  the  format
       string, the return value points right after the last consumed input character.  In case the whole input string is consumed, the return value points to the null byte at the end of the string.
       If strptime() fails to match all of the format string and therefore an error occurred, the function returns NULL.
```

rdiankov: @superfashi also handle when `pIsoFormatDateTime=""` and `pIsoFormatDateTime=nullptr`, ie 0 should return

superfashi: done

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -35,6 +35,45 @@
...
+static int64_t ConvertIsoFormatDateTimeToLinuxTimeUS(const char* pIsoFormatDateTime)
+{
+    if (pIsoFormatDateTime == nullptr) {
+        return 0;
+    }
+    // RFC 3339 Nano format (2006-01-02T15:04:05.999999999Z07:00)
+    struct tm datetime;
+    const char *remain = strptime(pIsoFormatDateTime, "%FT%T", &datetime);
+    if (remain == nullptr) {
+        return 0;
```
rdiankov, line 47: that's not correct

superfashi: Not sure what you mean it’s not correct. If you prefer !! check I can change it.

rdiankov:
thanks~

PR #1274, opened 2023-07-03
Remove RemoveKinBody calls from Read functions

superfashi:
After review it doesn't seem to make sense to call `RemoveKinBody` in `Read[xxx]` functions. 

Tested and doesn't seem to affect production.


rdiankov:
@superfashi were you able to test that there's no regressions?

superfashi:
> @superfashi were you able to test that there's no regressions?

Yes, see pm.

rdiankov:
thanks~

PR #1273, opened 2023-07-01
Add optional `uri` argument for LoadJSON and add ReadRobotJSON and ReadKinBodyJSON

superfashi:
This PR adds optional `uri` argument for `LoadJSON`.

It also adds `ReadRobotJSON` and `ReadKinBodyJSON` for completion.

PR #1277, opened 2023-07-05
Remove leftover RemoveKinBody calls from Read functions

superfashi:
Changes in #1273 is not included in #1274. This PR removes the leftover ones.


rdiankov:
thanks

PR #1278, opened 2023-07-05
add reset function to IkParameterization

eisoku9618:
I would like to cache `IkParameterization` so I need to have a way to reset its instance.


yoshikikanemoto (approved):
looks good to me

rdiankov:
Will need to change minor version since this breaks binary compat

eisoku9618:
@rdiankov @yoshikikanemoto Thank you for checking. I incremented minor version and changelog.

eisoku9618:
By the way, I think this change is not a ABI breaking change as far as I read https://www.acodersjourney.com/20-abi-breaking-changes/

note to self: since our CMakeLists.txt is like this, libopenraveXXX.so.YYY does not include patch version.

```CMake
set (OPENRAVE_SOVERSION ${OPENRAVE_VERSION_MAJOR}.${OPENRAVE_VERSION_MINOR})
set(OPENRAVE_LIBRARY_SUFFIX "${OPENRAVE_SOVERSION}" CACHE STRING "Suffix to append to library names")

set_target_properties(libopenrave PROPERTIES OUTPUT_NAME openrave${OPENRAVE_LIBRARY_SUFFIX}
                                  SOVERSION 0 # always have it 0 since we're including the soversion as part of the library name
                                  VERSION ${OPENRAVE_VERSION}
                                  COMPILE_FLAGS "${LIBOPENRAVE_COMPILE_FLAGS} ${FPARSER_CXX_FLAGS} -DOPENRAVE_DLL_EXPORTS -DOPENRAVE_DLL"
                                  LINK_FLAGS "${LIBOPENRAVE_LINK_FLAGS} ${FPARSER_LINK_FLAGS}")
```

image

rdiankov:
Yes it does. You added a function in a class, everything has to be recompiled because rd signature changed

rdiankov:
thanks

PR #1280, opened 2023-07-05
add backward compatibility for RaveOrientedBox deserialization

eisoku9618:
In the past, RaveOrientedBox was serialized with `halfExtents` / `pose` keywords.


rdiankov:
I just noticed this. If possible, let's serialize to halfExtents, that is more standard

cielavenir:
@eisoku9618 that scene was ok to load.

lazydroid (approved):
looks ok, thanks!

cielavenir:
@eisoku9618 sorry, this pull request is blocking testplanningcommon. Please talk with @yoshikikanemoto (and @kanbouchou) for serialization format.

yoshikikanemoto (approved):

yoshikikanemoto:
@rdiankov there was old scenes that use "pose" and "halfExtents" instead of "transform" and "extents". this change is just to support loading those old scenes.

cielavenir:
@yoshikikanemoto so we will use "transform" and "extents" as default. Thank you for clarification.

cielavenir (approved):

eisoku9618:
> If possible, let's serialize to halfExtents, that is more standard

Since many code use `extents` e.g. RaveOrientedBox / RaveAxisAlignedBox / PyOrientedBox / PyAABB so it takes some time to change all of them to `halfExtents`... :cry:

cielavenir:
I agree that the C++ variable name and the serialization key should be the same.

rdiankov:
Geometry already serializes to "halfExtents", so let's be consistent with the serialization, even if the variable names are not:
```c++
    case GT_Box:
        orjson::SetJsonValueByKey(rGeometryInfo, "halfExtents", _vGeomData*fUnitScale, allocator);
        break;
```

eisoku9618:
test failed. checking.

eisoku9618:
@rdiankov Could you review this MR again? We also need to merge one internal repository together with this change.

Pipeline #579072

rdiankov:
thanks~

PR #1281, opened 2023-07-06
Add conical frustum geometry (extending current cylinder)

superfashi:
This MR extends the geometry of a cylinder to support a different top and bottom radius (conical frustum).

Tested locally for viewer and collision.


src/libopenrave/kinbodygeometry.cpp
```diff
@@ -165,26 +162,52 @@ void AppendBoxTriangulation(const Vector& pos, const Vector& ex, TriMesh& tri)
...
+        const dReal unitY = RaveSin(dTheta * i);
+        const int32_t off = tri.vertices.size();
+
+        // line on the side
+        tri.vertices.emplace_back(pos.x + unitX * topRad, pos.y + unitY * topRad, pos.z + halfHeight); // line top
+        tri.vertices.emplace_back(pos.x + unitX * botRad, pos.y + unitY * botRad, pos.z - halfHeight); // line bottom
+
+        if (i > 0) {
+            // four triangles
+            tri.indices.insert(tri.indices.end(), {
```
rdiankov, line 185: in interests of not implicitly allocating vectors, just so that we can use `insert`. please restore the old `push_back` methods. you can put your comments there. thanks

rdiankov: in fact, unless, there's a bug, please don't make big changes to the implementation. it only increases review time. thanks

superfashi: this is not allocating vectors, it’s using initializer_list which is compile-time behavior.

superfashi: There is indeed an off-by-one bug in the original implementation. The first side edge is created twice (again at the end) causing this mesh not closing.

I did not change the implementation but merely refactored it (and added documentation).

rdiankov: this is about defining the triangle mesh, and there are no repeating triangles. also i don't see any  "mesh not closing", you'll have to show it to me.
let's use `push_back`, you don't have to worry about saving code space.
thanks,

rdiankov: Actually using indices is fastest (not push_back). Just timed both version,  your version is 2x+ slower. 
```
using indices: 6644[us]
using std::vector::insert: 14917[us]
```

You can see the perf script here:

https://pastebin.com/J5a8yq14


Can use this code:

```c++
void AppendCylinderTriangulationIndex(const Vector& pos, const dReal topRadius, const dReal bottomRadius, const dReal len, const int numverts, TriMesh& tri)
{
    // once again, cylinder is on z axis
    dReal dtheta = 2 * PI / (dReal)numverts;
    int base = tri.vertices.size();
    tri.vertices.resize(2*(numverts+1) + 4);
    tri.indices.resize((numverts+1) * 12);

    int vindex = 0;
    int iindex = 0;
    tri.vertices[vindex++] = Vector(0,0,len) + pos;
    tri.vertices[vindex++] = Vector(0,0,-len) + pos;
    tri.vertices[vindex++] = Vector(topRadius,0,len) + pos;
    tri.vertices[vindex++] = Vector(bottomRadius,0,-len) + pos;
    for(int i = 0; i < numverts+1; ++i) {
        dReal s = sinf(dtheta * (dReal)i);
        dReal c = cosf(dtheta * (dReal)i);
        int off = (int)tri.vertices.size();
        tri.vertices[vindex++] = Vector(topRadius*c, topRadius*s, len) + pos;
        tri.vertices[vindex++] = Vector(bottomRadius*c, bottomRadius*s, -len) + pos;

        tri.indices[iindex++] = (base);       tri.indices[iindex++] = (off-2);       tri.indices[iindex++] = (off);
        tri.indices[iindex++] = (base+1);       tri.indices[iindex++] = (off+1);       tri.indices[iindex++] = (off-1);
        tri.indices[iindex++] = (off-2);   tri.indices[iindex++] = (off-1);       tri.indices[iindex++] = (off);
        tri.indices[iindex++] = (off);     tri.indices[iindex++] = (off-1);       tri.indices[iindex++] = (off+1);
    }
}
```

superfashi: Reason it's faster because the `[]` operator skips bound checks. If you change it to `.at` it goes back to the same level as `.insert` or `.push_back`. Although no worry of the segfault here if the size is calculated correctly.

Also the code is incorrect as `.size()` now reports the full size, have to do more math there to calculate the vertices index, making the code harder to read and maintain.

Your decision though.

superfashi: Refactored into a even faster version.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -165,26 +162,54 @@ void AppendBoxTriangulation(const Vector& pos, const Vector& ex, TriMesh& tri)
     tri.indices.insert(tri.indices.end(), &indices[0], &indices[nindices]);
 }
 
-void AppendCylinderTriangulation(const Vector& pos, const dReal& rad, const dReal& len, const int& numverts, TriMesh& tri) {
+static void AppendConicalFrustumTriangulation(const Vector& pos, const dReal topRad, const dReal botRad, const dReal halfHeight, const uint numFaces, TriMesh& tri)
```
rdiankov, line 165: still slower than when using index
```
rdiankov@PF3VMBC7:~ $ ./a.out 
using indices: 66060[us]
using std::vector::iterator: 80034[us]
rdiankov@PF3VMBC7:~ $ ./a.out 
using indices: 68879[us]
using std::vector::iterator: 84180[us]
rdiankov@PF3VMBC7:~ $ ./a.out 
using indices: 65620[us]
using std::vector::iterator: 79354[us]
rdiankov@PF3VMBC7:~ $ ./a.out 
using indices: 65254[us]
using std::vector::iterator: 78898[us]
rdiankov@PF3VMBC7:~ $ ./a.out 
using indices: 65312[us]
using std::vector::iterator: 79268[us]
rdiankov@PF3VMBC7:~ $ ./a.out 
using indices: 69626[us]
using std::vector::iterator: 82956[us]
rdiankov@PF3VMBC7:~ $ ./a.out 
using indices: 65528[us]
using std::vector::iterator: 78486[us]
```

rdiankov: Looks like the problem with the original algorithm's duplicate vertices is that it started at `i=0`. If you set `i=1`, then there's no duplication:
```c++
void AppendCylinderTriangulationIndex(const Vector& pos, const dReal topRadius, const dReal bottomRadius, const dReal len, const int numverts, TriMesh& tri)
{
    // once again, cylinder is on z axis
    dReal dtheta = 2 * PI / (dReal)numverts;
    int base = tri.vertices.size();
    tri.vertices.resize(2*(numverts+1) + 4);
    tri.indices.resize((numverts+1) * 12);

    int vindex = 0;
    int iindex = 0;
    tri.vertices[vindex++] = Vector(0,0,len) + pos;
    tri.vertices[vindex++] = Vector(0,0,-len) + pos;
    tri.vertices[vindex++] = Vector(topRadius,0,len) + pos;
    tri.vertices[vindex++] = Vector(bottomRadius,0,-len) + pos;
    for(int i = 1; i < numverts+1; ++i) {
        dReal s = sinf(dtheta * (dReal)i);
        dReal c = cosf(dtheta * (dReal)i);
        int off = (int)tri.vertices.size();
        tri.vertices[vindex++] = Vector(topRadius*c, topRadius*s, len) + pos;
        tri.vertices[vindex++] = Vector(bottomRadius*c, bottomRadius*s, -len) + pos;

        tri.indices[iindex++] = (base);       tri.indices[iindex++] = (off-2);       tri.indices[iindex++] = (off);
        tri.indices[iindex++] = (base+1);       tri.indices[iindex++] = (off+1);       tri.indices[iindex++] = (off-1);
        tri.indices[iindex++] = (off-2);   tri.indices[iindex++] = (off-1);       tri.indices[iindex++] = (off);
        tri.indices[iindex++] = (off);     tri.indices[iindex++] = (off-1);       tri.indices[iindex++] = (off+1);
    }
}
```

rdiankov: fixed indexing issue

superfashi:
Reviewed in person

rdiankov:
thanks~

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -1468,6 +1510,14 @@ uint8_t KinBody::GeometryInfo::GetSideWallExists() const
     return mask;
 }
 
+dReal KinBody::GeometryInfo::GetCylinderRadius() const
+{
+    if (_vGeomData.z == 0 || _vGeomData.x != _vGeomData.z) {
```
Puttichai, line 1515: @superfashi Shouldn't this condition be `_vGeomData.z != 0 && _vGeomData.x != _vGeomData.z` instead?

superfashi: You are right. Can you send a PR? I am currently on something else. Thank you.

PR #1283, opened 2023-07-06
Fix openravejson 20230706

kanbouchou:
``LoadJsonValue(rValue, std::vector<Transform>)`` does not compile because `template<class T> inline void LoadJsonValue(const rapidjson::Value& v, std::vector<T>& t)` does not see `template<class T> inline void LoadJsonValue(const rapidjson::Value& v, OpenRAVE::geometry::RaveTransform<T>& t)` so added forward declaration


rdiankov:
thanks

PR #1285, opened 2023-07-11
Fix condition in GetCylinderRadius warning message

superfashi:
Minor fix on condition in GetCylinderRadius warning message

reported https://github.com/rdiankov/openrave/pull/1281#discussion_r1259078224


rdiankov:
Thanks

PR #1286, opened 2023-07-11
Draft: Add GetCylinder[xxx] to GeometryInfo Python binding and add Cone backward compatibility

superfashi:
This PR adds `GetCylinderTopRadius` and `GetCylinderBottomRadius` to `GeometryInfo` Python binding.

Also adds support to Cone (when top or bottom radius is 0)


rdiankov:
There's a potential problem here (and with current production) of when user sets bottomRadius=0 because they want a cone. Due to back compat, our system will currently think that is a full cylinder.
We should change the place where we store top and bottom radius to `_vGeomData.z` and `_vGeomData.w` respectively so that if both `_vGeomData.z==0 && _vGeomData.w==0`, we know it is a regular cylinder and should just use `_vGeomData.x`. Otherwise, we would use topRadius=`_vGeomData.z` and bottomRadius=`_vGeomData.w`

superfashi:
> There's a potential problem here (and with current production) of when user sets bottomRadius=0 because they want a cone. Due to back compat, our system will currently think that is a full cylinder.
> 
> We should change the place where we store top and bottom radius to `_vGeomData.z` and `_vGeomData.w` respectively so that if both `_vGeomData.z==0 && _vGeomData.w==0`, we know it is a regular cylinder and should just use `_vGeomData.x`. Otherwise, we would use topRadius=`_vGeomData.z` and bottomRadius=`_vGeomData.w`

Good idea, I’ll implement this.

superfashi:
@rdiankov Added support for cone.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -162,8 +162,63 @@ void AppendBoxTriangulation(const Vector& pos, const Vector& ex, TriMesh& tri)
     tri.indices.insert(tri.indices.end(), &indices[0], &indices[nindices]);
 }
 
+static void AppendConeTriangulation(const Vector& pos, const dReal rad, dReal halfHeight, const uint numFaces, TriMesh& tri, const bool upsideDown = false)
```
rdiankov, line 165: @superfashi have you visualized this to confirm it creates triangles in the correct clockwise order? 
also, let's call it `radius`.

src/libopenrave/kinbodygeometry.cpp
```diff
@@ -1045,9 +1100,13 @@ void KinBody::GeometryInfo::SerializeJSON(rapidjson::Value& rGeometryInfo, rapid
         break;
 
     case GT_Cylinder:
-        orjson::SetJsonValueByKey(rGeometryInfo, "topRadius", GetCylinderTopRadius()*fUnitScale, allocator);
-        orjson::SetJsonValueByKey(rGeometryInfo, "height", GetCylinderHeight()*fUnitScale, allocator);
-        orjson::SetJsonValueByKey(rGeometryInfo, "bottomRadius", GetCylinderBottomRadius()*fUnitScale, allocator);
+        if (IsConicalFrustum()) {
```
rdiankov, line 1103: @superfashi in order to simplify implementation of tools using this format, I think we should standardize to output `topRadius` and `bottomRadius` always, even if they are the same value. And `radius` will not be output anymore, we will only support reading it for back compat purposes.

superfashi: 1. Need to fix internal test since now we break the (assumed) invariant of Deserialize -> Serialize to the same data.
2. How about collada serializer/deserializer? Do we add top/bottomRadius change there as well?

src/libopenrave/kinbodygeometry.cpp
superfashi: Yes I have tested visually. Name changed.

src/libopenrave/kinbodygeometry.cpp
superfashi: If 2 is not a consideration, I have pushed 1.

superfashi:
Closed by #1288

PR #1288, opened 2023-07-14
Add Conical Frustum geometry type (close #1286)

superfashi:
This PR reverts #1281 and adds a separate Conical Frustum (`conicalfrustum`) geometry type.

Tested locally and on CI and checked visually.


rdiankov:
thanks~

PR #1289, opened 2023-07-14
Add ReadableInterface support to Link and LinkInfo

superfashi:
This PR adds `ReadableInterface` support to `Link` and `LinkInfo`.

To achieve this, readable interface related methods are extracted from `InterfaceBase` into a separate class `ReadableInterfaceBase`, so that `Link` could use readable interfaces without being an openrave interface.


ziyan:
@superfashi @woswos Is this ready? What are the left over items?

include/openrave/interface.h
```diff
@@ -75,13 +75,67 @@ class OPENRAVE_API InfoBase
...
+    virtual void SetReadableInterfaces(const READERSMAP& mapReadables, bool bClearAllExisting);
+
+    /// \brief clears the readable interfaces
+    virtual void ClearReadableInterfaces();
+    virtual void ClearReadableInterface(const std::string& id);
+
+    /// \brief updates the readable interfaces. returns true if there are any changes
+    virtual bool UpdateReadableInterfaces(const std::map<std::string, ReadablePtr>& newReadableInterfaces);
+
+    boost::shared_mutex& GetReadableInterfaceMutex() const
```
ziyan, line 108: @superfashi You added this mutex, any reason why it wasn't needed before?

ziyan: Isn't locking the env itself enough?

src/libopenrave/kinbodylink.cpp
```diff
@@ -200,6 +201,21 @@ void KinBody::LinkInfo::SerializeJSON(rapidjson::Value &value, rapidjson::Docume
         value.AddMember("geometries", geometriesValue, allocator);
     }
 
+    if (!_mReadableInterfaces.empty()) {
+        rapidjson::Value rReadableInterfaces;
+        rReadableInterfaces.SetObject();
+        for (std::map<std::string, ReadablePtr>::const_iterator it = _mReadableInterfaces.begin(); it != _mReadableInterfaces.end(); it++) {
+            // skip serializing __collada__ since we won't need it in json
+            if (it->first == "__collada__") {
```
ziyan, line 209: This won't be there for link info.

src/libopenrave/kinbodylink.cpp
```diff
@@ -319,6 +335,16 @@ void KinBody::LinkInfo::DeserializeJSON(const rapidjson::Value &value, dReal fUn
         }
     }
 
+    if (value.HasMember("readableInterfaces") && value["readableInterfaces"].IsObject()) {
+        for (rapidjson::Value::ConstMemberIterator it = value["readableInterfaces"].MemberBegin(); it != value["readableInterfaces"].MemberEnd(); ++it) {
+            // skip over __collada__ since it will most likely fail to deserialize
```
ziyan, line 340: Won't be there for link info, we can remove.

include/openrave/interface.h
superfashi: It was there before inside `InterfaceBase`, but everything was using one mutex lock. I think locking the env is fine but is it better to have finer lock granularity?

ziyan: I see. There was a lock already.

ziyan (approved):

rdiankov:
also, let's do this for JointInfo too

PR #1290, opened 2023-07-25
Overwrite uri from atts

LLipter:
A recent [commit](https://github.com/rdiankov/openrave/commit/24e05b21716c436c20ba53a8138bc38145ac7f6f) breaks 
`test_robotbridgestate/test_NotifyRobotChanged[True]`. Not sure how to fix it correctly, but allowing overwriting URI from `atts` can fix the test.

### DIFF
--- a/src/libopenrave-core/environment-core.h
+++ b/src/libopenrave-core/environment-core.h
@@ -1797,6 +1797,14 @@ public:
                 robot->__struri = preader->_filename;
             }
         }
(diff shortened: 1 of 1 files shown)


rdiankov:
let's fix the real bug, we're not going to re-introduce this functionality.

rdiankov:
Looks like big was in the test and not in OpenRAVE

PR #1293, opened 2023-08-03
Make the unit standard value functions constexpr

shossjer:
This change requires c++14 to compile but the code can if needed be changed to support c++11 by writing something like
```
    return unit == OpenRAVE::LU_Meter ? 1.0 :
           unit == OpenRAVE::LU_Millimeter ? 1e3 :
           ...
           throw ...;
```

### DIFF
--- a/include/openrave/units.h
+++ b/include/openrave/units.h
@@ -156,7 +156,7 @@ inline T GetLengthUnitStandardValue(const std::string& s) {
 
 // \brief how many units in a meter
 template <typename T>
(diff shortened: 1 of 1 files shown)


rdiankov:
Thanks~

cielavenir:
don't worry, after Debian11 migration, our standard is C++14.

cielavenir:
for record: this was reverted by https://github.com/rdiankov/openrave/commit/3bc6152ed7a26925c0bfc34d5bf40ac751617925

PR #1294, opened 2023-08-07
cast GeometryType to int when serializing

cielavenir:
https://github.com/rdiankov/openrave/commit/d841c7633ec34cbd95c6486811eb2d5401b198ad#diff-f6b2317998343e31af6a1b2108eb6b453aeadb49c362a4fd3c24b6f36455cbeeR46 changed GeometryType. Usually it is ok as types are coerced, but in serialize(), now `_info._type` is considered as unsigned char. This caused unexpected GetKinematicsGeometryHash change. (I suppose that) this invalidated robot cache (robot.md5).

/cc @ntohge


rdiankov:
thanks

PR #1295, opened 2023-08-09
Fix crop margin visualization

seanzhang15:
This PR fixes the following issues with crop margin visualization:

1. Labels remain in the viewer. Fixed by properly removing them when setting crop margin to be invisible. 
2. Crop margin visualization is lower than expected in the z-direction. Fixed by translating the box by the correct value.


rdiankov:
Thanks

PR #1299, opened 2023-08-14
Fix crop margin visualization for cages

seanzhang15:
In https://github.com/rdiankov/openrave/pull/1295, the crop margin visualization is adjusted in the z direction using the geometry's outer extents. However, when the geometry is a cage, the same logic does not apply, and the visualization was still wrong. 

This PR fixes the issue by handling both cases explicitly.


rdiankov:
Thanks

PR #1300, opened 2023-08-17
expose _readableInterfaces for PyLinkInfo

superfashi:
`PyKinBodyInfo` is exposing `_readableInterfaces`, make it the same for `PyLinkInfo`.

@chh56


rdiankov:
thanks

PR #1304, opened 2023-08-27
Correct the typo: tan(self.tvar) -> tan(self.var) in the "subsinv" list

nothingstopsme:
A typo has been spotted in the member list "subsinv" of the class "Variable"

Not sure if that affected pair in the "subsinv" list is ever used for symbol substitution when solving equations, and I do not have any particular problems of which I can put my finger on that typo and say it is the cause; but I think it might be a good idea to have it corrected before it leads to any possible future issues.


cielavenir:
I have no knowledge on ikfast. @yoshikikanemoto could you check?

rdiankov:
thanks!

PR #1308, opened 2023-09-12
Fix logging on Python3

cielavenir:
Raw stream should be used for Py2.7+. Actually it raises error on Py3.

This is the simulation of commonlogging.py.

also fixes colorize and threading import.


cielavenir:
pipelineid=391379 Pipeline #626784

rdiankov:
thanks

PR #1309, opened 2023-09-13
export some of major openravepy symbols so that user can call python from c++ with openrave major classes

yoshikikanemoto:


eisoku9618 (approved):
looks good to me. need to bump version before merge so that no one will complain about link error.

rdiankov:
Thanks

PR #1311, opened 2023-09-20
Draft: Add manipulator/grippername to XML reader

Puttichai:
### Summary

Now XML reader can read  `grippername` specified under `manipulator`. This is so that a gripper name can be specified directly from an XML file.


yoshikikanemoto:
@Puttichai lgtm. changelog? patch version?

yoshikikanemoto:
@rdiankov can we merge this quickly? people still find OpenRAVE XML useful to model complicated robot model and this is for it

PR #1314, opened 2023-10-02
Optimizations for environments with many links / geometries

rschlaikjer:


include/openrave/kinbody.h
```diff
@@ -1310,7 +1317,8 @@ class OPENRAVE_API KinBody : public InterfaceBase
         ///
         /// the list is stored inside _GetInfo()._mapExtraGeometries. Note that the pointers are copied and not the data, so
         /// any be careful not to modify the geometries afterwards
-        void SetGroupGeometries(const std::string& name, const std::vector<KinBody::GeometryInfoPtr>& geometries);
+        /// \param deferPostprocessChangedParameters If true, this function will not post a Prop_LinkGeometryGroup update to the body. It will be up to the caller to ensure this occurs.
+        void SetGroupGeometries(const std::string& name, const std::vector<KinBody::GeometryInfoPtr>& geometries, bool deferPostprocessChangedParameters = false);
```
rdiankov, line 1321: seems you are only using the `deferPostprocessChangedParameters` for the internal computations, in which case i would create a different private function that skips the postprocessing, and leave the public function as is.

src/libopenrave/libopenrave.cpp
```diff
@@ -2292,9 +2292,6 @@ void TriMesh::Append(const TriMesh& mesh)
 {
     int offset = (int)vertices.size();
     vertices.insert(vertices.end(), mesh.vertices.begin(), mesh.vertices.end());
-    if( indices.capacity() < indices.size()+mesh.indices.size() ) {
```
ziyan, line 2295: @rschlaikjer why the pre-allocation is removed?

include/openrave/kinbody.h
```diff
@@ -1312,6 +1319,13 @@ class OPENRAVE_API KinBody : public InterfaceBase
         /// any be careful not to modify the geometries afterwards
         void SetGroupGeometries(const std::string& name, const std::vector<KinBody::GeometryInfoPtr>& geometries);
 
+private:
+        /// \brief stores geometries for later retrieval
+        ///
+        /// This call is identical to SetGroupGeometries except that it does not automatically post a Prop_LinkGeometryGroup update to the body. It will be up to the caller to ensure this occurs.
+        void SetGroupGeometriesNoPostprocess(const std::string& name, const std::vector<KinBody::GeometryInfoPtr>& geometries);
```
ziyan, line 1326: @rschlaikjer Please start private function name with `_`

plugins/fclrave/fclspace.cpp
```diff
@@ -569,9 +588,14 @@ void FCLSpace::_ResetGeometryGroupsCallback(boost::weak_ptr<FCLKinBodyInfo> _pin
...
         pinfo->nGeometryUpdateStamp++;
-        FCLKinBodyInfoRemover remover(boost::bind(&FCLSpace::RemoveUserData, this, pbody)); // protect
-        InitKinBody(pbody, pinfo, false);
-        remover.ResetRemove(); // succeeded
+        try {
+            ReloadKinBodyLinks(pbody, pinfo);
+        }
+        catch (...) {
+            // If something goes wrong processing this body, remove it from the FCL space.
+            RemoveUserData(pbody);
```
ziyan, line 596: @rschlaikjer Isn't the RAII remover much cleaner? Why change to this?

plugins/fclrave/fclspace.h
```diff
@@ -152,6 +152,7 @@ class FCLSpace : public boost::enable_shared_from_this<FCLSpace>
 
         KinBodyWeakPtr _pbody;
         int nLastStamp;  ///< KinBody::GetUpdateStamp() when last synchronized ("is transform up to date")
+        int nLastLinkReloadStamp; ///< KinBody::GetUpdateStamp() when we last ran ReloadKinBodyLinks
```
ziyan, line 155: I would just initialize all these `int` members in case of missed initialization

plugins/fclrave/fclspace.cpp
ziyan: If you don't want the bind cost, why not make a specialized remover class?

src/libopenrave/libopenrave.cpp
rschlaikjer: This kind of reserve is an antipattern, especially the way that it plays into the larger call scope here - IIRC this function is called repeatedly, and with this explicit reserve
- You force a realloc, but more importantly
- You force that realloc _to only be as large as the new items added this cycle_.

If a vector internally reallocs during an insert, it will resize exponentially, so you'll end up with order log2(n) reallocs. If you force it to reserve on each call, you will end up doing n reallocs instead.

cielavenir: How about this, then `indices` should know how much it should allocate before hand:

```cpp
#include <boost/range/adaptor/transformed.hpp>

void TriMesh::Append(const TriMesh& mesh)
{
    int offset = (int)vertices.size();
    vertices.insert(vertices.end(), mesh.vertices.begin(), mesh.vertices.end());
    auto indicesWithOffset = mesh.indices | boost::adaptors::transformed([&](const int &indice){return indice+offset;});
    indices.insert(indices.end(), indicesWithOffset.begin(), indicesWithOffset.end());
}
```

(uh-oh, there exists auto...)

cielavenir: oh

why don't you use indices.resize() like L2303 vertices, then reserve happens automatically

plugins/fclrave/fclspace.cpp
cielavenir: (I found something like cancellable boost_exit but I could not find one)

src/libopenrave/libopenrave.cpp
rschlaikjer: Both of these attempts to outsmart std::vector's internal logic seem to be slower than just letting it grow based on its internal heuristics. Simplified test case:
```c++
int main()
{
    std::vector<long> v;
    for (long i = 0; i < ITERCOUNT i++) {
#ifdef RESERVE
        v.reserve(v.size() + 1);
#elif defined RESIZE
        v.resize(v.size() + 1);
#endif
        v.emplace_back(i);
    }
}
```
With itercount = 100'000, it's clear just how pessimal reserve is:
```c++
mujin@G6DC229006MN:/h/m/cpp_units$ g++ -O3 improper_reserve.cpp && time ./a.out && g++ -O3 -DRESERVE improper_reserve.cpp && time ./a.out && g++ -O3 -DRESIZE improper_reserve.cpp && time ./a.out
// simple
real	0m0.002s
user	0m0.000s
sys	0m0.002s
// reserve
real	0m5.141s
user	0m0.964s
sys	0m4.177s
// resize
real	0m0.002s
user	0m0.002s
sys	0m0.000s
```

If we change itercount to 1'000'000'000 and ignore reserve because it would be too slow to complete, it seems like even resize isn't as fast as just calling emplace back:
```c++
mujin@G6DC229006MN:/h/m/cpp_units$ g++ -O3 improper_reserve.cpp && time ./a.out  && g++ -O3 -DRESIZE improper_reserve.cpp && time ./a.out
// simple
real	0m3.242s
user	0m0.869s
sys	0m2.374s
// resize
real	0m7.839s
user	0m2.399s
sys	0m5.438s

```

rschlaikjer: Oops, sorry, flaw in my test case for resize there. Leaving it for posterity, but with correct semantics
```c++
int main()
{
    std::vector<long> v;
    for (long i = 0; i < 1'000'000'000; i++) {
#ifdef RESERVE
        v.reserve(v.size() + 1);
        v.emplace_back(i);
#elif defined RESIZE
        v.resize(v.size() + 1);
        v.back() = i;
#else
        v.emplace_back(i);
#endif
    }
}
```
resize still seems to be slower than the naive implementation
```c++
mujin@G6DC229006MN:/h/m/cpp_units$ g++ -O3 improper_reserve.cpp && time ./a.out  && g++ -O3 -DRESIZE improper_reserve.cpp && time ./a.out
// simple
real	0m3.331s
user	0m0.876s
sys	0m2.455s
// resize
real	0m3.516s
user	0m1.196s
sys	0m2.320s
```

cielavenir: Let's try more practical case...

```
#include <vector>
#include <stdlib.h>

int main(int argc, char **argv)
{
    int step = strtol(argv[1], NULL, 10);
    std::vector<long> v;
    for (long i = 0; i < 1'000'000'000; i++) {
#ifdef RESERVE
        if(i%step==0)v.reserve(v.size() + step);
        v.emplace_back(i);
#elif defined RESIZE
        if(i%step==0)v.resize(v.size() + step);
        v[i] = i;
#else
        if(i%step>=0)v.emplace_back(i);
#endif
    }
}
```

```
$ g++ -O3 improper_reserve.cpp && time ./a.out 1000 && g++ -O3 -DRESIZE improper_reserve.cpp && time ./a.out 1000
// simple
real	0m13.288s
user	0m9.170s
sys	0m4.111s
// resize
real	0m12.503s
user	0m8.279s
sys	0m4.223s
```

rschlaikjer: There seems to be a bug in your test case as written where it won't emplace the same amount of elements for the final case (should just be no if check). Interestingly, I get results that prefer the simple case even when fixed:
```c++
#include <stdlib.h>
#include <stdio.h>

#include <vector>

int main(int argc, char** argv)
{
    int step = strtol(argv[1], NULL, 10);
    std::vector<long> v;
    for (long i = 0; i < 1'000'000'000; i++) {
#ifdef RESERVE
        if (i % step == 0)
            v.reserve(v.size() + step);
        v.emplace_back(i);
#elif defined RESIZE
        if (i % step == 0)
            v.resize(v.size() + step);
        v[i] = i;
#else
        v.emplace_back(i);
#endif
    }
    fprintf(stderr, "Total vector size: %u\n", v.size());
}
```

```
g++ -O3 foo.cpp && time ./a.out 1000 && g++ -O3 -DRESIZE foo.cpp && time ./a.out 1000
Total vector size: 1000000000
// Naive
real	0m2.288s
user	0m0.780s
sys	0m1.508s
Total vector size: 1000000000
// Resize
real	0m4.161s
user	0m2.810s
sys	0m1.351s
````

Comment (rschlaikjer) [reply src/libopenrave/libopenrave.cpp:2295]: Confused by your result actually - thought it might be a cpu arch difference but ran the test on a controller and saw the same skew I get locally
```
root@controller1007:~# g++ -O3 foo.cpp && time ./a.out 1000 && g++ -O3 -DRESIZE foo.cpp && time ./a.out 1000
Total vector size: 1000000000

real	0m5.908s
user	0m1.992s
sys	0m3.915s
Total vector size: 1000000000

real	0m13.029s
user	0m8.849s
sys	0m4.180s
```

Comment (cielavenir) [reply src/libopenrave/libopenrave.cpp:2295]: @rschlaikjer because you deleted the division...

Comment (rschlaikjer) [reply src/libopenrave/libopenrave.cpp:2295]: Ah, I see what you were getting at. Still seems to be about margin of error, and of course without the resize call you don't _need_ the additional check...
```c++
#elif defined RESIZE
        if (i % step == 0)
            v.resize(v.size() + step);
        v[i] = i;
#else
        if (i % step == 0)
            asm volatile("nop");
        v.emplace_back(i);
```
```

mujin@G6DC229006MN:/h/mujin$ g++ -O3 foo.cpp && time ./a.out 1000 && echo $? && g++ -O3 -DRESIZE foo.cpp && time ./a.out 1000
Total vector size: 1000000000

real	0m4.135s
user	0m2.694s
sys	0m1.439s
0
Total vector size: 1000000000

real	0m4.149s
user	0m2.748s
sys	0m1.400s
```

cielavenir:
```cpp
#include <vector>
#include <stdlib.h>

int main(int argc, char **argv)
{
    int step = strtol(argv[1], NULL, 10);
    std::vector<long> v;
    for (long i = 0; i < 1'000'000'000; ) {
#if defined(RESIZE)
        v.resize(v.size() + step);
        for (long j = 0; j < step; ++j, ++i) {
            v[i] = i;
        }
#else
        v.emplace_back(i); ++i;
#endif
    }
}
```

```
$ g++ -O3 improper_reserve.cpp && time ./a.out 1000 && g++ -O3 -DRESIZE improper_reserve.cpp && time ./a.out 1000

real	0m8.288s
user	0m2.634s
sys	0m5.619s

real	0m7.148s
user	0m2.389s
sys	0m4.758s
```

rschlaikjer: Running that same test case on c1007
```
root@controller1007:~# g++ -O3 foo.cpp && time ./a.out 1000 && g++ -O3 -DRESIZE foo.cpp && time ./a.out 1000

real	0m5.705s
user	0m1.688s
sys	0m4.016s

real	0m5.803s
user	0m1.554s
sys	0m4.241s
```

The difference here is again below the noise floor. At the end of the day, I think that the use case for explicit `resize` (or reserve) is that you know the _final_ size of the container. Resizing speculatively is more of a gamble - if your resize is large relative to the current size of the vector, it will be advantageous (doubling interval is small, you will save some reallocs). If your resize is small relative to the current size of the vector, it's just overhead (doubling interval already large, resize already within cap).

cielavenir: @ziyan please decide which, simple emplace_back or resize.

src/libopenrave/kinbody.cpp
```diff
@@ -4312,50 +4304,120 @@ void KinBody::_ComputeInternalInformation()
...
             it->first = -1;
             it->second = -1;
         }
-        vector<uint32_t> vcosts(_veclinks.size()*_veclinks.size(),0x3fffffff); // initialize to 2^30-1 since we'll be adding
-        for(size_t i = 0; i < _veclinks.size(); ++i) {
-            vcosts[i*_veclinks.size()+i] = 0;
+
+        // All of our arrays are essentially 2d look up tables, so create a wrapper to generate an array index from a 2d point
+        const size_t linksSize = _veclinks.size();
+        auto MakeIndex = [&](size_t x, size_t y) -> size_t { return x * linksSize + y; };
```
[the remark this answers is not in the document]
rschlaikjer: Can you elaborate on the reason behind these rules? 

With regard to lambdas (which do not appear to be called out in cpp.md?) - this expression is heavily repeated, so extracting and naming it makes it clearer what's going on / reduces risk of programmer error. Of the other options considered:
- `#define`s aren't naturally scoped, and require an #undef
- A separate function semantically separates the logic from where it's used and can result in lots of trivial helper functions littering the namespace.

Lambdas are naturally scoped, can occur next to the logic they are part of, and are (for simple cases such as this) trivially flattened by the compiler.

Similarly curious behind the reason for avoiding `auto`, especially when I see a fair amount of `decltype`, which is almost identical in function but slower for the compiler to process.

include/openrave/kinbody.h
```diff
@@ -2526,6 +2540,17 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
 
+private:
+    /// \brief Initialize a kinbody with one link composed of a list of geometries
+    ///
+    /// This is an internal function to provide a common implementation to the set of exposed InitFromGeometries APIs
+    /// \param geometries a list of geometry infos to be initialized into new geometry objects, note that the geometry info data is copied
+    /// \param visible if true, will be rendered in the scene
+    /// \param uri the new URI to set for the interface
+    template <typename GeometryIterableT>
+    bool _InitFromGeometriesInternal(const GeometryIterableT& geometries, const std::string& uri);
```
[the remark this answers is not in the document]
rschlaikjer: Not sure I agree that this is a small amount of code? It's 20+ lines, and would need to be duplicated three times to meet the current API spec and would still need to be duplicated twice if you removed the pointer version to handle both `list` and `vector`.

src/libopenrave/libopenrave.cpp
rschlaikjer: I think this is effectively resolved by the change to resize + assign as per Rosen's suggestion.

src/libopenrave/kinbody.cpp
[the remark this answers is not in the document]
ziyan: One reason to avoid `auto` that I find useful is to force the developer to think about the type. Sometimes they don't realize the resulting type after a long line of calls. They may not be paying enough attention to avoid copy for example. It also makes code review easier, since it shows the developer thought about it hard, and reviewers do not need to do the job of a compiler to figure out the type again in their heads.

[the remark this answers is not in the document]
rdiankov: It's better to be explicit than implicit. Adding trivial functions and typing out the types gives clarity to the code. Clarity, which I think is paramount as a codebase expands. 
I want to avoid a philosophical argument here, so I'm not going to comment on this further, please follow the C++ coding guidelines that you can find on gitlab.
Thanks

include/openrave/kinbody.h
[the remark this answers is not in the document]
rschlaikjer: Consolidated, see changes to other repos 
https://pastebin.mujin.co.jp/?ac586f6aa9761f4c#E6g3e1rpMLMLx6CG9g3GGkiUmU3opdPhL9KDdYdUAg7f

plugins/fclrave/fclspace.cpp
```diff
@@ -609,6 +613,29 @@ void FCLSpace::_Synchronize(FCLKinBodyInfo& info, const KinBody& body)
...
+        : _space(space), _ptr(ptr)
+    {
+    }
+    ~ScopedUserDataRemover()
+    {
+        if (!!_ptr) {
+            _space.RemoveUserData(_ptr);
+        }
+    }
+    void reset() {
```
rdiankov, line 630: let's capitalize functions

CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ set( CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS TRUE )
 # Define here the needed parameters
 set (OPENRAVE_VERSION_MAJOR 0)
 set (OPENRAVE_VERSION_MINOR 138)
```
rdiankov, line 7: let's increment minor version since kinbody.h changed

src/libopenrave/kinbodylink.cpp
```diff
@@ -733,54 +733,53 @@ KinBody::Link::GeometryPtr KinBody::Link::GetGeometry(int index)
...
+}
+
+template <typename GeometryIterableT>
+void KinBody::Link::_InitGeometriesInternal(const GeometryIterableT& geometries, bool bForceRecomputeMeshCollision) {
     _vGeometries.resize(geometries.size());
     size_t i = 0;
-    FOREACH(itinfo,geometries) {
-        _vGeometries[i].reset(new Geometry(shared_from_this(),*itinfo));
+    FOREACH(itGeometry, geometries) {
+        _vGeometries[i].reset(CreateGeometry(shared_from_this(), *itGeometry));
```
rdiankov, line 751: Just curious why a function is necessary here since replacing `new KinBody::Link::Geometry(plink, geometryInfo)` is just as clarifying..

rschlaikjer: The function is necessary to allow the template here to generalize - you can call this with either pointers or non-pointers, so the part of this function that actually changes based on that is extracted (`new KinBody::Link::Geometry(plink, *geometryInfoPtr)` vs `new KinBody::Link::Geometry(plink, geometryInfo)`).

rdiankov: okay, i removed the template member function, and now no need for  `CreateGeometry`

rdiankov:
thanks~

cielavenir:
Today I found that `python2 -m pytest mujintestcommon/ -v -s -k test_GetRobotControllerDOFOrder[robotTestInfo7] --noskiproot` started to fail after this merge.

I checked that manually building openrave:c91a4e442 and rerunning it is passing.

/cc @ntohge

cielavenir:
test resource missed connectedBody, adding it fixed the issue for unknown reason

PR #1315, opened 2023-10-02
WIP: initialize __mapReadableInterfaces on InitFromKinBodyInfo

eisoku9618:
`InitFromKinBodyInfo()` updates readable interfaces (i.e. keeps existing readable interfaces unless specified), but I think it should initialize them because the function name contains `Init` instead of `Update`.

```python
kbi = KinBodyInfo()
kbi.DeserializeJSON({'readableInterfaces': {'new': 'hoge'}})
body.SetReadableInterface('old', stringreaders.CreateStringReadable('old', 'fuga'))
body.InitFromKinBodyInfo(kbi)
body.GetReadableInterfaces()
```

For example, `body` should have `new` readable interface only in this case.

We should initialize readable interfaces not only for `InitFromKinBodyInfo()` but also for the other `InitXXX` such as `InitFromBoxes()`.


rdiankov:
Thanks

PR #1317, opened 2023-10-06
restore transform when EIO_SkipDOFValues and not added to env

eisoku9618:
@rdiankov cc @yoshikikanemoto @kanbouchou @ziyan Since d841c7633ec34cbd95c6486811eb2d5401b198ad commit, users can call `KinBody::ExtractInfo()` whenever body is not added to environment. This has a side effect that the transform is reset to identity because

1. `KinBody::KinBodyStateSaver::~KinBodyStateSaver()` does not do anything when body is not added to environment
2. `KinBody::ExtractInfo()` interanally set tranform to identity and DOFs to zeros in order to collect `KinBodyInfo` relying on `KinBodyStateSaver`

As a result, just calling `KinBody::ExtractInfo()` to a body which is not added to env resets its transform.

```
body.InitFromBoxes([[0,0,0,1,1,1]])
body.SetTransform([1,0,0,0,1,2,3])
print('before: %r' % body.GetTransformPose())
body.ExtractInfo(ExtractInfoOptions.SkipDOFValues)
print('after: %r' % body.GetTransformPose())
```

```
before: array([1., 0., 0., 0., 1., 2., 3.])
after: array([1., 0., 0., 0., 0., 0., 0.])
```

With this PR, we can restore at least transform and warn about DOF resetting.

related discussion
- https://github.com/rdiankov/openrave/pull/1316#issuecomment-1749914852

### DIFF
--- a/src/libopenrave/kinbody.cpp
+++ b/src/libopenrave/kinbody.cpp
@@ -6072,6 +6072,12 @@ void KinBody::ExtractInfo(KinBodyInfo& info, ExtractInfoOptions options)
 
     // in order for link transform comparision to make sense
     KinBody::KinBodyStateSaver stateSaver(shared_kinbody(), Save_LinkTransformation);
+    boost::shared_ptr<void> restoreTransformEvenWhenBodyIsNotAddedToEnv = nullptr;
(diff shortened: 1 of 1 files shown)


rdiankov:
thanks~

PR #1318, opened 2023-10-11
optimize jitterers by only processing links affected by active DOFs

yoshikikanemoto:
### motivation

Found inefficient code when i was debugging behavior of jitterer.

### changes

1. break from the `linkdistthresh` logic as soon as it found a violation
2. change the `linkdistthresh` logic to only consider links that are affected by active dofs

### test

jittering was 10% faster on some truckbot env.

- tested on simulation that goes through the new code
- autotester


rdiankov:
thanks

PR #1322, opened 2023-10-27
Fix nonAdjacentLinks and _listNonCollidingLinksWhenGrabbed were affected by collision callbacks.

yoshikikanemoto:
### description of the issue

- OpenRAVE does lazy initialization to `KinBody::_vNonAdjacentLinks` and `Grabbed::_listNonCollidingLinksWhenGrabbed`.
- IKSolver sets the collision callback to exclude end effector links when it solves IK without end effector collision check
- When the lazy initialization was triggered under the above collision callback was set, the result of the initialization was different from the one without the callback and wrong.

### changes

- Introduced `CO_IgnoreCallbacks` to ignore collision callbacks from collision checkers.
- Set `CO_IgnoreCallbacks` when it initializes  `KinBody::_vNonAdjacentLinks` or `Grabbed::_listNonCollidingLinksWhenGrabbed`.

### test

- `testopenrave:fixNonAdjacentInit20231026`
- pipeline: 405967

cc: @Puttichai


rdiankov:
Great catch~

cielavenir:
@yoshikikanemoto I'm afraid but this pull request broke `testopenrave/legacy/test_planning.py::test_PlanWithCollision[pr2test1.env.xml]`. I confirmed that the test case passes when I revert 7a4e6348ce58e539d034a8d62a91504c3e01a5ab . Please check.

/cc @ntohge

cielavenir:
Yoshiki suggested fix, pushed to https://github.com/rdiankov/openrave/pull/1327

PR #1323, opened 2023-10-27
Fixed dangling FCL cache

yoshikikanemoto:
### description

This PR fixes cache of `FCLCollisionManagerInstance` for self-collision checker wasn't cleared for previously grabbed bodies even when they were removed from the env.

This issue was not only causing memory leak, but also when the same address was assigned to new bodies, `FCLCollisionManagerInstance` for the removed bodies were assigned to the new bodies, resulted in false collision checking results.

Also added handling of dangling `FCLCollisionManagerInstance`  just in case (I haven't seen  it after this fix)

### changes

- When a KinBody is removed from env, also removes it from self-collision checkers of other kinbodies
- When `FCLCollisionManagerInstance` with null `ptrackingbody` is found in the cache, remove it and assign a new one.

### test

- Added a reproduction test case to `testopenrave:fixDanglingFCLCache20231027`
- Pipeline #660658

@Puttichai


rdiankov:
thanks~

PR #1324, opened 2023-10-27
Fix incorrect initialization of _listNonCollidingLinksWhenGrabbed when other grabbed bodies are modified before Grabbed::ComputeListNonCollidingLinks.

yoshikikanemoto:
### description of the issue

- The lazy initialization of `Grabbed::_listNonCollidingLinksWhenGrabbed` was relying on the state savers including `Save_GrabbedBodies` generated on creation of the `Grabbed` instance
- Grabbed bodies by the grabber and the grabbed body can be changed / recreated after the `Grabbed` instance is created and before the initialization of `_listNonCollidingLinksWhenGrabbed` is triggered and resulted in incorrect `_listNonCollidingLinksWhenGrabbed`.

### change

- changed to use the current grabbed bodies when the initialization of `_listNonCollidingLinksWhenGrabbed` is triggered.

### test

- added reproduction to `testopenrave:fixNonCollidingLinksWhenGrabbedShuffle20231027`
- Pipeline #660662

@Puttichai


rdiankov:
Thanks~

PR #1325, opened 2023-10-28
More fix for self-collision checker initialization issues for grabbed bodies.

yoshikikanemoto:
### motivation

1. Under certain condition, grabbed kinbodies were not initialized in self-collision checker
2. There was a problem on my previous PR (https://github.com/rdiankov/openrave/pull/1321)

### changes

1. Fix not initializing grabbed bodies to self-collision checker when cloning KinBody, restoring grabbed state from the state saver.
3. Fix cloning `_listNonCollidingLinksWhenGrabbed` to different env

### test

- tested on simulation and actual robot
- Pipeline #661562

FYI @Puttichai


rdiankov:
Thanks

PR #1328, opened 2023-11-07
add AABBFromOrientedBox

eisoku9618:
Add an utility function to convert an OBB to an AABB in the reference frame of the OBB

PR #1331, opened 2023-11-13
Fix QtCoin SetUserText

cielavenir:
https://github.com/rdiankov/openrave/pull/1328/commits/c11d17683d35495e4c1c3712019bfd68906b818d has typo, which prevents compilation when the machine has libcoin-dev

Why autotester works fine currently? Because jhbuild does not have libcoin-dev and qtcoin is not built.

/cc @jerasman @wkentaro @florisatmujin (I believe vision team should have libcoin-dev installed on their machine and they should be affected)


cielavenir:
clarify: This affects only local build and autotester does not have meanings.

florisatmujin:
> [c11d176](https://github.com/rdiankov/openrave/commit/c11d17683d35495e4c1c3712019bfd68906b818d) has typo, which prevents compilation when the machine has libcoin-dev
> 
> Why autotester works fine currently? Because jhbuild does not have libcoin-dev and qtcoin is not built.
> 
> /cc @jerasman @wkentaro @florisatmujin (I believe vision team should have libcoin-dev installed on their machine and they should be affected)

Yes, we use this a lot because the visualizer works better. And I just encountered the issue as well.

Thank you for fixing, I just confirmed by building locally with the branch of this PR that it works.

yoshikikanemoto:
thanks @cielavenir

PR #1333, opened 2023-11-23
Maintain active geometry group name

eisoku9618:
So that `CollisionChecker.GetBodyGeometryGroup()` returns group name correctly. 
- `GetBodyGeometryGroup()` is supported only by fclrave so far, so I updated fclrave only and not oderave etc.
- pipelineid=416799


eisoku9618:
I found an issue, so mark this as Draft.

eisoku9618:
fixed

kanbouchou:
@eisoku9618 this looks good, but is this still draft? can you update the status if you don't have any more change?

kanbouchou:
@rdiankov could you review this together with application side change? Together with application side change, this improves geometry group switching time a lot.

rdiankov:
thanks~

PR #1336, opened 2023-11-25
link crlibm only if found, via openrave_static_libraries

cielavenir:
aarch64 does not build crlibm and aarch64 build is currently broken


rdiankov:
Thanks

cielavenir:
clarifying that I ran the test in prior:

Pipeline #684133

PR #1339, opened 2023-12-04
Fix update gripper20231204

LLipter:
When a parameter inside `gripperInfo/customParameters` is modified, openrave will wipe out other members after loading diff notification received from webstack. This PR fixes the issue by recursively updating the custom parameter.


yoshikikanemoto (approved):

yoshikikanemoto:
LGTM. @rdiankov can you check and merge?

src/libopenrave/robot.cpp
```diff
@@ -77,7 +89,10 @@ void RobotBase::GripperInfo::DeserializeJSON(const rapidjson::Value& value, dRea
         if (memberName == "id" || memberName == "name" || memberName == "grippertype" || memberName == "gripperJointNames") {
             continue;
         }
-        orjson::SetJsonValueByKey(docGripperInfo, memberName, it->value);
+
+        // update objects recursively 
+        _PartialUpdate(docGripperInfo, memberName.c_str(), it->value, docGripperInfo.GetAllocator());
```
ziyan, line 94: @LLipter We already have `OpenRAVE::orjson::UpdateJson` why don't you use it instead?

ziyan: Or, add the recursive version of it to `openravejson.h` instead

LLipter: I moved the function to `openravejson.h` as you suggested. `OpenRAVE::orjson::UpdateJson` is not enough as it only updates the top-level key-value pair. We need to update all levels recursively.

yoshikikanemoto:
@Puttichai this one

yoshikikanemoto:
@ziyan can you check the latest change and approve?

PR #1341, opened 2023-12-09
Optimize collision check by FCL for GeometryType.Container and GeometryType.Cage

yoshikikanemoto:
@Puttichai found slowness of collision checking when a robot is fully inside a container. It turned out slowness came from the fact that `OpenRAVE` was always doing mesh-mesh collision checking. By using `fcl::Container`, `OpenRAVE` can generate `GeometryType.Container` and `GeometryType.Cage` in fcl space as a set of box primitives and speed up computation time of the collision checking. Observed computation time decreased by 50% on some examples.

tested on a real system and unit tests.


rdiankov:
Thanks~

cielavenir:
@rdiankov should this target be production branch?

yoshikikanemoto:
> @rdiankov should this target be production branch?

yes, i'll push to production branch

cielavenir:
confirmed https://github.com/rdiankov/openrave/commits/production/ now!

PR #1342, opened 2023-12-14
Fix CheckEndEffectorCollision

Puttichai:
### Summary

In `Manipulator::CheckEndEffectorCollision`, even though a manipulator's child link is collision-disabled, still need to check collision for the grabbed bodies that are grabbed by the link.

Debugged and fixed this problem with @yoshikikanemoto


rdiankov:
thanks~

PR #1343, opened 2023-12-18
Throw exception if input is non-contiguous for ExtractContiguousArrayToPointer

YutaKojio:
If an input for `ExtractContiguousArrayToPointer` is numpy array with non-contiguous memory, `PyArray_GETCONTIGUOUS` creates copy, not returns reference of the input.
In this case, the reference count for  `arrPtr` is `1`, and it is decreased to `0` by the destructor of `AutoPyArrayObjectDereferencer` when exiting this function.
This causes an unexpected behavior since the memory is released and the data can be destroyed.

This MR changes to throw excepetion for that case to be aware of unintended usage.

Pipeline #702392


rdiankov:
Maybe no need to increment patch for such a minor change

YutaKojio:
sure. I reverted the version change.

rdiankov:
thanks~

PR #1344, opened 2023-12-22
Fix for collision report to have collided geometry name correctly

eisoku9618:
- collided `Geometry` has been stored in CollisionReport since https://github.com/rdiankov/openrave/pull/990
- geometry group has been used for collision check since https://github.com/rdiankov/openrave/pull/1333
- geometry group has not a `Geometry` but a `GeometryInfo`

Because of these 3 points, now in production branch, `CollisionReport::__str__()` does not show us a collided geometry name when geometry group (i.e. padded geometries) is used.

In order to address this issue, I would like to store a geometry name in `FCLGeometryInfo` and pass the name to collision report in `CheckNarrowPhaseGeomCollision()`.

pipelineid=426516 pipelineid=426623

cc @MiraiHattori @Puttichai


plugins/fclrave/fclcollision.cpp
```diff
@@ -1217,20 +1216,23 @@ std::pair<FCLSpace::FCLKinBodyInfo::LinkInfo*, LinkConstPtr> FCLCollisionChecker
     return std::make_pair(link_raw, LinkConstPtr());
 }
 
-std::pair<FCLSpace::FCLKinBodyInfo::FCLGeometryInfo*, GeometryConstPtr> FCLCollisionChecker::GetCollisionGeometry(const fcl::CollisionObject &collObj)
+void FCLCollisionChecker::GetCollisionGeometryAndName(const fcl::CollisionObject &collObj, GeometryConstPtr& pgeom, std::string& geometryName)
 {
     const std::shared_ptr<const fcl::CollisionGeometry>& collgeom = collObj.collisionGeometry();
     FCLSpace::FCLKinBodyInfo::FCLGeometryInfo* geom_raw = static_cast<FCLSpace::FCLKinBodyInfo::FCLGeometryInfo *>(collgeom->getUserData());
```
kanbouchou, line 1222: let's use this opportunity to unify naming, please change `geom_raw` to `geomRaw`

include/openrave/collisionchecker.h
```diff
@@ -88,6 +88,7 @@ class OPENRAVE_API CollisionReport
     //@{
     KinBody::LinkConstPtr plink1, plink2; ///< the colliding links if a collision involves a bodies. Collisions do not always occur with 2 bodies like ray collisions, so these fields can be empty.
     KinBody::GeometryConstPtr pgeom1, pgeom2; ///< the specified geometries hit for the given links
```
kanbouchou, line 90: Could you add a comment saying these might be null if Collision is coming from geometry info?

rdiankov:
https://github.com/rdiankov/openrave/pull/985 fixes this, so i cannot merge this for fear of many conflicts.

PR #1345, opened 2023-12-22
Free items even if viewer is shutting down

rschlaikjer:
The viewer uses a delayed free mechanism where `Item`s that hit a refcount of zero are added to a list that is then actually deleted on the next call to Reset or UpdateFromModel. This breaks down when the viewer is destructed - the weak pointer to the viewer fails to lock (the viewer's refcount has reached zero, it is being destructed), and so the objects never get sent back to the viewer for deletion. In this scenario, delete the objects directly to prevent memory leaks when the viewer is repeatedly opened and closed.

There's a larger question about why the deleter is implemented this way - it appears that when the delete list is processed the only thing done other than `delete` is `PrepForDeletion`, which resets the user data pointers on the kinbody item. There is no comment explaining why this is necessary / this cannot be done in the ordinary-course destructor.

### DIFF
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -30,7 +30,7 @@ void DeleteItemCallbackSafe(QtOSGViewerWeakPtr wpt, Item* pItem)
     if( !!pviewer ) {
         pviewer->_DeleteItemCallback(pItem);
     } else {
-        RAVELOG_INFO("Viewer has been destroyed, deallocating item %p directly");
+        RAVELOG_WARN_FORMAT("Viewer has been destroyed, deallocating item %p directly", pItem);
         delete pItem;
     }
 }
--- a/plugins/qtosgrave/qtosgviewer.cpp
+++ b/plugins/qtosgrave/qtosgviewer.cpp
@@ -29,6 +29,9 @@ void DeleteItemCallbackSafe(QtOSGViewerWeakPtr wpt, Item* pItem)
     QtOSGViewerPtr pviewer = wpt.lock();
     if( !!pviewer ) {
         pviewer->_DeleteItemCallback(pItem);
+    } else {
+        RAVELOG_INFO("Viewer has been destroyed, deallocating item %p directly");
+        delete pItem;
     }
 }
 


plugins/qtosgrave/qtosgviewer.cpp
```diff
@@ -29,6 +29,9 @@ void DeleteItemCallbackSafe(QtOSGViewerWeakPtr wpt, Item* pItem)
     QtOSGViewerPtr pviewer = wpt.lock();
     if( !!pviewer ) {
         pviewer->_DeleteItemCallback(pItem);
+    } else {
+        RAVELOG_INFO("Viewer has been destroyed, deallocating item %p directly");
```
rdiankov, line 33: perhaps you want `RAVELOG_WARN_FORMAT`
also, maybe it is prudent to call `pItem->PrepForDeletion();`

rschlaikjer: > also, maybe it is prudent to call pItem->PrepForDeletion();

Well, this is sort of what I had a question about in my initial comment - if we look at the docstring for PrepForDeletion

> /// \brief called when OpenRAVE::Environment is locked and item is about to be removed

it suggests that prep for deletion is supposed to occur only while the environment is locked, but since this gets invoked from `QtOSGViewer::_DeleteItemCallback`, it's invoked whenever the object is dereferenced, which it _seems_ can be called without the environment locked (via `_Reset`, via `LoadEnvironment`, which locks the env _after_ calling `_Reset`), so it's unclear if this is true. 

If prep for deletion should always be called when the object is deleted, then it sounds like we have two cases:
- Viewer is live, we dereference an item and immediately call `PrepForDeletion`, but do not invoke the destructor/deallocate until the next `QtOSGViewer::UpdateFromModel`. 
- Viewer destructing, we dereference an item and immediately call `PrepForDeletion` _and_ delete it. 

Is there something in the *Item destructor that can cause issues / do you recall the reason for the two-stage delete? It seems like the ownership of these objects is local to the viewer, so I would not immediately guess that there would be races there. The initial version of the two-stage delete was added in `43a2e4765`, and `PrepForDeletion` in `69eeef7a8`.

rdiankov:
hmm.. okay, let's try it. thanks~

PR #1346, opened 2023-12-23
Fix en- and decrypt functions when not building with gpgme

PeterBowman:
This fixes a compilation issue when building OpenRAVE with `-DOPT_ENCRYPTION=OFF`. Regarding the error code for `OPENRAVE_EXCEPTION_FORMAT`, I have picked `ORE_NotImplemented` similarly to:

https://github.com/rdiankov/openrave/blob/012d3488334befa3a6dffd0b4847bb5fda1afe46/src/libopenrave/openravemsgpack.cpp#L359-L362

### DIFF
--- a/src/libopenrave-core/jsonparser/gpgutils.cpp
+++ b/src/libopenrave-core/jsonparser/gpgutils.cpp
@@ -210,11 +210,11 @@ bool GpgEncrypt(std::istream& inputStream, std::ostream& outputStream, const std
 #else
 
 bool GpgDecrypt(std::istream&, std::ostream&) {
-    throw OPENRAVE_EXCEPTION_FORMAT("Encryption not enabled in OpenRAVE, but decryption was requested.");
+    throw OPENRAVE_EXCEPTION_FORMAT0("Encryption not enabled in OpenRAVE, but decryption was requested.", ORE_NotImplemented);
 }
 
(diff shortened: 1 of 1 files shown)


cielavenir:
oh why https://github.com/PeterBowman/openrave/actions/runs/7310573801 is not linked automatically

PeterBowman:
> oh why https://github.com/PeterBowman/openrave/actions/runs/7310573801 is not linked automatically (but ah github action config does not test this change..)

Hello, @cielavenir. For that to happen, you need to add either `pull_request` or `pull_request_target` in [ci.yaml](https://github.com/rdiankov/openrave/blob/012d3488334befa3a6dffd0b4847bb5fda1afe46/.github/workflows/ci.yaml#L3-L4) ([reference](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore)). That would create a local workflow run automatically linked to this PR. Examples: [1](https://github.com/roboticslab-uc3m/kinematics-dynamics/blob/1facb7cfb839b45a1e767b23f946a7a2e088d2ee/.github/workflows/ci.yml#L3-L11), [2](https://github.com/robotology/yarp/blob/6e5676bf19b89cb0a75c9a83d593b484db93a5b7/.github/workflows/ci.yml#L3-L13).

rdiankov:
thanks!

PR #1347, opened 2024-01-10
release GIL before potentially heavy computations

yoshikikanemoto:
so that other python threads can make progresses

### DIFF
--- a/python/bindings/openravepy_int.cpp
+++ b/python/bindings/openravepy_int.cpp
@@ -2001,6 +2001,7 @@ void PyEnvironmentBase::Add(PyInterfaceBasePtr pinterface, py::object oAddMode, 
             addMode = py::extract<InterfaceAddMode>(oAddMode);
         }
     }
+    PythonThreadSaver threadsaver;
     _penv->Add(pinterface->GetInterfaceBase(), addMode, cmdargs);
 }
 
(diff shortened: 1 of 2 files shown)


yoshikikanemoto:
tested on one actual system and AutoTester Pipeline #727062

rdiankov:
One concern is that unlocking Gil for small operations might waste more CPU cycles. For example, are you sure Init function does enough computation?

yoshikikanemoto:
> One concern is that unlocking Gil for small operations might waste more CPU cycles. For example, are you sure Init function does enough computation?

I checked my profiling result again. `KinBody::InitFromGeometries` was indeed taking time so I am sure. But I think we can assume `InitFromBoxes` / `InitFromSpheres` would finish fast. so i removed GIL unlock from there.

yoshikikanemoto:
> > One concern is that unlocking Gil for small operations might waste more CPU cycles. For example, are you sure Init function does enough computation?
> 
> I checked my profiling result again. `KinBody::InitFromTriMesh` was indeed taking time so I am sure. But I think we can assume `InitFromBoxes` / `InitFromSpheres` would finish fast. so i removed GIL unlock from there.

Actually, `KinBody::InitFromTriMesh` was taking time on converting Python type to C++ type so unlocking GIL won't make a difference. I'll make another change.

yoshikikanemoto:
> Actually, `KinBody::InitFromTriMesh` was taking time on converting Python type to C++ type so unlocking GIL won't make a difference. I'll make another change.

done @rdiankov

rdiankov:
thanks~

PR #1348, opened 2024-01-10
Correct the condition to skip checking temp config

Puttichai:
### Summary

In the beginning of the while loop in the `Check` function, previously we checked whether to check `_vtempconfig` (which gets updated in every iteration) via `SetAndCheckState` by the following condition
```c++
istep >= start
```

The above condition is not correct since there is a scenario that results in the code skipping checking of some unchecked configurations. Consider the following scenario
- At the end of the first iteration of the while loop, `bComputeNewStep` is set to `false`. By now `_vtempconfig` has already been updated with new values (since `bHasMoved==true`).
- In the second iteration of the while loop, since `istep` remains `0` and `start` is `1`, the code will skip checking `_vtempconfig` even though `_vtempconfig` has previously been updated.

This PR updates the condition to
```c++
istep >= start || !bComputeNewStep
```


Puttichai:
@rdiankov Could you help review this pull request? I believe this issue is critical.

rdiankov:
thanks

PR #1349, opened 2024-01-10
new api SampleRangeSameDeltaTime for efficient trajectory sampling

kanbouchou:
Overview
------------

- `SamplePointsSameDeltaTime` is more efficient but doesn't support range (start and stop). This MR adds such api (`SampleRangeSameDeltaTime`).


rdiankov (changes requested):

python/bindings/include/openravepy/openravepy_trajectorybase.h
```diff
@@ -56,6 +56,10 @@ class OPENRAVEPY_API PyTrajectoryBase : public PyInterfaceBase
 
     object SamplePointsSameDeltaTime2D(dReal deltatime, bool ensureLastPoint, PyConfigurationSpecificationPtr pyspec) const;
 
+    object SampleRangeSameDeltaTime2D(dReal deltatime, dReal startTime, dReal stopTime, bool ensureLastPoint) const;
+
+    object SampleRangeSameDeltaTime2D(dReal deltatime, dReal startTime, dReal stopTime, bool ensureLastPoint, PyConfigurationSpecificationPtr pyspec) const;
```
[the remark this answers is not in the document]
kanbouchou: This is mostly because all the other methods such as `Insert` and `Sample` are using `PyConfigurationSpecificationPtr` as well, and I followed the same pattern. If we were to change this new api to use non-pointer (`PyConfigurationSpecification`), then all others should change at the same time for consistency?

src/libopenrave/trajectory.cpp
```diff
@@ -125,31 +125,44 @@ void TrajectoryBase::SamplePoints(std::vector<dReal>& data, const std::vector<dR
...
     }
-    if (ensureLastPoint && vtimes.back() < duration) {
+    if (ensureLast && vtimes.back() < duration) {
         vtimes.push_back(duration);
     }
+}
+
+void TrajectoryBase::SamplePointsSameDeltaTime(std::vector<dReal>& data, dReal deltatime, bool ensureLastPoint) const
+{
+    std::vector<dReal> vtimes;
```
rdiankov, line 143: @kanbouchou any way to do this without creating a new vector on new allocations?

kanbouchou: @rdiankov I've introduced `RangeGenerator` to replace this vector of times, please check again.

python/bindings/openravepy_trajectory.cpp
```diff
@@ -353,6 +346,42 @@ object PyTrajectoryBase::SamplePointsSameDeltaTime2D(dReal deltatime,
     return this->SamplePointsSameDeltaTime2D(deltatime, ensureLastPoint, pyspec);
 }
 
+object PyTrajectoryBase::SampleRangeSameDeltaTime2D(dReal deltatime,
+                                                    dReal startTime,
+                                                    dReal stopTime,
+                                                    bool ensureLastPoint) const
+{
+    std::vector<dReal>& values = _vdataCache;
```
rdiankov, line 354: @kanbouchou are you sure that `SampleRangeSameDeltaTime2D` won't be called from multiple threads? Users might think that a sample function might be safe to call from multiple threads...
perhaps can use thread local storage via `boost::thread_specific_ptr`?

kanbouchou: > are you sure that SampleRangeSameDeltaTime2D won't be called from multiple threads?

It depends on the user, so I'm not 100% sure. Thread safety property of `PyTrajectoryBase` class hasn't changed in this MR though, `GetWaypoints` and others are using `_vdataCache` as well.

If we want to ensure thread-safety of concurrent sampling / waypoint getting on this side, we could use thread-local stuff, or just a mutex. Or, we explicitly document that this class is not thread-safe and user would have to ensure it on the application side. What do you think?

rdiankov: Hmm... problem is that we don't have a good model of how users might be using trajectories. So although it might not be the case that they sample from the same trajectory in different threads, there's no guarantee.
In this case, temporary cache is probably best handled with thread local storage...

kanbouchou: I'll make sampling and waypoints reading multi thread safe.

kanbouchou: Made caches TLS and also added comment about thread safety in class description. Please check again.

rdiankov:
thanks!

PR #1353, opened 2024-01-17
Exclude virtual links from jittering computation

Puttichai:
### Summary

The robot links that do not have geometries (virtual links) should be excluded from jittering computation. Their AABB positions can interfere with the jittering process.


rdiankov:
sure, thanks~

PR #1355, opened 2024-01-26
add GetId() python bindings to KinBody, Link, Joint, Geometry

yoshikikanemoto:
when overwriting diff of `KinBodyInfo` to `KinBody` referenced by `referenceUri`, `GetId` python binding makes targetting things to overwrite easier.


rdiankov:
thanks

PR #1356, opened 2024-02-19
Test 128bit

eisoku9618:
Add MultiUInt64 class which has array of uint64_t to represent bigger integer than uint64_t

* Support the conversion between Python Int and C++ MultiUInt64
* Support the conversion between RapidJSON Array and C++ MultiUInt64
* Support the conversion from Python Int >= 2**64 to RapidJSON String


rdiankov:
not relevant anymore

PR #1359, opened 2024-02-27
allow EnvironmentBodyRemover not to restore a grabbed body when the grabbing link does not exist

eisoku9618:
Currently we allow `EnvironmentBodyRemover` not to restore active manipulator on restore if the active manipulator does not exist anymore in the body in case of CBAS change.

I think that we should do the same thing for grabbed bodies because a grabbing link might not exist in case of CBAS change.


src/libopenrave/robot.cpp
```diff
@@ -447,42 +447,69 @@ void RobotBase::RobotStateSaver::Release()
...
+        if( !!pmanip ) {
+            _pBodyRobot->SetActiveManipulator(pmanip);
+        }
+        else {
+            pmanip = _pBodyRobot->GetActiveManipulator();
+            RAVELOG_WARN_FORMAT("env=%s, robot=%s, cannot restore previous active manip=%s because it does not exist anymore. current active manip=%s", _pBodyRobot->GetEnv()->GetNameId()%_pBodyRobot->GetName()%_activeManipName%(!!pmanip ? pmanip->GetName() : ""));
+        }
+    }
+    if( !_pGrabbedInfos.empty() ) {
+        // it might be ok with grabbing link doesn't exist if ConnectedBody acitve state changes.
```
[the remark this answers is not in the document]
eisoku9618: @yoshikikanemoto Eventually we need to take care of grabbed bodies restoring failures either by 
`EnvironmentBodyRemover` itself or by callers of `EnvironmentBodyRemover`.

My only concern is an inconsistency of current `EnvironmentBodyRemover` between active manipulator and grabbed bodies. `EnvironmentBodyRemover` tries to restore both active manipulator and grabbed bodies, and the inconsistency is that it allows not restoring active manipulator while it terminates the program (c++11 and later) if unable to restore grabbed bodies.

> it is not obvious how this class should be doing.

How about adding the input arguments to `EnvironmentBodyRemover` to clarify how this class should behave?

```c++
EnvironmentBodyRemover(KinBodyPtr pBody, bool abortOnActiveManipulatorLost=false, bool abortOnGrabbedBodiesLost=true);
```


If this is still not good and we can leave the inconsistent behavior of `EnvironmentBodyRemover` for now, I can make another utility function `SetConnectedBodyActiveStatesKeepingGrabbedBodiesAsMuchAsPossible` and move the implementation to this.

```c++
void SetConnectedBodyActiveStatesKeepingGrabbedBodiesAsMuchAsPossible(pRobot, requestedConnectedBodyActiveStates) {
    std::vector<KinBody::GrabbedInfoPtr> pGrabbedInfos;
    pRobot->GetGrabbedInfo(pGrabbedInfos);

    // manually restore grabbed state instead of using EnvironmentBodyRemover because a grabbing link might get removed on connected body active state change.
    pRobot->ReleaseAllGrabbed();
    {
        EnvironmentBodyRemover robotremover(pRobot);
        pRobot->SetConnectedBodyActiveStates(requestedConnectedBodyActiveStates);
    }

    std::vector<KinBody::GrabbedInfoPtr>::iterator itRemoveFirst = pGrabbedInfos.begin();
    for (std::vector<KinBody::GrabbedInfoPtr>::const_iterator itGrabbedInfo = pGrabbedInfos.begin(); itGrabbedInfo != pGrabbedInfos.end(); ++itGrabbedInfo) {
        const KinBody::GrabbedInfoPtr pGrabbedInfo = *itGrabbedInfo;
        bool needToDelete = false;
        if( !pRobot->GetLink(pGrabbedInfo->_robotlinkname) ) {
            RAVELOG_WARN_FORMAT("env=%s, body=%s, cannot re-grab '%s' because grabbing link '%s' does not exist anymore.", pRobot->GetEnv()->GetNameId()%pRobot->GetName()%pGrabbedInfo->_grabbedname%pGrabbedInfo->_robotlinkname);
            needToDelete = true;
        }
        else if( !pRobot->GetEnv()->GetKinBody(pGrabbedInfo->_grabbedname) ) {
            RAVELOG_WARN_FORMAT("env=%s, body=%s, cannot re-grab '%s' because it does not exist in the environment.", pRobot->GetEnv()->GetNameId()%pRobot->GetName()%pGrabbedInfo->_grabbedname);
            needToDelete = true;
        }
        else {
            for( std::set<std::string>::const_iterator itLinkName = pGrabbedInfo->_setIgnoreRobotLinkNames.begin(); itLinkName != pGrabbedInfo->_setIgnoreRobotLinkNames.end(); ++itLinkName ) {
                if( !pRobot->GetLink(*itLinkName) ) {
                    RAVELOG_WARN_FORMAT("env=%s, body=%s, cannot re-grab '%s' because '%s' in _setIgnoreRobotLinkNames does not exist anymore.", pRobot->GetEnv()->GetNameId()%pRobot->GetName()%pGrabbedInfo->_grabbedname%(*itLinkName));
                    needToDelete = true;
                    break;
                }
            }
        }
        if( needToDelete ) {
            continue;
        }
        // will regrasp this info
        if( itRemoveFirst != itGrabbedInfo ) {
            *itRemoveFirst = std::move(*itGrabbedInfo);
        }
        ++itRemoveFirst;
    }
    const std::vector<KinBody::GrabbedInfoConstPtr> pConstGrabbedInfos(pGrabbedInfos.begin(), itRemoveFirst);
    pRobot->ResetGrabbed(pConstGrabbedInfos);
}
```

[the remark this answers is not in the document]
eisoku9618: discussed with @yoshikikanemoto 
At least, it is better to use a bitmask instead of two booleans for the argument.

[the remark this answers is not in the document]
eisoku9618: @yoshikikanemoto As we discussed, I changed this PR to use a bit mask for the options. Could you check this PR and the changes in the related internal repositories? pipelineid=456553

rdiankov:
@snozawa can you take a look at this MR? Thanks

yoshikikanemoto:
> @snozawa can you take a look at this MR? Thanks

@snozawa Eisoku revived this PR just now. Can you check?

snozawa:
@eisoku9618 

As for the restoring of `Grabbed`, 
- The `production` branch uses `KinBodyStateSaver`.
- This PR uses `ResetGrabbed`.

The former one just restores the original ptr of `Grabbed` instance, while the latter one creates the new `Grabbed` instance.

In this case, the former one preserves `_pGrabberSaver` and `_pGrabbedSaver`.
But, the latter one does not, so the resultant behavior of `Grabbed` becomes different.

So, I personally would like to revert the grabbed-restoring code, and would like to use the `KinBodyStateSaver` as in the original `production` branch.

Then, the next discussion would be how to handle the exception from the `KinBodyStateSaver`.
I'm not sure what's the best way to do it, but one of the way is adding the `Save_GrabbedBodiesNoExceptionOnInfoLost` or something.

Btw, throwing from `KinBodyStateSaver` invokes another issue : https://github.com/rdiankov/openrave/issues/1437

src/libopenrave/robot.cpp
```diff
@@ -559,42 +559,81 @@ void RobotBase::RobotStateSaver::Release()
...
-                _pBodyRobot->SetActiveManipulator(pmanip);
+EnvironmentBodyRemover::~EnvironmentBodyRemover() noexcept(true) {
+    _pBody->GetEnv()->AddKinBody(_pBody, IAM_StrictNameChecking, _environmentBodyIndex);
+    if( !!_pBodyRobot && !_activeManipName.empty() ) {
+        RobotBase::ManipulatorPtr pmanip = _pBodyRobot->GetManipulator(_activeManipName);
+        // it might be ok with manipulator doesn't exist if ConnectedBody acitve state changes.
+        if( !!pmanip ) {
+            _pBodyRobot->SetActiveManipulator(pmanip);
+        }
+        else if( !(_restoreOptions & EBRRO_AbortOnActiveManipulatorLost) ) {
```
snozawa, line 584: @eisoku9618 
Is this equivalent to the followig?
the following is slightly readable since it does not invert the flag:
```
        else if( (_restoreOptions & EBRRO_AbortOnActiveManipulatorLost) ) {
            stringstream ss;
            ss << "available manipulators are [";
            for( const RobotBase::ManipulatorPtr& pManip : _pBodyRobot->GetManipulators() ) {
                ss << pManip->GetName() << ", ";
            }
            ss << "]";
            throw OPENRAVE_EXCEPTION_FORMAT(_("env=%s, robot=%s, cannot restore the original active manip=%s because it does not exist anymore. %s"), _pBodyRobot->GetEnv()->GetNameId()%_pBodyRobot->GetName()%_activeManipName%ss.str(), ORE_Failed);
        }
        else {
            pmanip = _pBodyRobot->GetActiveManipulator();
            RAVELOG_WARN_FORMAT("env=%s, robot=%s, cannot restore the original active manip=%s because it does not exist anymore. current active manip=%s", _pBodyRobot->GetEnv()->GetNameId()%_pBodyRobot->GetName()%_activeManipName%(!!pmanip ? pmanip->GetName() : ""));
        }
```

src/libopenrave/robot.cpp
```diff
@@ -559,42 +559,81 @@ void RobotBase::RobotStateSaver::Release()
...
-        _pBody->GetEnv()->Add(_pBody, IAM_StrictNameChecking);
-        _grabbedStateSaver.Restore();
-        _grabbedStateSaver.Release();
-        if( !!_pBodyRobot && !_activeManipName.empty() ) {
-            OpenRAVE::RobotBase::ManipulatorPtr pmanip = _pBodyRobot->GetManipulator(_activeManipName);
-            // it might be ok with manipulator doesn't exist if ConnectedBody acitve state changes.
-            if( !!pmanip ) {
-                _pBodyRobot->SetActiveManipulator(pmanip);
+EnvironmentBodyRemover::~EnvironmentBodyRemover() noexcept(true) {
+    _pBody->GetEnv()->AddKinBody(_pBody, IAM_StrictNameChecking, _environmentBodyIndex);
```
snozawa, line 577: @eisoku9618 @kanbouchou 

Just to confirm.
Is it ok to always set `_environmentBodyIndex` when adding?

snozawa: NVM, this is already answered in other thread.

include/openrave/robot.h
```diff
@@ -1452,6 +1452,30 @@ class OPENRAVE_API RobotBase : public KinBody
     friend class Grabbed;
 };
 
+enum EnvironmentBodyRemoverRestoreOptions : uint8_t
+{
+    EBRRO_NoAbortOnInfoLost = 0, ///< will not abort even if any info cannot be restored.
```
snozawa, line 1457: what about like this?
```suggestion
    EBRRO_NoAbortOnInfoLost = 0, ///< will not abort even if any info handled in this enum cannot be restored. Note that the remover would abort if there is an exception which is not handled in this enum.
```

src/libopenrave/robot.cpp
```diff
@@ -559,42 +559,81 @@ void RobotBase::RobotStateSaver::Release()
...
-                _activeManipName = _pBodyRobot->GetActiveManipulator()->GetName();
-            }
+EnvironmentBodyRemover::EnvironmentBodyRemover(KinBodyPtr pBody, int restoreOptions) : _pBody(pBody), _environmentBodyIndex(_pBody->GetEnvironmentBodyIndex()), _restoreOptions(restoreOptions) {
+    if( _pBody->IsRobot() ) {
+        // If the manip comes from a connected body, the information of which manip is active is lost once the robot
+        // is removed from env. Need to save the active manip name so that we can set it back later when the robot
+        // is re-added to the env.
+        _pBodyRobot = RaveInterfaceCast<RobotBase>(_pBody);
+        if( !!_pBodyRobot->GetActiveManipulator() ) {
+            _activeManipName = _pBodyRobot->GetActiveManipulator()->GetName();
```
snozawa, line 569: @eisoku9618 

Any problem if we use `RobotStateSaver`'s `Save_ActiveManipulator` here?
maybe you need to restore by the name, not by the pointer?

snozawa:
talked in person with @eisoku9618 

- If no-abort option is supplied, if the grabbed is not restored or manipulator is not restored, warning is shown.
- If it's from the old connected body, that's reasonable behavior. But, if it's from the unexpected bug, it might be dangerous.
- So, before calling `EnvironmentBodyRemover`, or at the very beginning of `EnvironmentBodyRemover`, check the CBAS change. and then, check if the restored active manipulator is available or not. same for grabbed bodies. if not available, do not resotre them.

what do you think? @kanbouchou @yoshikikanemoto

eisoku9618:
> So, before calling EnvironmentBodyRemover, or at the very beginning of EnvironmentBodyRemover, check the CBAS change. and then, check if the restored active manipulator is available or not. same for grabbed bodies. if not available, do not resotre them.

@kanbouchou @yoshikikanemoto May I know your opinions about this. I also think this is a good direction. Some callers of EnvironmentBodyRemover for CBAS change need to reject the CBAS change request when restoring active manipulator and grabbed bodies is impossible, others accept CBAS change without trying restoring active manipulator and grabbed bodies. (i.e. depending on the context of the callers)

PR #1360, opened 2024-03-05
Support robot controller axis manufacturer code

kanbouchou:
Overview
===========

- Introduce `robotControllerAxisManufacturerCode` to joint class to support servos from different manufacturers in a daisy chain.
- Fix bug with `_vmimic` indefinitely growing in `PyJointInfo::_Update`

Detail
=======
- We have `robotControllerAxisProductCode` field in joint class already, and we could encode manufacturer name in there as well, but need to introduce a rule (separator) which I wanted to avoid. Also, `AttachedSensorInfo` has `_sensorMaker` so using same pattern here as well.
-


rdiankov:
@kanbouchou Isn't that what `ElectricMotorActuatorInfo::model_type` is for?

rdiankov:
> @kanbouchou Isn't that what `ElectricMotorActuatorInfo::model_type` is for?

nm, i see the comment for `robotControllerAxisProductCode`. In that case, i would add a little more to the comment to `robotControllerAxisManufacturerCode` to explain why it is different from product code.
Thanks,

kanbouchou:
> > @kanbouchou Isn't that what `ElectricMotorActuatorInfo::model_type` is for?
> 
> nm, i see the comment for `robotControllerAxisProductCode`. In that case, i would add a little more to the comment to `robotControllerAxisManufacturerCode` to explain why it is different from product code. Thanks,

I've added a bit more comment. Basically, product code by itself is not guaranteed to be globally unique because different manufacturers can happen to have same product code. That's why we need manufacturer code in addition to product code to handle devices from different manufacturers by a single device driver.

rdiankov:
thanks~

PR #1361, opened 2024-03-08
Fix usage of CompareFirst/SecondBodyName return value

Puttichai:
### Summary

The recently added `CompareFirstBodyName` and `CompareSecondBodyName` functions in `CollisionPairInfo` return `0` when the given string compares _equal_ to the first and second body names, respectively.

There are some places that expect to test for equality but are missing the test `== 0`.

For example, in `StateCheckEndEffector::_CheckCollisionPair`, the condition reads
```c++
if( !!pbody && cpinfo.CompareSecondBodyName(pbody->GetName()) ) {
```
while the [original code](https://github.com/rdiankov/openrave/blob/6bbddfbe67e983bfde2e2903dce852f34bb4a788/plugins/ikfastsolvers/ikfastsolver.cpp#L574) reads
```c++
if( it->GetBody() == pcolliding ) {
```


rdiankov:
thanks!

PR #1362, opened 2024-03-13
Don't re-extract body info if nothing has changed

rschlaikjer:
Instead of unconditionally resetting `BodyState` in `_UpdatePublishedBodies`, first test whether the state has already been initialized from the given body / update stamp. If it has, skip re-extracting all data. Since bodies are in a relatively stable order, this significantly improves average-case performance.

See rb PR 2010 for performance profiles.


rdiankov:
thanks~

PR #1364, opened 2024-03-13
Cache absence of collision bodies

rschlaikjer:
Previously, when an FCL collision manager was ensuring it had all bodies, it would only cache the collision body pointers if the body actually had any geometry that was enabled for collision. 
This meant that if the environment contained bodies with large numbers of non-colliding links, those bodies would _not_ get cached, and result in a lot of overhead every time a collision check is run.
    
With this patch, cache state is updated for non-colliding objects as well to avoid this extra work.

See rb PR 2010 for performance profiles.


rdiankov:
@kanbouchou can you review this?

kanbouchou:
> @kanbouchou can you review this?

@rdiankov I checked and looks good to me.

@rschlaikjer please test on real systems as well as making sure unit tests and integrated tests pass.

plugins/fclrave/fclmanagercache.cpp
```diff
@@ -232,35 +232,58 @@ void FCLCollisionManagerInstance::InitEnvironment(const std::vector<int8_t>& exc
...
+
+    // Clear our scratch buffer
+    _tmpSortedBuffer.resize(0);
+
+    // We need to make sure we resize _vecCachedBodies to allow direct lookup by body index, which we can do the first time we encounter a non-null body.
     bool ensuredVecCachedBodies = false;
-    std::vector<CollisionObjectPtr> vcolobjs;
+
+    // Local scratch buffer of collision objects to reuse to avoid allocations
+    std::vector<CollisionObjectPtr> vCollisionObjects;
```
rdiankov, line 249: @rschlaikjer can this vector be cached somewhere so we don't have to keep on recreating it?

plugins/fclrave/fclmanagercache.cpp
```diff
@@ -232,35 +232,55 @@ void FCLCollisionManagerInstance::InitEnvironment(const std::vector<int8_t>& exc
...
         }
+
+        // Calculate the collision objects for this body.
+        // Note that if there are no collision volumes for an entity, _AddBody will return false.
+        // We still want to cache the _absence_ of collision data however, to avoid expensive recomputes on non-colliding bodies, so cache the returned data even if it's empty.
+        const FCLSpace::FCLKinBodyInfoPtr& pinfo = _fclspace.GetInfo(body);
+        _AddBody(body, pinfo, _ensureBodiesCollisionObjectsCache, _linkEnableStatesBitmasks, false);
+        KinBodyCache& cache = _vecCachedBodies.at(bodyIndex);
+        cache.SetBodyData(pbody, pinfo, _linkEnableStatesBitmasks);
+        cache.vcolobjs.swap(_ensureBodiesCollisionObjectsCache);
```
rdiankov, line 281: @rschlaikjer  shouldn't `_ensureBodiesCollisionObjectsCache` be cleared to make sure the old `CollisionObjectPtr` are destroyed?

rschlaikjer: good catch, thansk

rdiankov:
thanks~

PR #1367, opened 2024-04-02
Fix check function

Puttichai:
### Summary

Add a missing check for `_vtempconfig` that got updated in the last iteration in the check loop.


rdiankov:
thanks~

PR #1369, opened 2024-04-08
Fix body-link transform update

superfashi:
Currently the body transform is incorrectly updated inside `KinBody::UpdateFromKinBodyInfo` due to an early change to   `_baseLinkInBodyTransform`, which resulted in a wrong link-body transformation state. This resulted in body transform changed in an unexpected way. This MR fixes this issue.

### DIFF
--- a/src/libopenrave/kinbody.cpp
+++ b/src/libopenrave/kinbody.cpp
@@ -2107,7 +2107,14 @@ void KinBody::SetDOFValues(const std::vector<dReal>& vJointValues, const Transfo
     if( _veclinks.size() == 0 ) {
         return;
     }
-    SetTransform(bodyTransform);
+    Transform baseLinkTransform = bodyTransform * _baseLinkInBodyTransform;
+    Transform tbase = baseLinkTransform*_veclinks.at(0)->GetTransform().inverse();
+    _veclinks.at(0)->SetTransform(baseLinkTransform);
+
+    // apply the relative transformation to all links!! (needed for passive joints)
+    for(size_t i = 1; i < _veclinks.size(); ++i) {
+        _veclinks[i]->SetTransform(tbase*_veclinks[i]->GetTransform());
+    }
     SetDOFValues(vJointValues,checklimits);
 }
 
--- a/src/libopenrave/kinbody.cpp
+++ b/src/libopenrave/kinbody.cpp
@@ -1398,9 +1398,18 @@ void KinBody::SetTransform(const Transform& bodyTransform)
     if( _veclinks.size() == 0 ) {
         return;
     }
-    Transform baseLinkTransform = bodyTransform * _baseLinkInBodyTransform;
-    Transform tbaseinv = _veclinks.front()->GetTransform().inverse();
-    Transform tapply = baseLinkTransform * tbaseinv;
+    /*
+     * B (body transform), L_i (link i transform)
+     * BL_i (link i related to body transform, unchanged)
+     * B' (new body transform), L'_i (new link i transform)
+     *
+     * L_i = B x BL_i    => BL_i = B^-1 x L_i
+     * B = L_1 x BL_1^-1 => B^-1 = BL_1 x L_1^-1
+     * L'_i = B' x BL_i
+     *      = B' x B^-1 x L_i
+     *      = (B' x BL_1 x L_1^-1) x L_i
+     */
+    const Transform tapply = bodyTransform * _baseLinkInBodyTransform * _veclinks.front()->GetTransform().inverse();
     FOREACH(itlink, _veclinks) {
         (*itlink)->SetTransform(tapply * (*itlink)->GetTransform());
     }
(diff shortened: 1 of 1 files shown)


src/libopenrave/kinbody.cpp
```diff
@@ -2098,14 +2107,7 @@ void KinBody::SetDOFValues(const std::vector<dReal>& vJointValues, const Transfo
...
     }
-    Transform baseLinkTransform = bodyTransform * _baseLinkInBodyTransform;
-    Transform tbase = baseLinkTransform*_veclinks.at(0)->GetTransform().inverse();
-    _veclinks.at(0)->SetTransform(baseLinkTransform);
-
-    // apply the relative transformation to all links!! (needed for passive joints)
-    for(size_t i = 1; i < _veclinks.size(); ++i) {
-        _veclinks[i]->SetTransform(tbase*_veclinks[i]->GetTransform());
-    }
+    SetTransform(bodyTransform);
```
Puttichai, line 2110: The code was written to not call `SetTransform` to avoid unnecessarily calling `_UpdateGrabbedBodies` and `_PostprocessChangedParameters`. Please revert this change.

src/libopenrave/kinbody.cpp
```diff
@@ -6287,13 +6289,6 @@ UpdateFromInfoResult KinBody::UpdateFromKinBodyInfo(const KinBodyInfo& info)
         }
         updateFromInfoResult = UFIR_Success;
     }
-    if( info._vLinkInfos.size() > 0 ) {
```
Puttichai, line 6290: `_baseLinkInBodyTransform` and `_invBaseLinkInBodyTransform` are part of the input `KinBodyInfo` and it is incorrect to leave them not updated.

src/libopenrave/kinbody.cpp
```diff
@@ -6395,10 +6390,9 @@ UpdateFromInfoResult KinBody::UpdateFromKinBodyInfo(const KinBodyInfo& info)
     }
 
     // transform
-    if( info.IsModifiedField(KinBodyInfo::KBIF_Transform) || (info._vLinkInfos.size() > 0 && info._vLinkInfos[0]->IsModifiedField(KinBody::LinkInfo::LIF_Transform)) ) {
-        Transform bodyTransform = info._transform * _invBaseLinkInBodyTransform;
-        if( GetTransform().CompareTransform(bodyTransform, g_fEpsilon) ) {
-            SetTransform(bodyTransform);
+    if( info.IsModifiedField(KinBodyInfo::KBIF_Transform) ) {
+        if( GetTransform().CompareTransform(info._transform, g_fEpsilon) ) {
```
Puttichai, line 6394: This if-statement is actually the only change that this pull request needs. Here we should be comparing `GetTransform()` (`this` kinbody's transform) with `info._transform` (the transform of the input kinbody).

Also, I think `info._vLinkInfos.size() > 0 ...` check above should not be removed.

Puttichai (changes requested):

src/libopenrave/kinbody.cpp
superfashi: This is where the bug come from. The `_baseLinkInBodyTransform` is the relative transformation of link to the body, and is simply not equal to the transformation of the first link.

src/libopenrave/kinbody.cpp
superfashi: If it is ensured that not calling `_UpdateGrabbedBodies` and `_PostprocessChangedParameters` will leave some internal state hanging, then this change can be reverted.

src/libopenrave/kinbody.cpp
superfashi: The check should be removed simply because the internal state will not change whether the input `info` changed at this point.

src/libopenrave/kinbody.cpp
Puttichai: Yes, I understand that `_baseLinkInBodyTransform` should be the relative transform you mentioned.

When `KinBodyInfo` is generated (supposedly from `KinBody::ExtractInfo`), the kinbody transform is set to 
identity first before link infos are extracted. So `info._vLinkInfos[0]->GetTransform()` should, in fact, represent the `_baseLinkInBodyTransform`.

Puttichai (changes requested):

src/libopenrave/kinbody.cpp
Puttichai: If  `_UpdateGrabbedBodies` and `_PostprocesChangedParameters` are absolutely necessary (for example, there will be problems if those are not called), then you can leave this change here.

superfashi:
Before the update, $L = B \times BL$. However, $B = L \times BL^{-1}$ is actually recorded.

If we just update $B$ to be $B'$, we actually want to update $L' = B' \times BL = L \times BL^{-1}$.

However, if this line exists, the code preemptively updated the internal state of $BL$, so you lose the original $BL$ which you needed for the calculation of $L'$.

I can be fairly sure that originally this bug is discovered, and instead of fixing the right place, an extra `Transform bodyTransform = info._transform * _invBaseLinkInBodyTransform;` is introduced in L6399 to "fix" the calculation. However this fix fails to cover the other case when the transform of base link isn't identity.

src/libopenrave/kinbody.cpp
superfashi: If you have strong opinion on keeping this, I can revert. Otherwise let's go with the method call which I think is cleaner.

src/libopenrave/kinbody.cpp
```diff
@@ -1398,9 +1398,18 @@ void KinBody::SetTransform(const Transform& bodyTransform)
...
+     * BL_i (link i related to body transform, unchanged)
+     * B' (new body transform), L'_i (new link i transform)
+     *
+     * L_i = B x BL_i    => BL_i = B^-1 x L_i
+     * B = L_1 x BL_1^-1 => B^-1 = BL_1 x L_1^-1
+     * L'_i = B' x BL_i
+     *      = B' x B^-1 x L_i
+     *      = (B' x BL_1 x L_1^-1) x L_i
+     */
+    const Transform tapply = bodyTransform * _baseLinkInBodyTransform * _veclinks.front()->GetTransform().inverse();
```
rdiankov, line 1412: @superfashi unless there's any reason to put this all in one line, please restore the three lines. We have them like that for clarity.

superfashi: It's slightly faster due to not copying anything but I'll revert.

rdiankov:
THanks!

PR #1370, opened 2024-04-09
Use convex hull as a fallback

Puttichai:
### Summary

This pull request adds a new safety measure to make sure that the padded mesh is safe to use.

The current library OpenRAVE uses for convex decomposition computation may, in some corner cases, output a list of convex hulls that do not entirely cover the original mesh vertices. To ensure safety, a check is added and if it turns out that the output hulls do not completely cover the original vertices, convex hull computation will be used instead.

The reason convex hull is not always used is that in many cases, the hull will be too big that when used for collision checking, it will give out too many false positives.

### DIFF
--- a/python/databases/convexdecomposition.py
+++ b/python/databases/convexdecomposition.py
@@ -285,19 +285,19 @@ class ConvexDecompositionModel(DatabaseGenerator):
         self.linkgeometry = []
         with self.env:
             links = self.robot.GetLinks()
-            for il,link in enumerate(links):
+            for il, link in enumerate(links):
                 geomhulls = []
                 geometries = link.GetGeometries()
-                for ig,geom in enumerate(geometries):
+                for ig, geom in enumerate(geometries):
                     trimesh = geom.GetCollisionMesh()
                     if len(trimesh.indices) == 0:
                         geom.InitCollisionMesh()
                         trimesh = geom.GetCollisionMesh()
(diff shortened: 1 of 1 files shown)


eisoku9618:
memo

- the current library OpenRAVE uses for convex decomposition computation: https://code.google.com/archive/p/convexdecomposition/
- V-HACD: https://github.com/kmammou/v-hacd
   - one of the most popular open-source ACD methods
- CoACD: https://github.com/SarahWeiii/CoACD
   - one of the latest ones (SIGGRAPH2022)

Puttichai:
@kanbouchou I notice there is this branch of yours https://github.com/rdiankov/openrave/tree/fix_convex_decomposition_thin_geometry

Do you still remember what kind of problems you encountered back then?

rdiankov:
thanks~

kanbouchou:
> @kanbouchou I notice there is this branch of yours https://github.com/rdiankov/openrave/tree/fix_convex_decomposition_thin_geometry
> 
> Do you still remember what kind of problems you encountered back then?

I vaguely recall that I encountered a strange convex hull of a thin object and tried to fix it by this branch. However, I thought it is difficult to predict how this affects different shapes out there. So I think I ended up modifying some parameters for convex decomposition such as `minTriangleConvexHullThresh` and not pursuing fixing the convex hull decomposition code.

PR #1371, opened 2024-04-17
Add centidegree unit

ziyan:
Centidegree unit is commonly used in the world of AGVs. Adding support for it can avoid using mysterious constants everywhere in the code.


include/openrave/units.h
```diff
@@ -80,7 +80,8 @@ OPENRAVE_API TimeUnit GetTimeUnitFromString(const std::string& pTimeUnit, TimeUn
 enum AngleUnit : int8_t
 {
     AU_Radian = 0,
-    AU_Degree = 1
+    AU_Degree = 1,
+    AU_Centidegree = 2,
```
rdiankov, line 84: value should be `-2`
also need corresponding MR on schemaprofile repo to support this..

ziyan: Ok, I guess it's for comparing `AU_Degree > AU_Centidegree`

ziyan:
Ping @rdiankov

rdiankov:
thanks~

PR #1372, opened 2024-04-17
Allow env.drawlabel to specify size of characters

yoshikikanemoto:


plugins/qtosgrave/osglodlabel.cpp
```diff
@@ -42,7 +42,7 @@ OSGLODLabel::OSGLODLabel(const std::string& label, const RaveVector<float>& colo
 
     // Set up other text element properties
     text->setText(label);
-    text->setCharacterSize(0.05);
+    text->setCharacterSize(0.05 * scale);
```
Hsifnus, line 45: How does scaling the arguments in the `text->setFontResolution` below here by `scale` as opposed to keeping them constant look?

braineo: i think we should expose the size as it is in the osg to the API. that's more intuitive for user to call. the height is in the same scale of the object rendered in the viewer. we can make the default height 0.05 then it is 50 mm height in the environment.


```suggestion
    text->setCharacterSize(height);
```

plugins/qtosgrave/osglodlabel.cpp
```diff
@@ -13,7 +13,7 @@ void OSGLODLabel::SetFont(osgText::Font* font) {
 }
 
 // OSG text label that scales by camera distance and also disappears if far away enough
-OSGLODLabel::OSGLODLabel(const std::string& label, const RaveVector<float>& color, osg::ref_ptr<osgText::Font> font) : osg::LOD() {
+OSGLODLabel::OSGLODLabel(const std::string& label, const RaveVector<float>& color, float scale, osg::ref_ptr<osgText::Font> font) : osg::LOD() {
```
braineo, line 16:
```suggestion
OSGLODLabel::OSGLODLabel(const std::string& label, const RaveVector<float>& color, float height, osg::ref_ptr<osgText::Font> font) : osg::LOD() {
```

braineo (changes requested):

yoshikikanemoto:
@rdiankov can we have this for better annotation to 3D viewer?

```python
# openrave.py -i ./src/robots/pr2-beta-sim.robot.xml
h=env.drawlabel('height=200mm', robot.GetLink('head_tilt_link').GetTransformPose()[4:], [0,0,0], 0.2)
h2=env.drawlabel('default(50mm)', robot.GetLink('r_wrist_roll_link').GetTransformPose()[4:], [0,0,0], 0.05)
h3=env.drawlabel('height=25mm', robot.GetLink('l_wrist_roll_link').GetTransformPose()[4:], [0,0,0], 0.025)
```

image

braineo:
🎉🎉🎉🎉

PR #1373, opened 2024-04-18
Clamp camera distance when zooming, handle all scroll events with Zoom()

seanzhang15:
This PR clamps camera distance to a maximum of 10^10 when zooming. This prevents the viewer published state from breaking the json parser due to large values. 

To override the osg default zoom behavior, this PR also lets the `Zoom` function handle all scroll events.

### DIFF
--- a/plugins/qtosgrave/osgviewerwidget.cpp
+++ b/plugins/qtosgrave/osgviewerwidget.cpp
@@ -72,12 +72,11 @@ public:
         switch( ea.getEventType() )
         {
         case osgGA::GUIEventAdapter::SCROLL:
-            if(_posgviewerwidget->IsInOrthoMode()) {
+            {
                 double factor = ea.getScrollingMotion() == osgGA::GUIEventAdapter::SCROLL_DOWN ? 1.1 : 0.9;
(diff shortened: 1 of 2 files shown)


plugins/qtosgrave/osgviewerwidget.cpp
```diff
@@ -1355,10 +1353,10 @@ void QOSGViewerWidget::Zoom(float factor)
     if (IsInOrthoMode()) {
         // if we increase _currentOrthoFrustumSize, we zoom out since a bigger frustum maps object to smaller part of screen
         _currentOrthoFrustumSize = _currentOrthoFrustumSize / factor;
-        _SetCameraViewOrthoProjectionPlaneSize(_currentOrthoFrustumSize);
+        _SetCameraViewOrthoProjectionPlaneSize(ClampDistance(_currentOrthoFrustumSize));
```
braineo, line 1356: frustum should have no effects on camera distance. when zoomig in ortho mode, the camera distance does not change

rdiankov:
thanks

braineo:
this only handles the mouse input, you also want to handle when the distance is set programmatically

PR #1376, opened 2024-04-26
Speed up poseTransformPoints

Puttichai:
### Summary

@longhongc found that `poseTransformPoints` is actually super slow. The changes in this pull request helps speed up `poseTransformPoints` function by up to 100x.

Changes include
- Take advantage of contiguous numpy array when extracting values from the inputs
- Compute transformed points explicitly instead of using `RaveTransform::operator*` as we can cache intermediate values used in computation.

```
Transforming 10 points
numpy: 0.000010 seconds per call
original poseTransformPoints: 0.000023 seconds per call
new poseTransformPoints: 0.000005 seconds per call

Transforming 1000 points
numpy: 0.000021 seconds per call
original poseTransformPoints: 0.001202 seconds per call
new poseTransformPoints: 0.000012 seconds per call

Transforming 1000000 points
numpy: 0.016562 seconds per call
original poseTransformPoints: 1.237362 seconds per call
new poseTransformPoints: 0.007604 seconds per call
```


rdiankov:
thanks~

PR #1377, opened 2024-04-29
Fix Manipulator's GetGripperDOFValues and GetArmDOFValues

Puttichai:
### Summary

Made changes to `Manipulator`'s `GetGripperDOFValues` and `GetArmDOFValues` so that they return an empty vector when the corresponding indices vector (gripper indices/arm indices) is empty.


rdiankov:
Thanks!

PR #1379, opened 2024-04-30
use std map instead of rapidjson document to save memory

kanbouchou:
## Summary

- In our application, we store many `IkFailureInfo`, and `rapidjson::Document _rCustomData` takes too much memory in the order of tens of GB. `rapidjson::Document` has large memory overhead.

## Alternatives

- Each `IkFailureInfo` has `rapidjson::Document` (**current way**). Easy to code, flexible but takes too much memory.
- Each `IkFailureInfo` has `rapidjson::Value`, and have a single `rapidjson::Document` for a vector of them. As flexible as `rapidjson::Document`, and almost as easy to code. Uses less memory, about one eighth of current approach.
- Replace `rapidjson::Document` with `map<string, vector<dReal>>`. Easy to code, memory usage is very small. Drawback is decreased flexibility compared to json representation and inability to pre-allocate memory. This is the **proposed approach**.
  -  To mitigate the memory preallocation issue, we could make the map value as a pair of boolean representing validity and vector holding real data, and in `IkFailureInfo::Reset`, instead of `map::clear()`, invalidate each entry by setting the validity flag to keep memory allocated. If key keeps changing, map size can grow over time.
 - Replace `rapidjson::Document` with `vector<pair<string, vector<dReal>>>`. Memory usage is small and preallocation is possible. A bit harder to use, less flexible compared to json representation.

## Experiment
Measure memory usage to store key-value pair (key is `"parameterkey"` and value is empty vector).

|                                         | Memory usage per key-value pair (Kbytes) |Speed per key-value pair insertion (nsec)|
|-----------------------------------------|-----------------------------|-----------------------------|
| rapidjson::Document                     | 4.2                         |1104|
| rapidjson::Value                        | 0.52                        |130|
| std::map<string, vector<dReal>>         | 0.15                        |45|
| std::vector<pair<string, vector<dReal>> | 0.09                        |35|

[key_value_space_usage_test.zip](https://github.com/rdiankov/openrave/files/15173075/key_value_space_usage_test.zip)

[key_value_space_usage_speed_test.zip](https://github.com/rdiankov/openrave/files/15175490/key_value_space_usage_speed_test.zip)


rdiankov:
Can you test the speed here?
My one issue with `std::map` is that every element is its own allocation on the heap...

To tell you the truth, I'd rather this be a `std::vector`, that we carefully control how it is allocated/deallocated. And instead of strings for the names, we can use enums.

kanbouchou:
> Can you test the speed here? My one issue with `std::map` is that every element is its own allocation on the heap...

I've updated the description, please check. `rapidjson::Document` is more than a 1us, `std::map` is 45ns, `std::vector` is 35ns.

> To tell you the truth, I'd rather this be a `std::vector`, that we carefully control how it is allocated/deallocated. And instead of strings for the names, we can use enums.

Yes, we can do it with vector of enum and actual data pair (`vector<pair<int, vector<dReal>>`). However, we cannot reserve the memory of inner vector easily. Also, when inserting an element, we need to search for existing element with same enum value.

Another option is `std::map` whose value is a pair of boolean and actual data. In `IkFailureInfo::Reset`, we skip calling `std::map::clear` but set boolean value to false to invalidate the data structure to avoid memory deallocation. Concern with this approach is map size can theoretically increase over time without bound, but if keys are a small finite set, it's ok in practice.

What do you think?

rdiankov:
To tell you the truth, I think having a custom data struct for key/value store is not so great. It creates a lot of ambiguity and is not very memory efficient to represent everything with `std::vector<dReal>`.  How about we completely get rid of the custom data struct and instead have different members for the different fields? On Mujin side, we can derive from IkFailureInfo and have our own memory-efficient structures

kanbouchou:
> To tell you the truth, I think having a custom data struct for key/value store is not so great. It creates a lot of ambiguity and is not very memory efficient to represent everything with `std::vector<dReal>`. How about we completely get rid of the custom data struct and instead have different members for the different fields? On Mujin side, we can derive from IkFailureInfo and have our own memory-efficient structures

Yes, we could do it that way. So we would have a derived class of `IkFailureInfo` which can do more efficient internal representation of what Mujin needs. In that case, 
- I would add following api's for accessing custom values to `IkFailureInfo`.
```
virtual void IkFailureInfo::SetCustomRealValue(const char* key, dReal value) = 0;
virtual void IkFailureInfo::SetCustomRealValues(const char* key, const vector<dReal>& value) = 0;
virtual dReal IkFailureInfo::GetCustomRealValue(const char* key) = 0;
virtual const vector<dReal>& IkFailureInfo::GetCustomRealValues(const char* key) = 0;
```
- I would simply remove `rapidjson::Document _rCustomData` member from `IkFailureInfo` and let each user handle it in their specific ways if needed.

Does it sound good?

rdiankov:
If openrave doesn't use custom data anymore, and we have derived classes that will have their own data storage and functions, then what is the value of having SetCustomRealValue/GetCustomRealValue in the base class?

kanbouchou:
> If openrave doesn't use custom data anymore, and we have derived classes that will have their own data storage and functions, then what is the value of having SetCustomRealValue/GetCustomRealValue in the base class?

`IkFailureAccumulatorBase::GetNextAvailableIkFailureInfo()` returns a reference to `IkFailureInfo`, and to access custom members of our derived class, we either need to downcast to `DerivedIkFailureInfo` and access its members directly, or access them via virtual functions. In terms of speed, `static_cast` is sligherly faster than virtual accessor functions (`dynamic_cast` is much slower). Do you want to go with downcasting by `static_cast` rather than having accessor apis?

```
0 base:total=23091.8 usec, average=2.30918 nsec(10000000 samples)
00
1 virtual:total=28271.2 usec, average=2.82712 nsec(10000000 samples)
BB
2 static_cast:total=23334.7 usec, average=2.33347 nsec(10000000 samples)
BB
3 dynamic_cast:total=67286.9 usec, average=6.72869 nsec(10000000 samples)
BB
```
[cast_virtual_speed_test.zip](https://github.com/rdiankov/openrave/files/15176687/cast_virtual_speed_test.zip)

rdiankov:
Since we control "IkFailureAccumulatorBasic" in dataaccumulator.h, static casting should be safe.

kanbouchou:
> Since we control "IkFailureAccumulatorBasic" in dataaccumulator.h, static casting should be safe.

Yes, in that context it's fine to use static cast.

But I'm afraid things may not work when we use ikfast solver in openrave who doesn't know the `MujinIkFailureInfo` so it would require it's own derived class, say `OpenRAVEIkFailureInfo` to store custom values. When we serialize `IkFailureInfo` instance to binary format, the serializer needs to know which of the derived classes the type of the instance is. Moreover, when we deserialize the binary data, we don't know how to parse the data without knowing the derived class type.

I'd like to propose to add `map<string, pair<bool, vector<dReal>>> _mapCustomData;` in base `IkFailureInfo` class again. Then, `IkFailureInfo::Reset` would just set the boolean of the value pair to false without calling `map::clear`. Then, we should not need to have any derived classes and thus the data I/O can remain clean. Memory reallocation problem you were afraid of is avoided by skipping `map::clear`, and I think slight computational time increase of map key search when adding a new custom key-value pair should be small because of log(N) complexity and expectedly small number (much less than 100) of keys. If you like, key type of `_mapCustomData` can be integer (enum value) as well.

rdiankov:
We need to figure out how to make openrave as fast as possible. This code is on a critical path for a lot of things; and the usage of maps is actually one thing I been wanting to get rid of for quite sometime.
Being general is always at odds with being optimal. From the beginning, openrave has tried to be general to a point where a lot of things became too slow...
I think the question we should be asking ourselves is: what is the design to make this most optimal? Let's not worry too much about generality 😉

kanbouchou:
> We need to figure out how to make openrave as fast as possible. This code is on a critical path for a lot of things

Yes, reducing overhead is very important especially for this code.

> and the usage of maps is actually one thing I been wanting to get rid of for quite sometime. Being general is always at odds with being optimal. From the beginning, openrave has tried to be general to a point where a lot of things became too slow... I think the question we should be asking ourselves is: what is the design to make this most optimal? Let's not worry too much about generality 😉

I agree that we have to avoid slowness because we aim for too much generality. However, I think we still need a support for openrave iksolver, many projects still rely on it and I think we should continue to support it. If you don't like map, what about following options?
- virtual accessors (https://github.com/rdiankov/openrave/pull/1379#issuecomment-2088257517), overhead is like 0.5ns.
- `vector<pair<int, vector<dReal>>>`  to represent custom values in base class. Each element is key-value pair where key is represented by an integer. To invalidate, we never clear this outer vector, and set the integer -1 or 0 to preserve memory.

rdiankov:
Not sure what you mean by virtual accessors, can you give an example?

And using `vector<pair<int, vector<dReal>>>` without clearing the vectors sounds like a good option that will allow us to transition without changing too much code. Just to avoid `std::pair`.

Thanks

kanbouchou:
> Not sure what you mean by virtual accessors, can you give an example?

I meant to define a set of api's in `IkFailureInfo` and leave underlying data structure and implementation to the derived class.
```
virtual void IkFailureInfo::SetCustomRealValue(const char* key, dReal value) = 0;
virtual void IkFailureInfo::SetCustomRealValues(const char* key, const vector<dReal>& value) = 0;
virtual dReal IkFailureInfo::GetCustomRealValue(const char* key) = 0;
virtual const vector<dReal>& IkFailureInfo::GetCustomRealValues(const char* key) = 0;
```

> And using `vector<pair<int, vector<dReal>>>` without clearing the vectors sounds like a good option that will allow us to transition without changing too much code.

I'm also fine with this approach as well. Each element can be a struct instead of pair to be clearer. 
```
///\brief holds custom data key-value pair. To be used to avoid reallocating memory.
template<typename DataType> struct KeyValueData
{
  void Reset() { customDataIndex = 0; }
  DataType data; ///<   if customFieldIndex is 0, invalid
  uint64_t customDataIndex; ///< represents semantic meaning of data. If 0, data is invalid.
};
```

What about going forward with adding `vector<KeyValueData<vector<dReal>>> _vCustomData` in `IkFailureInfo`?

rdiankov:
I wouldn't call it `DataType` since it makes me think of int, float, etc.
Perhaps `DataNameId` would be better.
Are all our current custom data fields Reals? Do we use integers, and more specifically, do we store large amount of data in this custom field? (I hope not).

kanbouchou:
> Do we use integers, and more specifically, do we store large amount of data in this custom field? (I hope not).

We have less than 10 custom data fields currently. So far, we are not using integer. Currently used types are vector of float as well as float scalar.

cielavenir:
solutionIndices is float - actually since I entered Mujin I wonder why.

kanbouchou:
> solutionIndices is float - actually since I entered Mujin I wonder why.

@cielavenir `solutionIndices` is part of `IkReturn`'s custom map, so different from the custom data in `IkFailureInfo`

rdiankov:
Hmm... might as well get rid of IkReturn's map while we are at it, but can do that at a later time.
Sure, for now, we can go forward with `vector<KeyValueData<vector<dReal>>>`

include/openrave/openravecontainer.h
```diff
@@ -0,0 +1,265 @@
...
+        data = data_;
+    }
+
+    uint64_t nameId = 0; ///< id for the semantic meaning of data. 0 means data is not valid.
+    DataType data; ///< user-defined data. If nameId is 0, invalid
+};
+
+/// \brief Associative data structure similar to map, optimized to avoid memory allocation. More suitable for relatively small number of elements.
+/// Similar to map, nameId(key) is unique. However, there can be multiple invalid nameIds.
+template<typename DataType> struct Map
```
rdiankov, line 64: Can we give it a less generic name than `Map`? `CustomVectorMap`, `VectorBackedMap`,  etc

include/openrave/openravecontainer.h
```diff
@@ -0,0 +1,265 @@
...
+
+#include <stdint.h>
+#include <vector>
+
+namespace OpenRAVE {
+
+namespace orcontainer {
+
+/// \brief Represents Key-value pair.
+template<typename DataType> struct NamedData
```
rdiankov, line 32: How about name it to `NameIdData`?

include/openrave/openravecontainer.h
```diff
@@ -0,0 +1,265 @@
...
+
+    ///\brief Checks validity of this object
+    ///\return whether valid
+    inline bool IsValid() const { return nameId > 0; }
+
+    ///\brief gets value
+    inline const DataType& GetValue() const { return data; }
+
+    ///\brief gets key
+    inline uint64_t GetKey() const { return nameId; }
```
rdiankov, line 47: `GetNameId` is better..

include/openrave/openravecontainer.h
```diff
@@ -0,0 +1,265 @@
...
+    /// \brief gets number of valid entries
+    /// \return number of valid entries
+    inline size_t GetSize() const { return _numValidElements; }
+
+    /// \brief invalidates all entries while preserving memory
+    void Clear()
+    {
+        // Just invalidate each element, but keep memory allocated
+        for (NamedData<DataType>& customData : _vNamedDatas) {
+            customData.Invalidate();
```
rdiankov, line 75: Inside the `Clear` function, you call `Invalidate`? Can you standardize on one name?

include/openrave/openravecontainer.h
```diff
@@ -0,0 +1,265 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /// \brief inserts element
+    /// \param nameId key to insert value for
+    /// \param data value for key.
+    void Insert(uint64_t nameId, const DataType& data)
```
rdiankov, line 158: Not too important, but you should probably have a return value similar to map::insert indicating `true` if new position was inserted, or `false` if it was overwriting.

include/openrave/openravecontainer.h
```diff
@@ -0,0 +1,265 @@
...
+    {
+        OPENRAVE_ASSERT_FORMAT0(nameId != 0, "nameId cannot be 0. 0 is reserved for invalid", OpenRAVE::ORE_InvalidArguments);
+
+        int64_t invalidIndex = -1;
+        for (size_t index = 0; index < _vNamedDatas.size(); ++index) {
```
rdiankov, line 163: you should make a comment here that you are looking through the entire array in order to find invalid `nameId`s too

include/openrave/openravecontainer.h
```diff
@@ -0,0 +1,265 @@
...
+
+    /// \brief gets iterator to the element after last element
+    /// \return iterator to the element after last element
+    Iterator GetEnd() const
+    {
+        return Iterator(_vNamedDatas, _endValidElementsIndex, _endValidElementsIndex);
+    }
+
+private:
+    std::vector<NamedData<DataType>> _vNamedDatas; ///< Vector of elements. Vector is not sorted. For small size, faster to keep it unsorted and do brute-force search.
```
rdiankov, line 255: Why not maintain this in binary order via std::_upper_bound? Given the low number of objects we deal with here, perhaps doesn't make a difference ;0/

src/libopenrave/iksolver.cpp
```diff
@@ -84,18 +84,25 @@ void IkFailureInfo::SaveToJson(rapidjson::Value& rIkFailureInfo, rapidjson::Docu
     if( !_description.empty() ) {
         orjson::SetJsonValueByKey(rIkFailureInfo, "description", _description, alloc);
     }
-    if( !_rCustomData.IsNull() && _rCustomData.MemberCount() > 0 ) {
-        rapidjson::Value rCustomDataCopy;
-        rCustomDataCopy.CopyFrom(_rCustomData, alloc);
-        rIkFailureInfo.AddMember("mapdata", rCustomDataCopy, alloc);
+    if (_mapCustomData.GetSize() > 0) {
+        rapidjson::Value rCustomData(rapidjson::kArrayType);
```
rdiankov, line 88: Why not save `rCustomData` as an object? Should be easier for python to access, and we don't have to worry about speed/memory too much.

include/openrave/openravecontainer.h
kanbouchou: Yes, I was thinking that we may not benefit much from sorting given relatively small number of elements.

Also, because we are not actually removing elements from this vector in `Erase` to preserve memory, it is computationally expensive to keep it sorted. One insertion or erasing operation can trigger to manipulate many elements to keep this vector sorted.

src/libopenrave/iksolver.cpp
kanbouchou: I think json only accepts key as string, and here `nameId` is integer, so cannot make this `rCustomData` an object.

include/openrave/openravecontainer.h
kanbouchou: Thanks added.

include/openrave/openravecontainer.h
kanbouchou: Thanks, changed.

include/openrave/openravecontainer.h
kanbouchou: Actually, it seems vector of key-value pair is slower than having two separate vectors, for each of key and value. Let me restructure and remove this class.

rdiankov: Really?! What's your test?

kanbouchou:
```
g++ -O3 map_impl_test.cpp && ./a.out 
numElements=2
0 stdmap_insert::total=201.37 usec, average=2.0137 nsec(100000 samples)
1 ormap1_insert::total=703.282 usec, average=7.03282 nsec(100000 samples)
2 ormap2_insert::total=702.533 usec, average=7.02533 nsec(100000 samples)
3 stdmap_find::total=501.288 usec, average=5.01288 nsec(100000 samples)
4 ormap1_find::total=647.529 usec, average=6.47529 nsec(100000 samples)
5 ormap2_find::total=704.756 usec, average=7.04756 nsec(100000 samples)
numElements=8
6 stdmap_insert::total=2234.99 usec, average=22.3499 nsec(100000 samples)
7 ormap1_insert::total=3717.06 usec, average=37.1706 nsec(100000 samples)
8 ormap2_insert::total=3584.03 usec, average=35.8403 nsec(100000 samples)
9 stdmap_find::total=2756.14 usec, average=27.5614 nsec(100000 samples)
10 ormap1_find::total=3564.97 usec, average=35.6497 nsec(100000 samples)
11 ormap2_find::total=3819.25 usec, average=38.1925 nsec(100000 samples)
numElements=32
12 stdmap_insert::total=12598.2 usec, average=125.982 nsec(100000 samples)
13 ormap1_insert::total=38134.2 usec, average=381.342 nsec(100000 samples)
14 ormap2_insert::total=28884.3 usec, average=288.843 nsec(100000 samples)
15 stdmap_find::total=14316.1 usec, average=143.161 nsec(100000 samples)
16 ormap1_find::total=22810.9 usec, average=228.109 nsec(100000 samples)
17 ormap2_find::total=23629.2 usec, average=236.292 nsec(100000 samples)
```
ormap1 is single vector of key-value pairs, ormap2 is two separate vectors. For insertion, which is on critical path, two vector approach is faster. But generally speaking std map is the fastest.

[Uploading map_impl_test.zip…]()

include/openrave/openravecontainer.h
kanbouchou: Thanks changed to `VectorBackedMap`

include/openrave/openravecontainer.h
kanbouchou: image
Maybe having two vectors, we get better cache hit. 

With ormap2 with two separate vectors, we are accessing the name id vector most of the time, and we access the value vector just once after finding where to actually write value. We are reading small amount of memory contiguously. 

With vector of key-value pair, we are interested in keys only for most of the time, but keys are more sparsely located on memory and maybe higher chance of cache miss?

rdiankov: Makes sense about the cache lines.

rdiankov:
woot~

PR #1380, opened 2024-04-30
Fix restart viewer thread does not restart

yoshikikanemoto:
It seems after d99b5bd6 , `RaveDestroy()` terminates the viewer thread and later `SetViewer()` call ends up waiting forever for the viewer thread to do the job. This PR allows to restart the viewer thread on `RaveInitialize()` and `SetViewer()`.


rdiankov:
thanks~

PR #1381, opened 2024-05-01
Add TimeStamp UnitInfo

shossjer:


include/openrave/units.h
```diff
@@ -89,23 +89,38 @@ OPENRAVE_API const char* GetAngleUnitString(AngleUnit unit);
 OPENRAVE_API AngleUnit GetAngleUnitFromString(const char* pAngleUnit, AngleUnit defaultAngleUnit);
 OPENRAVE_API AngleUnit GetAngleUnitFromString(const std::string& pAngleUnit, AngleUnit defaultAngleUnit);
 
+/// \brief time stamp unit
+enum StampUnit : int8_t
```
rdiankov, line 93: Can we call this `TimeStampUnit`?

include/openrave/units.h
```diff
@@ -89,23 +89,38 @@ OPENRAVE_API const char* GetAngleUnitString(AngleUnit unit);
...
+{
+    SU_Second = 0,
```
rdiankov, line 95: Would be great if you can add that it is the time from linux epoch

include/openrave/units.h
```diff
@@ -418,6 +432,68 @@ inline T GetAngleUnitConversionScale(const AngleUnit sourceAngleUnit, const Angl
...
+    if( unit == OpenRAVE::SU_Microsecond ) {
+        return T(1e6);
+    }
+
+    throw OPENRAVE_EXCEPTION_FORMAT("Unsupported time stamp unit '%s'", GetStampUnitString(unit), ORE_StampUnitInvalid);
+    return T{};
+}
+
+template <typename T>
+inline T GetStampUnitConversionScale(const std::string &sourceStampUnit, const std::string &targetStampUnit)
```
rdiankov, line 480: I think TimeStamp units conversation has to be handled differently since we could have integers, and I don't want to introduce floating point...

rdiankov:
@shossjer Given this breaks back compat, do you have a set of changes throughout the entire system ready for just the name change?
Thanks,

rdiankov:
thanks~

PR #1382, opened 2024-05-02
Use GripperInfo's chuckingDirection and gripperJointNames

snozawa:
# Summary
- Officially support `chuckingDirection` in `gripperInfo` to deprecate gripper settings in `ManipulatorInfo`. Now it's recommended to use `GripperInfo` for `chuckingDirection` and `gripperJointNames`.
- `Manipulator` has caches for `chuckingDirection` and `gripperJointNames` independent from those in `ManipulatorInfo`. Those in `Manipulator` are determined both by `GripperInfo` and `ManipulatorInfo`.
- Keep code in `ManipulatorInfo` for backward compatibility purpose.

# Note for backward compatibility
- In the downstream repo, I already implemented migration. So, in many cases, we don't need the backward compatibility code and `GripperInfo` should have `chuckingDirection`/`gripperJointNames`.
- However, if we load the scenes saved by the old openrave on the latest openrave, it requires the code in `ManipulatorInfo`.
- In addition, some of the downstream code uses `AddManipulator` to define temporary manipulator, and it requires setting in `ManipulatorInfo`.
- Thus, this MR keeps the infomration in `ManipulatorInfo`. Instead, `Manipulator` class has its own caches. That way, if the openrave loads the cleanly-migrated scene (e.g. setting is only in `GripperInfo`) and saves it, it keeps clean (e.g. `ManipulatorInfo` does not have settings).
- Also, deprecate the unused API for `Manipulator` about chucking
  - Remove `SetClosingDirection` and `SetChuckingDirection`, since these seem unused.


rdiankov:
@snozawa what are the mujin repos tied to this pull request? thanks

cielavenir:
@rdiankov complete description is available at https://tiny.mujin.co.jp/fkge8

----

By the way, it should be time to ask about the company policy to pasting link (cc @ntohge), though I used shlink. @kanbouchou asked me not to paste direct links inside intranet, but I have not checked that with the CTO yet.

rdiankov:
thanks for the link. repos and branch names would have been enough, no need for links.

rdiankov:
thanks~

PR #1383, opened 2024-05-03
Save json with uri hint

snozawa:
# Summary
- Use uriHint by default.

# Description
- Without uriHint, `referenceUri` is contained into saved json file.
- It seems the current `jsonwriter.cpp` only contains all the resolved bodies into one file, so `referenceUri` seems unnecessary.
- In addition, if saved `referenceUri` is unnecessarily contained in the file,
  - When we load it by `jsonreader.cpp`, warning message is shown.
  - Other non-openrave model loader software cannot load it because referenced file does not exist.
- I added `noUriHint` option just in case. It's unnecessary right now, but in future, once we support file saving preserving the reference files, it might be necessary.


rdiankov:
Changing default option like this could affect a lot... we should have ziyan and yoshiki comment on this.

snozawa:
> Changing default option like this could affect a lot... we should have ziyan and yoshiki comment on this.

thanks for your comment!

@yoshikikanemoto @ziyan 

- I guess, the current `jsonwriter.cpp` does not require `referenceUri`. I'm not fully sure about the negative effect if we remove `referenceUri`. But, I'm not also sure there is positive effect if we have `referenceUri`.
- In my understanding, `referenceUriHint` is something like comment, and no behavioral effect for the saved json file. Is this correct?

Thanks!

PR #1384, opened 2024-05-07
Keep the grabbing state while while updateing the grabbed body 

LLipter:
When many fields inside a grabbed body are modified through the environment notifier, it is released automatically by openrave. This is because openrave will first delete the body and then add it back if the update result is `UFIR_RequireReinitialize`. The body is released when deleted from the environment in function `_InvalidateKinBodyFromEnvBodyIndex`.

Pipeline #831391


rdiankov:
Please have this change tested and Yoshiki to also take a look, just in case. Thanks

yoshikikanemoto:
@rdiankov the change looks good to me

yoshikikanemoto (approved):

rdiankov:
thanks~

PR #1385, opened 2024-05-09
Check current URI / filename are non-empty when creating body URI

rschlaikjer:
As per the existing comment, generating URIs that are only fragments results in problems elsewhere. 
Log a warning but do not generate these invalid URIs.


rdiankov:
Hmm.... vision relies on fragments.. see `targetTemplateSceneData` and `detectedBodyData`. You can also see behavior in order cycle orchestrator at `GetTargetTemplateSceneData`

Can you explain more why you need to make this change? What breaks when you don't have it?
Thanks

rschlaikjer:
This was a belt and suspenders check to prevent the issue addressed in controllercommon PR 849 - if these fragments are valid elsewhere in the system then perhaps this comment should just be removed?

PR #1388, opened 2024-05-20
Add reserve to TrajectoryBase

snozawa:
# Summary
- Add `Reserve` API to `TrajectoryBase` and implement it for `GenericTrajectory`.
# Description
- `Insert` might be called from the time-critical code path. To mitigate the risk of dynamic memory allocation inside of `Insert`, implement `Reserve` API.
- `Reserve` API has `options` flag, so that the user can decide whether to save `_vtrajdata` only or to save `_vtrajdata` and other data altogether. Other data like `_vaccumtime` might not be used as long as the user does not use `Sample`-related APIs. In such case, the user might want to skip unnecessary memory pre-allocation.


rdiankov:
@snozawa I understand the need. But in the interest of keeping the API clean and not having to worry about calling Reserve after Init is called, how about adding a second parameter to `TrajectoryBase::Init` instead, which will only allow the user to reserve the points on initialization of the trajectory?

snozawa:
> @snozawa I understand the need. But in the interest of keeping the API clean and not having to worry about calling Reserve after Init is called, how about adding a second parameter to `TrajectoryBase::Init` instead, which will only allow the user to reserve the points on initialization of the trajectory?

thanks for your comment!
That makes sense!
I updated `Init` function to support reserving of memory, and by default, reserving is disabled.

src/libopenrave-core/generictrajectory.cpp
```diff
@@ -217,7 +217,7 @@ class GenericTrajectory : public TrajectoryBase
         return it1->second < it2->second;
     }
 
-    void Init(const ConfigurationSpecification& spec) override
+    void Init(const ConfigurationSpecification& spec, const int nWayPointsToReserve=0, const int option=0) override
```
rdiankov, line 220: @snozawa Can we add an enum to this option?

snozawa: sure!
I added the enum.
Also, update the docstring for `CloningOptions` to guide to use the existing enum.

rdiankov:
Thanks!

PR #1389, opened 2024-05-21
optimize usages of _setGrabberLinkIndicesToIgnore

yoshikikanemoto:
Found that `Environment::UpdatePublishedBodies` was spending long time in `GetGrabbedInfo` because of the inefficient usage of `_setGrabberLinkIndicesToIgnore`.

before:
```python
for link in links:
     if link.GetIndex() in _setGrabberLinkIndicesToIgnore:
         # do something to ignored link
```

after:
```python
for linkIndex in _setGrabberLinkIndicesToIgnore:
     if linkIndex < len(links):
         link = links[linkIndex]
         # do something to ignored link
    else:
        log.warning('something is wrong!')
```


rdiankov:
Thanks~ To tell you the truth, I'd like to get rid of all our usages of `std::set`...

PR #1393, opened 2024-06-21
WIP: Relax mpmath eps

cielavenir:
Sorry, https://github.com/rdiankov/collada_robots/blob/master/kuka-kr5-r650.zae requires 1e-10...


cielavenir:
this is required only when we update sympy

rdiankov:
thanks~

PR #1394, opened 2024-06-24
Bump version as API changes were added

lfye:
This PR bumps the version and updates the changelog. This is required since API changes were added in commit: `f0f27064e9ca3ef30c441ab2b863d39c29724bf6`. Without bumping the version, dependent packages could encounter compilation issues such as the following if the new API is used:
```
error: no match for ‘operator=’ (operand types are ‘OpenRAVE::geometry::RaveAxisAlignedBox<float>’ and ‘OpenRAVE::AABB’ {aka ‘OpenRAVE::geometry::RaveAxisAlignedBox<double>’})
  180 |     OpenRAVE::geometry::RaveAxisAlignedBox<float> abLocalTarget; abLocalTarget = openravepy::ExtractAABB(oabLocalTarget);
```


yoshikikanemoto:
I think the change on RaveAxisAlignedBox / RaveOrientedBox wasn't ABI breaking so I asked @lfye to make it patch-version increment.

rdiankov:
thanks

PR #1396, opened 2024-07-02
Add TrajectorySerializeOptions

cielavenir:


cielavenir:
we are not editing virtual members so minor can be the same

rdiankov:
thanks~

PR #1401, opened 2024-07-11
Fix ci.yaml's set_e

cielavenir:
`set -e` does not work for things like `ninja -j4 && ninja install`. Need to split the command line.

https://github.com/rdiankov/openrave/actions/runs/9885745876 should fail in build stage, not test stage.


cielavenir:
@felixvd @hemangandhi could you test this?

hemangandhi:
> @felixvd @hemangandhi could you test this?

https://github.com/rdiankov/openrave/actions/runs/9889187390 has it, https://github.com/rdiankov/openrave/actions/runs/9889146119 doesn't. Let's see what happens (I might've also fixed the build T_T)

hemangandhi:
> > @felixvd @hemangandhi could you test this?
> 
> https://github.com/rdiankov/openrave/actions/runs/9889187390 has it, https://github.com/rdiankov/openrave/actions/runs/9889146119 doesn't. Let's see what happens (I might've also fixed the build T_T)

This worked :tada:

PR #1405, opened 2024-07-24
Fix UB in grabbed bodies deserialization

kashitsyn:


kashitsyn:
There is another issue in the same function, IDs of `_vGrabbedInfos`'s objects are always blank strings. Even though we set id it is not persisted anywhere and getting reset before the next call.

justism:
@ziyan @rdiankov 
As @kashitsyn mentioned, grabbed info `id` are not being persisted. Should this be addressed as well?  
And if so, should the grabbed `id` be a uniquely generated `id` for each grab relationship or should it be the body `id` of the grabbed body?  
The logic within `KinBody::KinBodyInfo::DeserializeJSON` seems to expect the former.

rdiankov:
Thanks guys~ `id` should be uniquely generated for each grab relationship.
Currently openrave doesn't use the grab `id` for anything besides deserialization, so has not been an important thing for us to look at

PR #1407, opened 2024-07-31
Encode string readables with NUL bytes correctly

rschlaikjer:
Also adds constructor/SetData methods to StringReadable that take rvalue references to minimize copying


rdiankov:
I can merge this as same timing as !1409 since it requires a MINOR version update

rdiankov:
thanks~

PR #1408, opened 2024-08-01
Use unordered_set instead of vector for testing uniqueness

rschlaikjer:
Use unordered set instead of vector to track which body indices have been used.
Repeated linear scan for environments with nontrivial body counts is extremely slow.

### DIFF
--- a/src/libopenrave-core/environment-core.h
+++ b/src/libopenrave-core/environment-core.h
@@ -2970,7 +2970,9 @@ public:
                 }
             }
         }
-        std::vector<int> vUsedBodyIndices; // used indices of vBodies
+
+        // Set of indices that have already been used for vBodies
+        std::unordered_set<int> usedBodyIndexSet;
 
         // internally manipulates _vecbodies using _AddKinBody/_AddRobot/_RemoveKinBodyFromIterator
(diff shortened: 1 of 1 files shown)


rdiankov:
How about using an ordered vector itself with std::lower_bound? Should have better memory footprint

cielavenir:
@rdiankov set/vector-lower_bound: O(logN), unordered_set: O(1)

rschlaikjer:
Yeah, as Taiju says, binary search is still slower and I worry that the overhead of keeping the vector sorted (and so doing a lot of memmove if there are unordered inserts) will further degrade performance

cielavenir (approved):

rdiankov:
O(1), but memory allocation occurs. Have you done a comparison with real data?

rschlaikjer:
Loading 7k bodies:
unordered_set: 489ms
sorted vector + lower_bound: 767ms

rdiankov:
What if the vector has reserved memory for 7k items?

rschlaikjer:
With a reserve of bodyInfos size, 754ms.

rdiankov:
Got it, thanks

PR #1409, opened 2024-08-01
When removing a body, only test attached bodies for grabs

rschlaikjer:
When removing a body from the environment, we check whether any bodies were still grabbing it and, if so, ungrab it before deleting it. However, we scan the entire set of bodies in the environment to check if they are grabbing, which is a lot of overhead for environments with a lot of tracked bodies.
Since all grab relations seem to also count as attached body relations, limit the search for grab partners to only the set of bodies that are directly attached to the body to be removed.

### DIFF
--- a/src/libopenrave-core/environment-core.h
+++ b/src/libopenrave-core/environment-core.h
@@ -3489,11 +3489,15 @@ protected:
         }
         KinBody& body = *pbodyref;
         const std::string& name = body.GetName();
-        // before deleting, make sure no other bodies are grabbing it!!
-        for (KinBodyPtr& potherbody : _vecbodies) {
-            if( !!potherbody && potherbody->IsGrabbing(body) ) {
-                RAVELOG_WARN_FORMAT("env=%s, remove %s already grabbed by body %s!", GetNameId()%body.GetName()%potherbody->GetName());
-                potherbody->Release(body);
+
+        // Before deleting a body, ensure that no other bodies are still grabbing it.
+        // Since all grabbing bodies should be in the attached set for this body, we can localize the search to these bodies.
+        // Otherwise we have to scan the entire environment, which is slow for scenes with many bodies in them
+        for (KinBodyWeakPtr& pWeakOtherBody : body._listAttachedBodies) {
+            KinBodyPtr pOtherBody = pWeakOtherBody.lock();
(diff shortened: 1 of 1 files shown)


rdiankov:
Trying to understand the intent here: "Directly attached" as opposed to regular "Attached" because "this" won't be part of the returned pointers? Or because it is the first level of attachment?

rschlaikjer:
Because of the first level of attachment - if I am correct in thinking that all grabs have an associated direct attachment, then there's no point doing extra work to gather / search all the indirectly attached bodies.

src/libopenrave/kinbody.cpp
```diff
@@ -5151,6 +5151,37 @@ bool KinBody::IsAttached(const KinBody &body) const
...
+    for (const KinBodyWeakPtr& weakAttachedBody : _listAttachedBodies) {
+        // Ignore any bodies that are no longer live
+        KinBodyPtr attachedBody = weakAttachedBody.lock();
+        if (!attachedBody) {
+            continue;
+        }
+
+        // Ignore any bodies that are already present in our visited set
+        const KinBody* const kinBodyRawPtr = attachedBody.get();
+        if (visitedBodies.find(kinBodyRawPtr) != visitedBodies.end()) {
```
rdiankov, line 5173: @rschlaikjer isn't `_listAttachedBodies` supposed to be guaranteed to be unique bodies? Why do you check for duplicates?

Also, why not create:
```
void KinBody::GetDirectlyAttachedEnvironmentBodyIndices(std::vector<int>& vAttached) const
```

so that body indices can be returned instead of pointers? Should be faster, and the environment itself can get the pointer immediately by indexing `_vecbodies`

rschlaikjer: The comment on that field claims that it _can_ have duplicates; is that incorrect?
```
d071d0458a (rdiankov                2009-12-23 03:40:10 +0000)|  3685     std::list<KinBodyWeakPtr> _listAttachedBodies; ///< list of bodies that are directly attached to this body (can have duplicates)                                      ```

Re: methods, I forgot the env can just access `_listAttachedBodies`, so I have changed the code to iterate that directly rather than scanning / creating a new vector.

rdiankov:
much better, thanks~

PR #1410, opened 2024-08-02
Deduplicate body names using unordered set

rschlaikjer:
Significantly faster for high-body-count scenes - 

Before:
```
2024-08-02 11:39:01,380 openrave [INFO] [jsonreader.cpp:367 JSONReader::ExtractAll] Dedupe time for 7131 bodies: 108'424us
```

After:
```
2024-08-02 11:34:31,265 openrave [INFO] [jsonreader.cpp:367 JSONReader::ExtractAll] Dedupe time for 7131 bodies: 832us
```

### DIFF
--- a/src/libopenrave-core/jsonparser/jsonreader.cpp
+++ b/src/libopenrave-core/jsonparser/jsonreader.cpp
@@ -343,23 +343,20 @@ public:
         }
 
         // have to remove any duplicate names, prioritize ones that have higher index
-        for(int iBody = 0; iBody+1 < (int) envInfo._vBodyInfos.size(); ++iBody) {
-            bool bFoundLast = false;
-            int iTest = envInfo._vBodyInfos.size()-1;
-            while(iTest > iBody) {
-                const std::string& bodyname = envInfo._vBodyInfos[iBody]->_name;
-                if( envInfo._vBodyInfos[iTest]->_name == bodyname ) {
-                    if( !bFoundLast ) {
-                        RAVELOG_WARN_FORMAT("env=%d, remove redundant entry %d and replace with %d with body name '%s'", _penv->GetId()%iBody%iTest%bodyname);
-                        envInfo._vBodyInfos[iBody] = envInfo._vBodyInfos[iTest];
-                        bFoundLast = true;
-                    }
-                    else {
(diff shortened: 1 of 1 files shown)


src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -343,23 +343,20 @@ class JSONReader
...
+                // If this is the first time we've seen this body, add the name to the existing name set and continue
+                if (existingBodyNames.find(bodyName) == existingBodyNames.end()) {
+                    existingBodyNames.emplace(bodyName);
+                    continue;
                 }
-                --iTest;
+
+                // If this body duplicates a body with a higher index, remove it
+                RAVELOG_WARN_FORMAT("env=%d, remove redundant entry %d with body name '%s'", _penv->GetId()%iBody%bodyName);
+                envInfo._vBodyInfos.erase(envInfo._vBodyInfos.begin() + iBody);
```
rdiankov, line 359: I wonder if we can speed this up by:
```
std::swap(envInfo._vBodyInfos[iBody], envInfo._vBodyInfos[iValidSize]);
iValidSize--;
```

where iValidSize starts out as `envInfo._vBodyInfos.size()`

Then when the for loop ends, we can resize `envInfo._vBodyInfos` to `iValidSize`

rschlaikjer: Yeah, that should be faster. The duplicated case is I think usually a cold path but doesn't hurt.

rdiankov:
thanks

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -344,7 +344,7 @@ class JSONReader
 
         // have to remove any duplicate names, prioritize ones that have higher index
         {
-            std::unordered_set<std::string> existingBodyNames;
+            boost::unordered_set<string_view> existingBodyNames;
```
cielavenir: @rschlaikjer sorry that I missed this, but why don't we use `std::unordered_set<OpenRAVE::string_view>`?

cc @ntohge

cielavenir: resolved by adding std::hash, https://github.com/rdiankov/openrave/pull/1412

PR #1411, opened 2024-08-07
Two pass ungrab to avoid modification during iteration

rschlaikjer:


rdiankov:
Oops, have to watch out for that 😅

PR #1416, opened 2024-08-15
Fix CollisionPairInfo in CollisionReport

snozawa:
# Summary
- Fix bugs in `CollisionPairInfo` introduced in 9c311757e4a7f8f24c418326a3287695836c3d04
  - Fix typo : https://github.com/rdiankov/openrave/pull/1416/commits/f5a19f980fd0b236810705d7cf2f32548e539e0c
  - Fix wrong index : https://github.com/rdiankov/openrave/pull/1416/commits/35466196d39a936f82ae6eb655db8993ed92110a
- For example, the ikfast plugins are not possible to properly check the collision pairs: https://github.com/rdiankov/openrave/blob/production/plugins/ikfastsolvers/ikfastsolver.cpp#L552-L567

# Question
- Do we really need the following fallback code? @rdiankov 
https://github.com/rdiankov/openrave/blob/production/src/libopenrave/collisionchecker.cpp#L122
- If `bodyLinkGeom2Name` is not empty (e.g. somehow initialized) but does not contain ` `, it sounds irregular and I wonder if we can throw here.
- Same for other codes in this file.


kanbouchou:
Looks good to me.

rdiankov:
Thanks~
Shunichi -> If you want to add asserts instead of fallback, go ahead and see if our system works correctly.

PR #1417, opened 2024-08-15
Improve collision report geom name and fcl debug

snozawa:
# Summary
- Fix collision report to show geometry name more.
- Add `GetEnvBodiesInfoFromFCLSpace` command to `fclcollision` for debugging purpose.
- Strictly check the format of `CollisionLinkPair`, which is discussed in   https://github.com/rdiankov/openrave/pull/1416#issuecomment-2291368746

# issues and resolutions
- Fix collision report to show geometry name more.
  - `CollisionReport` does not tell the geometry name if it's coming from `extraGeometries`: https://github.com/rdiankov/openrave/blob/production/plugins/fclrave/fclspace.cpp#L109
    - According to the code path, the available geometry representation seems different : `Geometry` or `GeometryInfo`.
    - Update `CollisionReport` to take geometry name as `std::string` to cover both cases.
 - Add `GetEnvBodiesInfoFromFCLSpace` command to `fclcollision` for debugging purpose.
    - It returns the internal information of `fclspace` to write unit tests.
- Strictly check the format of `CollisionLinkPair`
   - Now `bodyLinkGeomXName` accepts empty string or string contains two white spaces as separators.
- Other improvements (remove unused function, print group name, ...etc)


plugins/fclrave/fclcollision.cpp
```diff
@@ -1328,4 +1299,55 @@ void FCLCollisionChecker::_PrintCollisionManagerInstanceLE(const KinBody::Link&
...
+            }
+            OpenRAVE::orjson::SetJsonValueByKey(rLink, "geometries", rGeometries, docInfo.GetAllocator());
+            rLinks.PushBack(rLink, docInfo.GetAllocator());
+        }
+        OpenRAVE::orjson::SetJsonValueByKey(rBody, "links", rLinks, docInfo.GetAllocator());
+        rBodies.PushBack(rBody, docInfo.GetAllocator());
+    }
+    OpenRAVE::orjson::SetJsonValueByKey(docInfo, "bodies", rBodies, docInfo.GetAllocator());
+    sout << OpenRAVE::orjson::DumpJson(docInfo);
+    return !!sinput;
```
kanbouchou: shouldn't this always return true?

plugins/fclrave/fclspace.cpp
```diff
@@ -106,7 +102,11 @@ void FCLSpace::ReloadKinBodyLinks(KinBodyConstPtr pbody, FCLKinBodyInfoPtr pinfo
                 // or, should the collision report store names of body, link, and geom?
```
kanbouchou: I think now we can delete this comment about having no geometry name.

plugins/fclrave/fclcollision.cpp
snozawa: oh, right!
I fixed it.

plugins/fclrave/fclspace.cpp
snozawa: right, i removed resolved todo comment.
meanwhile, i keep the unresolved comment about putting geometry group information somewhere.

rdiankov:
@rschlaikjer can you also take a look at this pull request? Thanks

plugins/fclrave/fclspace.cpp
```diff
@@ -152,10 +150,11 @@ void FCLSpace::ReloadKinBodyLinks(KinBodyConstPtr pbody, FCLKinBodyInfoPtr pinfo
                     enclosingBV += ConvertAABBToFcl(_tmpgeometry.ComputeAABB(Transform()));
                 }
             }
+            linkinfo->bFromExtraGeometries = false;
         }
 
         if( linkinfo->vgeoms.size() == 0 ) {
-            RAVELOG_DEBUG_FORMAT("env=%s, Initializing body '%s' (index=%d) link '%s' with 0 geometries (env %d) (userdatakey %s)", _penv->GetNameId()%pbody->GetName()%pbody->GetEnvironmentBodyIndex()%plink->GetName()%_penv->GetId()%_userdatakey);
+            RAVELOG_DEBUG_FORMAT("env=%s, Initializing body '%s' (index=%d) link '%s' with 0 geometries (env %d) (userdatakey %s, group=%s, bodygroup=%s)", _penv->GetNameId()%pbody->GetName()%pbody->GetEnvironmentBodyIndex()%plink->GetName()%_penv->GetId()%_userdatakey % _geometrygroup % pinfo->_geometrygroup);
```
rschlaikjer: No spaces around `%` in format args, you can use this pre-commit hook to do it automatically

```bash
#!/bin/bash
set -eEuo pipefail

# Get the set of all files changed in the index, excluding deletions
git diff-index --name-only --cached --diff-filter=d -z HEAD |
    grep -E -z '\.cpp$|\.h$' |
    while read -r -d '' fn
    do
        # Get the diff for that file
        DIFF=`git diff --cached HEAD -- "$fn"`

        # Get the _complete_ un-diff as well
        REVERSEDIFF=`git diff -R HEAD -- "$fn"`

        # For all additions, ensure there are no spaces around percent
        NEWDIFF=`echo "$DIFF" | sed '/^+/s/ % /%/g'`

        # Take the original file to a base state
        TEMPFILE=`mktemp`
        echo "$REVERSEDIFF" | patch "$fn" -o "$TEMPFILE"

        # Reapply the modified staged change
        NEWFILE=`echo "$NEWDIFF" | patch "$TEMPFILE" -o -`

        # Patch index
        blob_sha1=$(echo "$NEWFILE" |
                        git hash-object -w --stdin)
        printf '100644 blob %s\t%s\0' "$blob_sha1" "$fn" |
            git update-index -z --index-info

        # Clean up the tempfile
        rm -- "$TEMPFILE"
    done || true
```

plugins/fclrave/fclspace.cpp
```diff
@@ -101,12 +97,13 @@ void FCLSpace::ReloadKinBodyLinks(KinBodyConstPtr pbody, FCLKinBodyInfoPtr pinfo
                 if( !pfclgeom ) {
                     continue;
                 }
-                // because user data is null pointer here, in CheckNarrowPhaseGeomCollision report.pgeom ends up being null pointer and we lose the geometry information.
-                // maybe pass plink->GetGeometries()[index] where index=distance(vgeometryinfos.begin(), itgeominfo)?
-                // or, should the collision report store names of body, link, and geom?
-                // also, currently there is no information about which geometry group was used for collision checking.
+                // currently there is no information about which geometry group was used for collision checking.
```
rschlaikjer, line 100: Is the geometry group the `pinfo->_geometrygroup` above? If so, _should_ we save it?

src/libopenrave/collisionchecker.cpp
```diff
@@ -78,105 +78,84 @@ void CollisionPairInfo::SwapFirstSecond()
...
-    bodyLinkGeom2Name = bodyname;
-    bodyLinkGeom2Name.push_back(' ');
-    bodyLinkGeom2Name += linkname;
-    bodyLinkGeom2Name.push_back(' ');
-    bodyLinkGeom2Name += geomname;
+    _SetCollisionBodyLinkGeomName(bodyLinkGeom2Name, bodyname, linkname, geomname);
 }
 
-void CollisionPairInfo::ExtractFirstBodyName(string_view& bodyname) const
+inline void _ExtractBodyName(const std::string& bodyLinkGeomName, string_view& bodyname)
```
rschlaikjer, line 100: I'm assuming this `inline` is meant as an optimizer hint - for modern compilers it does not have this effect; `inline` semantics are to allow multiple definitions of the same function. If you _really_ want to force code inlining you can use `__attribute__((always_inline))` but generally this is best left to the optimizer.

src/libopenrave/collisionchecker.cpp
```diff
@@ -78,105 +78,84 @@ void CollisionPairInfo::SwapFirstSecond()
...
-        }
-        else {
-            bodyname = string_view(bodyLinkGeom2Name.c_str());
-        }
-    }
+    _ExtractBodyName(bodyLinkGeom2Name, bodyname);
 }
 
-void CollisionPairInfo::ExtractFirstLinkName(string_view& linkname) const
+inline void _ExtractLinkName(const std::string& bodyLinkGeomName, string_view& linkname)
```
rschlaikjer, line 120: same comment about `inline`

src/libopenrave/collisionchecker.cpp
```diff
@@ -231,21 +201,19 @@ void CollisionPairInfo::ExtractSecondBodyLinkGeomNames(string_view& bodyname, st
...
 
     size_t secondindex = bodyLinkGeomName.find_first_of(' ', firstindex+1);
-    if( secondindex == std::string::npos ) {
-        linkname = string_view(bodyLinkGeomName.c_str()+firstindex+1, bodyLinkGeomName.size()-firstindex-1);
-    }
-    else {
-        linkname = string_view(bodyLinkGeomName.c_str()+firstindex+1, secondindex - firstindex - 1);
-    }
+    OPENRAVE_ASSERT_FORMAT((secondindex != std::string::npos), "Failed to compare link since bodyLinkGeomName='%s' does not contain the second white space.", bodyLinkGeomName, ORE_Assert);
+    linkname = string_view(bodyLinkGeomName.c_str()+firstindex+1, secondindex - firstindex - 1);
```
rschlaikjer, line 216: Can this be replaced with `_ExtractLinkName`?

plugins/fclrave/fclspace.cpp
snozawa: thanks!
Fixed it!

plugins/fclrave/fclspace.cpp
snozawa: thanks for your comment!

Not sure right now, but we might want in future.

This PR fixes the geometry name since it's already introduced in `CollisionInfoPair` and there is an user of it. In that sense, we should show the geometry name unconditionally.

For `geometry group`, there is no user of it to read it via `CollisionInfoPair` and we can investigate whether it needs to show in the `CollisionInfoPair` in future.
At least, I have no strong motivation to remove this comment at this moment.

src/libopenrave/collisionchecker.cpp
snozawa: Thanks for your comment!

The functions in this file have originally `inline`.
So, I also do the same thing for the newly added function.

I'm ok to let the optimizer decide it.

src/libopenrave/collisionchecker.cpp
snozawa: same as https://github.com/rdiankov/openrave/pull/1417#discussion_r1724613138

src/libopenrave/collisionchecker.cpp
snozawa: thanks for valuable comment!

I tried to think about this, but it looks `_CompareLink` also using `firstindex` , and later, computing `bodyname` actually.
For example, I can write like this to return `firstidnex`, but when people will read this code, it's not obvious why this function returns `firstindex`, and why not for `secondindex`.

```
inline size_t _ExtractLinkName(const std::string& bodyLinkGeomName, string_view& linkname, const char* callerDescription)
{
    linkname = string_view();
    if( bodyLinkGeomName.empty() ) {
        return std::string::npos;
    }
    size_t firstindex = bodyLinkGeomName.find_first_of(' ');
    OPENRAVE_ASSERT_FORMAT((firstindex != std::string::npos), "Failed to %s since bodyLinkGeomName='%s' does not contain the first white space.", callerDescription%bodyLinkGeomName, ORE_Assert);

    size_t secondindex = bodyLinkGeomName.find_first_of(' ', firstindex+1);
    OPENRAVE_ASSERT_FORMAT((secondindex != std::string::npos), "Failed to %s since bodyLinkGeomName='%s' does not contain the second white space.", callerDescription%bodyLinkGeomName, ORE_Assert);
    linkname = string_view(bodyLinkGeomName.c_str()+firstindex+1,secondindex - firstindex - 1);
    return firstindex;
}
```

so, I'm personally fine with the original code.
what do you think?

plugins/fclrave/fclcollision.h
```diff
@@ -250,6 +247,9 @@ class FCLCollisionChecker : public OpenRAVE::CollisionCheckerBase
 
     void _PrintCollisionManagerInstanceLE(const KinBody::Link& link, FCLCollisionManagerInstance& envManager);
 
+    /// \brief get information from fcl space GetEnvBodies.
+    bool _GetEnvBodiesInfoFromFCLSpace(ostream& sout, istream& sinput);
```
rdiankov, line 251: @snozawa is this for debug purposes? Or are you going to be using it in production code? If so, where?

snozawa: thanks for comment!

This is only for debugging purpose to write more explicit unit test codes to observe the internal states of `fclcollision/fclspace` on various events, such as loading json, adding `KinBody`, removing `KinBody`, cloning stuff, ..etc.

rdiankov:
thanks~

PR #1418, opened 2024-08-15
Draft: Fix jitterer biasing flag setting

Puttichai:
### Summary

Currently there is no way to reset the flag `_busebiasing` in the jitterer once it has been set to `true`. 

When users send command `SetManipulatorBias` with `vbiasdirection = [0, 0, 0]`, the current behavior is that the jitterer leaves the flag `_busebiasing` as it is. This means that 
- if the jitterer has preivously been set with some bias direction
- later the user sets `vbiasdirection = [0, 0, 0]`

then when the jitterer is used next time, it will still use biasing since `_busebiasing` is `true` and the preivously set `_vbiasdirection` is used, which seems unexpected.

When the user sets `vbiasdirection = [0, 0, 0]`, the intention seems to be to _not_ use biasing. I think we should set the flag `_busebiasing` to `false` first and set it to true when after we confirm that manipulator exists and `vbiasdirection` is non-zero.


rdiankov:
Nice catch, thanks~

Issue #1042, opened 2022-01-25
IK generation failed for manipulator with prismatic joint and revolute joints


saikishor:
Hello, 

we are facing an issue with the OpenRave. It works like a charm for the manipulator with all revolute joints. however, when we try to generate the IK solution for the manipulator with a prismatic joint in between revolute joints. It fails in the generation showing the below error. 

```
INFO: depth=1, c=59, iter=3/4, adding newcases: set([Abs(j2 - 3.14159265358979323846264338328), Abs(j0 + 1.57079632679489661923132169164)])
INFO: depth=0, c=5, iter=3/4, adding newcases: set([Abs(j0 + 1.57079632679489661923132169164)])
INFO: generating cpp code...
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast.py", line 9522, in <module>
    code=solver.writeIkSolver(chaintree,lang=options.lang)
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast.py", line 2132, in writeIkSolver
    return CodeGenerators[lang](kinematicshash=self.kinematicshash,version=__version__).generate(chaintree)
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast_generator_cpp.py", line 517, in generate
    code += solvertree.generate(self)
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast.py", line 943, in generate
    return generator.generateChain(self)
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast_generator_cpp.py", line 704, in generateChain
    fcode += self.generateTree(node.jointtree)
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast_generator_cpp.py", line 1755, in generateTree
    code += n.generate(self)
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast.py", line 666, in generate
    return generator.generateCoeffFunction(self)
  File "/usr/lib/python2.7/dist-packages/openravepy/_openravepy_0_9/ikfast_generator_cpp.py", line 1425, in generateCoeffFunction
    assert(len(node.jointnames) == len(node.jointeval))
```

[...]

PR #1421, opened 2024-08-19
Optimize IsGrabbing / ResetGrabbed

rschlaikjer:
`ResetGrabbed` is on the hot path for syncing environment changes; store an index of grabbed info by name to avoid linear scan in `IsGrabbing` (and a lot of `weak_ptr` locks) and refactor ResetGrabbed to avoid deleting and recreating all grabs every time.

baseline:
```
2024-08-16 15:05:20,750 [...] reset grabbed 5000 bodies in 255'122us
2024-08-16 15:05:31,129 [...] reset grabbed 5000 bodies in 261'812us
2024-08-16 15:05:36,645 [...] reset grabbed 5000 bodies in 237'566us
```

with just map cache:
```
2024-08-19 11:44:43,991 [...] reset grabbed 5000 bodies in 40'754us
2024-08-19 11:44:43,663 [...] reset grabbed 5000 bodies in 42'764us
2024-08-19 11:44:43,335 [...] reset grabbed 5000 bodies in 46'285us
```

with refactored ResetGrabbed:
```
2024-08-16 14:25:46,796  [...] reset grabbed 5000 bodies in 2'302us
2024-08-16 14:25:46,970  [...] reset grabbed 5000 bodies in 2'321us
2024-08-16 14:25:47,142  [...] reset grabbed 5000 bodies in 2'385us
```

since bodies can be renamed while grabbed we have to hook `NotifyKinBodyNameChanged` to ensure cache consistency, but since grab tests are a much more frequent codepath than renames this tradeoff seems worthwhile.


include/openrave/kinbody.h
```diff
@@ -3703,6 +3705,10 @@ class OPENRAVE_API KinBody : public InterfaceBase
 
     std::vector<GrabbedPtr> _vGrabbedBodies; ///< vector of grabbed bodies
 
+    /// Quick lookup map of body name -> indexed into _vGrabbedBodies of the matching grab entry
+    using MapGrabbedBodyNameIndex = std::unordered_map<std::string, std::vector<GrabbedPtr>::size_type>;
+    MapGrabbedBodyNameIndex _mapGrabbedBodyNameIndex;
```
rdiankov: @rschlaikjer Maintaining any more members as cache than there is is already scary. Anyway we can do this by better ordering _vGrabbedBodies? Perhaps the vector can be ordered?
Also, I'm wondering if we really need `_vGrabbedBodies` to be a vector of shared pointers. If we design an efficient swap function, perhaps it'll be pretty quick to insert and/or remove elements from the sorted vector (which doesn't happen as much as querying anyway..)

rschlaikjer: IIRC initially I was just going to just turn `_vGrabbedBodies` into a map to avoid having a layer of indirection, but there exist APIs that seem to depend on the ordering of the members of `_vGrabbedBodies` (e.g. `GetGrabbedBody(int index)`), so it seemed like mutating the order could cause problems for existing code. 

Is the concern with cache the memory overhead? Or keeping it consistent?

I'm not sure how much the shared pointer to `Grabbed` is a performance issue; those are internal and generally swapped / referenced. The performance issue was locking the weak pointer to a kinbody _inside_ grabbed to fetch the name.

rdiankov: Not sure that form of the function is so important. If it means we can optimize, I would remove it.

Also, perhaps worthwhile to consider using environmentBodyIndex instead of pointers? Any grabbed body is guaranteed to be in the environment, so its index shouldn't change, and can query its pointer from the environment pretty easily.

rschlaikjer: Here's a quick version with the removal of GetGrabbedBody(int):
https://github.com/rdiankov/openrave/compare/rs-optimize-reset-grabbed-novector

It requires additional PRs (rb #2292, cc #1042) to account for the API change.

rdiankov: Okay, what was the purpose of `uintptr_t _grabbedBodyPtrId;`? I didn't see it used anywhere, and I'm also not a big fan of using pointers as Ids...

Also, why not make the key the environmentBodyIndex? Why do we need name if we can quickly use an integer to get to the Grabbed object?
And `GetGrabbed(std::vector<KinBodyPtr>& vbodies)` can also be returning a vector of environmentBodyIndex's


I think the more we move away from shared pointers, dynamically allocated std::strings, and sets/maps, the better it is for the entire systems.
If we're going to re-write, might as well do it with that in mind: how to reduce the dynamic memory allocation/object creation as much as possible.

rschlaikjer: > _grabbedBodyPtrId

oops, that's a remnant of a route I didn't go down. Dead code.

> Also, why not make the key the environmentBodyIndex? 

The inital reason was that when regrabbing we need to do a lot of looking up by name, since that's what the GrabbedInfo structs contain - if the map is based on index, then we need a double-lookup of name -> body -> env index.

What are the rules for environmentBodyIndex invalidation? Is it ever possible for a grabbed body's index to change without the grab relationship changing?

rdiankov: Why is the lookup by name necessary?  And `environmentBodyIndex` should not be changing for actively grabbed bodies.

test/test_grabs.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
+
+from common_test_openrave import *
```
cielavenir: @rschlaikjer Now `from common_test_openrave import *` is not allowed.

Well, I fixed.

cielavenir:
@rschlaikjer by the way, it is highly encouraged to add new tests to git.mujin/test/testopenrave

include/openrave/kinbody.h
rschlaikjer: We have to do a lookup by name because the GrabbedInfos only contain a name, and we want to match them to existing Grabbed structs to avoid destroying/recreating all grabs on reset grabbed (which causes a bunch of churn with attaching/detaching bodies, creating/destroying shared pointers). 

Modified the example map version to use env body index and merged into this branch. Still running tests to re-verify behaviour after the changes.

rschlaikjer:
Autotester seems to be in the usual range (pipeline 520344), tests have been moved to testopenrave 117, additional associated PRs rb 2292, cc 1042

include/openrave/kinbody.h
```diff
@@ -3586,6 +3590,8 @@ class OPENRAVE_API KinBody : public InterfaceBase
     }
 
 protected:
+    using MapGrabbedByEnvironmentIndex = std::unordered_map<size_t, GrabbedPtr>;
```
rdiankov: Why size_t? Shouldn't it be int?

rdiankov: what would be the impact if this was an ordered std::vector?

rschlaikjer: Re: performance impact, here's a test sweep:
https://gist.github.com/rschlaikjer/a20a0c846fa832f1a24805c531a2612b

Low hit rate:
2024-08-27-103532_3836x2114_scrot

High hit rate:
2024-08-27-103459_3836x2114_scrot

The sorted vector honestly does way better than I expected for search, but as expected, as N gets large you cannot escape the computational complexity of each implementation.
If we focus in on the low-N region, things are pretty noisy, but it doesn't _seem_ like there's a clear advantage to using a  vector:

2024-08-27-104358_3836x2114_scrot
2024-08-27-104327_3836x2114_scrot

rdiankov: Thanks for the experiments. Two comments:

1. Let's stick with experimenting with realistic numbers. For example, N will be at max 100
2. Your test case has an extra indirection with std vector (datum->data_id < search), which does not exist in the map case. In order to make this apples-to-apples test, please manage the vector elements as a pair, with first being the key. Also, probably best to use int as the key, since that is what the openrave code will be doing

Thanks

rschlaikjer: Interestingly, pair seems to be a lot worse - possible there's a mistake in my benchmark? For non SSO strings I wouldn't expect storing them as a pair to help with locality since the data is heaped somewhere else anyway, so you have the same dereference you'd have just looking at the Data member (I guess, one vs two derefs, so it's something). For ints... I'm not sure why it's so much worse than the others. 

https://gist.github.com/rschlaikjer/ddcf2d71458f83de9e569c92bb47a8d6

2024-08-28-091931_3836x2114_scrot
2024-08-28-091858_3836x2114_scrot

plotting - 
```R
map_data <- read.csv("map_results.csv")
vec_data <- read.csv("vec_results.csv")
vec_pair_data <- read.csv("vec_pair_results.csv")
par(mfrow=c(2,1))

plot(vec_data$haystack_size, vec_data$insert, col='blue', type='p', ylim=c(0, 1800))
lines(map_data$haystack_size, map_data$insert, col='green', type='p')
lines(vec_pair_data$haystack_size, vec_pair_data$insert, col='orange', type='p')
legend(90, 1500, legend=c("vector", "map", "vec_pair"), fill=c("blue", "green", "orange"))
title("Amortized insert time (ns) vs element count (90% hit rate, string32 key)")

plot(vec_data$haystack_size, vec_data$query, col='blue', type='p', ylim=c(0, 1200))
lines(map_data$haystack_size, map_data$query, col='green', type='p')
lines(vec_pair_data$haystack_size, vec_pair_data$insert, col='orange', type='p')
legend(90, 1200, legend=c("vector", "map", "vec_pair"), fill=c("blue", "green", "orange"))
title("Amortized query time (ns) vs element count (90% hit rate, string32 key)")
```

rdiankov: Okay, thanks for the experiments. For now, looks like unordered_map wins. Please update MINOR version and CHANGELOG to get ready for merge.

rschlaikjer: It looks like the pessimal insert is caused by all the shuffling the vector has to do - moving a pair (and associated string) is surprisingly expensive compared to just moving the pointer.

rdiankov:
thanks~

PR #1422, opened 2024-08-22
Fix: when loading environment from `{"referenceUri": "uri", "bodies": [...]}`, KinBodyInfo was merged instead of overwritten

yoshikikanemoto:
when loading scene like this, kinbody from `reference.mujin.json` would not be overwritten by kinbody from `container.mujin.json` . instead, they were merged and it may create invalid `KinBodyInfo`.

```json
{
  "referenceUri": "mujin:/reference.mujin.json",
  "bodies": [
    {
      "id": "sc1_container",
      "name": "sc1_container",
      "referenceUri": "mujin:/container.mujin.json",
      "uri": "mujin:/container.mujin.json"
    }
  ]
}
```

This PR changes this behavior from "merge" to "overwrite".

can be verified by downloading these three scenes and load `dynamic_state.mujin.json`
[container.mujin.json](https://github.com/user-attachments/files/16709195/container.mujin.json)
[reference.mujin.json](https://github.com/user-attachments/files/16709196/reference.mujin.json)
[dynamic_state.mujin.json](https://github.com/user-attachments/files/16709197/dynamic_state.mujin.json)

This error will be shown with the latest `production` branch.
```
openravepy._openravepy_.openravepy_int._OpenRAVEException: ("openrave (Assert): [virtual void OpenRAVE::KinBody::KinBodyInfo::DeserializeJSON(const Value&, OpenRAVE::dReal, int):357] Body 'my_container' has info with link[0] and link[1] having the same linkname 'base', which is not allowed. link[0].id='link01', link[1].id='my_container_link'", 'Assert')
```

AutoTester Pipeline #916128


ziyan:
@rschlaikjer Please check this, this usage pattern is only in env tracker

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -298,6 +298,36 @@ class JSONReader
...
             }
+            // if parent rEnvInfo is going to overwrite bodies in the environment of the referenceUri, do not deserialize them. otherwise, they would be merged instead of overwritten.
+            if( rEnvInfo.HasMember("bodies") && rEnvInfo["bodies"].Size() > 0 ) {
+                if( prReferenceEnvInfo->HasMember("bodies") && (*prReferenceEnvInfo)["bodies"].Size() > 0 ) {
+                    boost::shared_ptr<rapidjson::Document> prFilteredReferenceEnvInfo = boost::make_shared<rapidjson::Document>(&alloc);
+                    prFilteredReferenceEnvInfo->CopyFrom(*prReferenceEnvInfo, alloc);
+                    const rapidjson::Value& rEnvBodies = rEnvInfo["bodies"];
+                    rapidjson::Value& rRefBodies = (*prFilteredReferenceEnvInfo)["bodies"];
+                    for (rapidjson::Value::ConstValueIterator itEnvBody = rEnvBodies.Begin(); itEnvBody != rEnvBodies.End(); ++itEnvBody) {
+                        rapidjson::Value::ConstMemberIterator itEnvBodyId = itEnvBody->FindMember("id");
```
ziyan, line 314: Are we also matching name here? Or only `id`?

Should we be assigning a new `id` in an attempt to keep both bodies?

ziyan: If someone had a dynamic body with id=1, and later go into webstack and change static scene to have a body with id=1 as well, and reload the process, should both bodies be present?

yoshikikanemoto: @ziyan If we think about modifications on the referenced environment, there are a few conditions to consider. See below table. I am not certain about expected behaviors on condition 2-5. Can I get advice? @ziyan @rdiankov 


| no | condition                                                            | action                                                                                                                           |
|----|----------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------|
| 1  | name and id match                                                    | no doubt! replace the body                                                                                                       |
| 2  | name matches. but id doesn't. there is no body that has the same id  | maybe the body in reference env was replaced. replace the body?                                                                  |
| 3  | name matches. but id doesn't. there is a body that has the same id   | maybe the body in reference env was named and a new body was added with the original name. raise an error or discard the change in dynamic state? |
| 4  | id matches but name doesn't. there is no body that has the same name | maybe the body was renamed. replace the body?                                                     |
| 5  | id matches but name doesn't. there is a body that has the same name  | same as no. 3                                                                                                                       |
| 6  | nothing matches                                                      | create a new body.                                                                                                               |

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -298,6 +298,36 @@ class JSONReader
...
+                        }
+                        const char* envBodyId = itEnvBodyId->value.GetString();
+                        const size_t envBodyIdLen = itEnvBodyId->value.GetStringLength();
+                        for (rapidjson::Value::ValueIterator itRefBody = rRefBodies.Begin(); itRefBody != rRefBodies.End(); ++itRefBody) {
+                            rapidjson::Value::ConstMemberIterator itRefBodyId = itRefBody->FindMember("id");
+                            if( itRefBodyId == itRefBody->MemberEnd() || itRefBodyId->value.GetStringLength() == 0 ) {
+                                continue;
+                            }
+                            const char* refBodyId = itRefBodyId->value.GetString();
+                            if( strncmp(envBodyId, refBodyId, envBodyIdLen) == 0 ) {
```
rschlaikjer: I think this comparison will have false positives:
```
const char* envBodyId = itEnvBodyId->value.GetString(); // let's assume "myBody"
const size_t envBodyIdLen = itEnvBodyId->value.GetStringLength(); // len("myBody")
...
const char* refBodyId = itRefBodyId->value.GetString(); // let's assume "myBody2"
strncmp(envBodyId, refBodyId, envBodyIdLen) -> strncmp("myBody, "myBody2", strlen("myBody")) -> strings are equal!
```
need to check if lengths match, or just wrap both in string_views and use the compare operator on the string view instead of manually doing strncmp.

src/libopenrave-core/jsonparser/jsonreader.cpp
```diff
@@ -298,6 +298,36 @@ class JSONReader
                 }
                 return false;
             }
+            // if parent rEnvInfo is going to overwrite bodies in the environment of the referenceUri, do not deserialize them. otherwise, they would be merged instead of overwritten.
+            if( rEnvInfo.HasMember("bodies") && rEnvInfo["bodies"].Size() > 0 ) {
+                if( prReferenceEnvInfo->HasMember("bodies") && (*prReferenceEnvInfo)["bodies"].Size() > 0 ) {
+                    boost::shared_ptr<rapidjson::Document> prFilteredReferenceEnvInfo = boost::make_shared<rapidjson::Document>(&alloc);
+                    prFilteredReferenceEnvInfo->CopyFrom(*prReferenceEnvInfo, alloc);
```
rschlaikjer, line 305: Might be nice if we could avoid copies, though I guess reference environments should generally be smaller? Erasing the elements from the original list seems slow (rapidjson `Erase` has to `memmove` all subsequent), but maybe just `SetNull` on the items to filter and skip non-objects when processing?

src/libopenrave-core/jsonparser/jsonreader.cpp
yoshikikanemoto: oops. thanks. fixed.

src/libopenrave-core/jsonparser/jsonreader.cpp
yoshikikanemoto: We need to keep `prReferenceEnvInfo` untouched since it is cached. We can think of making shallow copy of a rapidjson object to avoid deepcopy for speed up but I think it would be a little big hustle and I am not sure if it is worth an effort as this procedure currently happens only on startup and reference environments should be compact compared to dynamic scene state with mobile robots as you mentioned.
`SetNull` instead of `Erase` requires many other places to be changed. I guess we don't want that only for this also.

src/libopenrave-core/jsonparser/jsonreader.cpp
yoshikikanemoto: updated the PR according to the above table for now.

ziyan: in 1, "replace the body", is this the right thing to do? @rschlaikjer what's the behavior right now if we use `NotifyEnvironmentChanged` to modify a body that exists in the static scene to give it extra readable interface and persist this in the dynamic scene? 

in my opinion, if `id` exists, it should always take priority, `name` is just a property that can be modified by the JSON to be applied in this case. Only if `id` does not exist, or is empty, OpenRAVE has this fallback behavior of using `name` in its place. This behavior does not exist in webstack. Doing all 6 cases is super confusing.

yoshikikanemoto: @ziyan `NotifyEnvironmentChanged`  does not use this code path and it applies update upon an existing body in the current environment as it does not have 'referenceUri`.

The above table is taking `id` priority. In addition, in case `id` does not match, it falls back to use `name` if possible (no 2). and if prioritizing `id` leads to generate duplicated `name`s, raises an error. (no 3 and 5).

rschlaikjer: Id matches, name matches: readable is added
https://pastebin.mujin.co.jp/?f72214fe53d642fb#6a2pGqRvRWh4sBt3cQHB39vkLwrnkarMQ58SGuhsXC1e

Id matches, name different: body is renamed, readable is added
https://pastebin.mujin.co.jp/?52c64fd37cd60439#FdXpwr8DkBUsDPH7Bhduhr8na9pQppb6h39HNjQBFvva

Id doesn't match, name matches: id is changed (??), readable is added
https://pastebin.mujin.co.jp/?d1c0b49fd4988e9a#2z4MsGqD1K8AFFod3ZUwLGHYgU342guX22cJqCrsaYCW

Id doesn't match, name doesn't match: adds new body
https://pastebin.mujin.co.jp/?091d610e9d8fa804#9s8LMJekJDNFvs2VehGv7G72aknboaNbCMF1HCmeaTbD

yoshikikanemoto: @ziyan Any more comments?

rdiankov:
@yoshikikanemoto @cielavenir please design a test for this and show that this does not have any regressions with the rest of our system. thanks

cielavenir:
cc @ntohge

(why he is not added to openrave yet? i have asked for years, really)

rdiankov:
> cc @ntohge
> 
> (why he is not added to openrave yet? i have asked for years, really)

Oh, is "ntohge" Norihito's gitlab handle? If so, I don't have any email record of his request. Please have Norihito send me an email asking for this along with his github handle. Thanks

src/libopenrave-core/jsonparser/jsonreader.cpp
ziyan: @yoshikikanemoto No more comments, thank you!

ziyan:
@rdiankov LGTM.

rdiankov:
thanks!

PR #1424, opened 2024-09-06
Init variables, check for valid file pointers before using, provide valid return values

lazydroid:
This MR fixes minor issues in the using uninitialized variables, returning nullptr instead of std::string, using file pointers without checking them for validity and similar things. This is not a complete list of issues, but the most dangerously looking ones.


rdiankov:
Can you have Puttichai look at the changes related to planning? Thanks

src/libopenrave-core/colladaparser/colladareader.cpp
```diff
@@ -296,7 +296,7 @@ class ColladaReader : public daeErrorHandler
         if (find(_vOpenRAVESchemeAliases.begin(), _vOpenRAVESchemeAliases.end(), urioriginal.scheme()) !=
             _vOpenRAVESchemeAliases.end()) {
             if (urioriginal.path().empty()) {
-                return nullptr;
+                return std::string();
```
cielavenir, line 299: thank you for this change, so this tried to construct std::string from c-string, which is not a compile error but will cause a runtime error.

cielavenir (approved):

plugins/configurationcache/configurationcachetree.cpp
```diff
@@ -828,6 +828,9 @@ int CacheTree::SaveCache(std::string filename)
 
     FILE* pfile;
     pfile = fopen(_fulldirname.c_str(),"wb");
+    if( pfile == NULL) {
+        return 0;
```
Puttichai, line 832: Could you help add a mention of what return value means in the header (`configurationcachetree.h`)? Thanks.

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
```diff
@@ -123,7 +123,7 @@ bool PPRamp::SolveMinTime2(Real amax,Real timeLowerBound)
     }
     _a2 = -_a1;
     PARABOLIC_RAMP_ASSERT(ttotal >= timeLowerBound);
-    Real ts1,ts2;
+    Real ts1 = 0.0, ts2 = 0.0;
```
Puttichai: In this context, I think it's better to initialize variables that represent time instants with `-1.0` rather than `0.0` to indicate that they are invalid.

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
```diff
@@ -352,7 +352,7 @@ Real PPRamp::CalcTotalTime(Real a) const
 
 int PPRamp::CalcTotalTimes(Real a,Real& t1,Real& t2) const
 {
-    Real ts1,ts2;
+    Real ts1 = 0.0, ts2 = 0.0;
```
Puttichai: Here too.

plugins/configurationcache/configurationcachetree.cpp
lazydroid: done.

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
lazydroid: what do you think about line 432 -- this was made before me:
```
Real PPRamp::CalcSwitchTime(Real a) const
{
    Real t1 = 0.0, t2 = 0.0;
    int res = CalcSwitchTimes(a,t1,t2);
```

src/libopenrave-core/colladaparser/colladareader.cpp
lazydroid: thank you!

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
cielavenir: @Puttichai We need to clarify: CalcSwitchTime one has committed by LiBoon in 7a528ba721. Should that be fixed to -1.0?

src/libopenrave/planningutils.cpp
```diff
@@ -4198,7 +4198,7 @@ int DynamicsCollisionConstraint::Check(const std::vector<dReal>& q0, const std::
                         BOOST_ASSERT(0);
                     } // end switch maskinterpolation
                     bool bFoundTimeInstant = false;
-                    dReal root;
```
lazydroid, line 4201: @Puttichai could you please provide a good default for this variable, because it's not assigned, still used for comparison within the loop (?), then conditionally assigned inside the loop and once again used after the loop (where it still may be non-initialized, because the loop was empty or the conditions were always false.

plugins/rplanners/ParabolicPathSmooth/ParabolicRamp.cpp
```diff
@@ -2898,7 +2902,7 @@ Real SolveMinTimeBounded(const Vector& x0,const Vector& v0,const Vector& x1,cons
                     solved = false;
                     if( 1 ) { // disable to avoid changing time
                         PARABOLIC_RAMP_PLOG("ParabolicRamp1D info: x0 = %.15e; x1 = %.15e; v0 = %.15e; v1 = %.15e; vm = %.15e; am = %.15e; endTime = %.15e", x0[i], x1[i], v0[i], v1[i], vmax[i], amax[i], endTime);
-                        Real newEndTime;
+                        Real newEndTime = 0.0;
```
lazydroid, line 2905: @Puttichai could you please take a look, what would be the proper default value? there are code paths (ramps.size() == 0) that skip assigning the value, but it's used later in calculations.

plugins/rplanners/ParabolicPathSmooth/ParabolicRamp.cpp
```diff
@@ -3147,7 +3151,7 @@ bool SolveAccelBounded(const Vector& x0,const Vector& v0,const Vector& x1,const
                 PARABOLIC_RAMP_PLOG("ParabolicRamp1D info: x0 = %.15e; x1 = %.15e; v0 = %.15e; v1 = %.15e; vm = %.15e; am = %.15e; newDuration = %.15e", x0[i], x1[i], v0[i], v1[i], vmax[i], amax[i], endTime);
 
                 if (numDilationTries > 0) {
-                    Real newEndTime;
+                    Real newEndTime = 0.0;
```
lazydroid, line 3154: @Puttichai same question here, code very similar, like copy-pasted from above.

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
```diff
@@ -254,7 +254,7 @@ bool PPRamp::SolveFixedTime(Real amax,Real endTime)
 
 bool PPRamp::SolveMinAccel(Real endTime)
 {
-    Real switch1,switch2;
+    Real switch1 = 0.0,switch2 = 0.0;
```
lazydroid, line 257: @Puttichai apparently, `CalcMinAccel(endTime,1.0,switch1)` has code paths that do not set `switch1` to any value. Shall I use 0.0 or -1.0, as we did for `CalcSwitchTimes()` ?

plugins/rplanners/ParabolicPathSmooth/DynamicPath.cpp
```diff
@@ -652,7 +652,7 @@ bool DynamicPath::TryShortcut(Real t1,Real t2,RampFeasibilityChecker& check)
     if(t1 > t2) {
         Swap(t1,t2);
     }
-    Real u1,u2;
+    Real u1 = 0.0, u2 = 0.0;
```
lazydroid: @Puttichai same question here, `GetSegment()` can abort early without writing to the result, what default values would you prefer here?

lazydroid:
There's a conditional passing of NULL pointer with null pointer dereference, do you have any idea why do we need to convert vectors to raw pointers instead of just iterating over the vectors with the approximately same result?

image

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
Puttichai: I checked `res = CalcSwitchTimes(a, t1, t2)`. The only case when `t1` and `t2` are not set to any values inside the function is when the return value `res` is `0`, meaning that there are no solutions.
I still think it makes sense to change the default values to `-1`.

src/libopenrave/planningutils.cpp
Puttichai: `root` is used only when `bFoundTimeInstant` is `true`. `root` is assigned to a value at the same time `bFoundTimeInstant` is set to `true`. So I think it's fine.

plugins/rplanners/ParabolicPathSmooth/ParabolicRamp.cpp
Puttichai: The entire section is inside the outer for-loop `for(size_t i=0; i<ramps.size(); i++)`, so I think this is fine too.

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
Puttichai: I'd say a correct fix would be to do this
```diff
--- a/plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
+++ b/plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
@@ -480,6 +480,9 @@ Real PPRamp::CalcMinAccel(Real endTime,Real sign,Real& switchTime) const
     Real c = (dx1 - dx0)*endTime;
     Real rat1,rat2;
     int res=SolveQuadratic(a,b,c,rat1,rat2);
+    if (res == 0) {
+        return -1.0;
+    }
     Real accel1 = (dx1-dx0)/rat1;
```

When `res > 0`, `rat1` and `rat2` are initialized.

plugins/rplanners/ParabolicPathSmooth/DynamicPath.cpp
Puttichai: Let's assign them to `-1.0` to say they are invalid.

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
lazydroid: done

plugins/rplanners/ParabolicPathSmooth/DynamicPath.cpp
lazydroid: done.

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
lazydroid: changed to -1

plugins/rplanners/ParabolicPathSmooth/ppramp.cpp
lazydroid: changed to -1

src/libopenrave/planningutils.cpp
```diff
@@ -4198,7 +4198,7 @@ int DynamicsCollisionConstraint::Check(const std::vector<dReal>& q0, const std::
...
+                    dReal root = 1234567890;
```
Puttichai:
```suggestion
                    dReal root = 0;
```

Puttichai (approved):
Looks good to me now.

lazydroid:
> Looks good to me now.

Thank you!

rdiankov:
thanks~

PR #1425, opened 2024-09-08
WIP: Fix grabbed bodies saver

snozawa:
# Summary
- Fix several bugs in kinbody state saver with `Save_GrabbedBodies` option.

# Bugs 1
- Fix typo. Otherwise, restoring toward cloned env does not correctly work. https://github.com/rdiankov/openrave/pull/1425/commits/684daec466ad53e09e17e3d446e3565d74101ae6

# Bugs 2
- Fix typo, which is newly introduced in https://github.com/rdiankov/openrave/pull/1425/commits/cbd375ccf505af64950798e54a4c3087f22e56e0

# Bugs 3
- Restoring code is duplicated between `KinBodyStateSaver` and `KinBodyStateSaverRef`.
- Apply the fixes introduced in `KinBodyStateSaver` in the past to `KinBodyStateSaverRef`.
  - https://github.com/rdiankov/openrave/pull/1425/commits/4c3ae1927af5fcc99c2c92adef277ab3e4182a35
  - https://github.com/rdiankov/openrave/pull/1425/commits/bc9f2e1006590a8221cceed3845cc5a6658219c9
  - https://github.com/rdiankov/openrave/pull/1425/commits/d05e69a284762e1d7af4fc9b9775d69d00ec44ae

# Bugs 4
- In the previous code, `_listNonCollidingLinksWhenGrabbed` does not restore the other grabbed bodies at all.
- Fixed via : https://github.com/rdiankov/openrave/pull/1425/commits/d9aecc5360862b7d9298f20c395ba94cb9be5370
- Not fully sure the above fix is ok in terms of performance.

# Bugs 5
- In `ReleaseAllWithLink` API, other grabbed bodies are not post processed, while `Release` API postprocesses it by calling `_RemoveGrabbedBodies`. Thus, unused `_listNonCollidingLinkWhenGrabbed` remains.
- Fixed via : https://github.com/rdiankov/openrave/pull/1425/commits/1fbd14a7115dfca682f99ca953ec1cee8686cb14

# Bugs 6
Fix the bugs of ResetGrabbed, introduced in https://github.com/rdiankov/openrave/pull/1421/.
- Issue
  1. Members of Grabbed is overwritten in ResetGrabbed, such as trelative. However, pGrabbedSaver and pGrabberSaver are initialized in the constructor and these are not consistent with the newly set tRelative.
  2. The pointer of Grabbed is held by KinBodyStateSaver. If ResetGrabbed overwrites the contents of Grabbed instance, the saved information gets changed.
- Resolution
  - Revert some of the change in https://github.com/rdiankov/openrave/pull/1421, Thanks for @rschlaikjer!
  - In detail, ResetGrabbed creates the new Grabbed instances for consistency. As a background
     - To resolve the issue 2, KinBodyStateSaver might be able to have copy of Grabbed. But, it might be more expensive than having the new Grabbed in ResetGrabbed. Based on such trade-off, this commit creates the new Grabbed in Reset
Grabbed.
     - To resolve the issue 1, ResetGrabbed might be able to update the pGrabbedSaver and pGrabberSaver. But, it still allocates memories and creating the new Grabbed would be simpler.
- fixed via  b9a02dfb6d5472ef1d99416df1226f6e9592a329

# Bugs 7
- Fix bugs that KinBodyStateSaver does not restore some of the Grabbed's members.
- Issue
  - When KinBody::Grab, KinBody::Release, ...etc are called, new Grabbed instance is created in KinBody and the original ptr for the original Grabbed instance is swapped.
  - Thus, the original information of Grabbed instance is unchanged. Holding the ptr of it as pGrabbed is enough to store the saved data for KinBodyStateSaver.
  - However, some of the grab-related API mutates the internal states of Grabbed instance, _listNonCollidingIsValid, _listNonCollidingLinksWhenGrabbed, and _setGrabberLinkIndicesToIgnore._
  - This might cause the unexpected deleting of grabbed bodies in the _listNonCollidingLinksWhenGrabbed, e.g. this might cause unexpected self colliding.
- Resolution
  - Introduce the `SavedGrabbedData`. This holds the `GrabbedPtr` same as before. In addition, this holds the copied data for _listNonCollidingIsValid, _listNonCollidingLinksWhenGrabbed, and _setGrabberLinkIndicesToIgnore.
- Fixed via 193fa6891804d736a69acda2a1de0eb3906d686e and ef73977a30603157ead5bb0c4281ed43ea774306

# Bug 8
- Related fix to Bug 7.
- Issue
  - Actually, the meaning of `Save_GrabbedBodies` were not clear in the previous code.
    - KinBodyStateSaver holds Grabbed pointers, and Grabbed uses KinBodyStateSaver with Save_GrabbedBodies. Therefore, the relationship is cyclic.
    - It looks `Save_GrabbedBodies` of KinBodyStateSaver wants to restore any data saved before the saver constructor.
    - However, the `Save_GrabbedBodies` used in `Grabbed::ComputeListNonCollidingLinks` wants to restore some of the data, but wants to keep others like `_listNonCollidingLinksWhenGrabbed`.
  - Resolving of this issue is also required for Bug 7.
- Resolution
  - Avoid cyclic relationships. Do not use `Save_GrabbedBodies` in `Grabbed::ComputeListNonCollidingLinks`.
  - Instead, call link-link collision checking APIs. This method is actually used in `KinBody::CheckSelfCollision` to achieve the collision checking with the attached bodies.
- Fixed via 7829f20e33bfb7764961984251d0ef809c0d0a22 and 267ec539d04534d041caaa5ebcdbb87166040c9d

# Bugs 9
- Fix the potential bug in `Restore` to the kinbody in the other env with `Save_GrabbedBodies` option.
- Issue
   - As well described in the existing comment,  restoring to the kinbody in other env with `Save_GrabbedBodies` option is allowable if it's called from `Environment::_Clone`.
   - This is because environmentBodyIndex used in `Restore` is not consistent between two different envs, but `Environment::_Clone` intentinally configures such index to be consistent between these two.
   - However, `Restore` is purely saver's API and in future, the user might  use it in the wrong situation. In such case, the restored information is not correct anymore and dangerous.
 - Resolution
   - To avoid this, deprecate restoring to the kindboy in the other env with `Save_GrabbedBodies` from saver.
   - Instead, add the dedicated private API to `KinBody` and `Environment::_Clone` calls it, since `Environment` is a friend of `KinBody`. That way, we can reduce the risk of calling restoring function in a wrong situation.
   - Restore to the kinbody in other env for other options is still available.
   - Restore to the kinbody in the same env with Save_GrabbedBodies is still available.

# Improvement 1
- To avoid the bugs like `Bugs 3` in future, share the same utility function between `KinBodyStateSaver` and `KinBodyStateSaverRef`.
- Introduce `KinBody::_RestoreKinBodyGrabbedBodiesFromSavedData`.
  - https://github.com/rdiankov/openrave/pull/1425/commits/2c2184d5172e6d5e1c08fc6564317e11c6402af7
  - https://github.com/rdiankov/openrave/pull/1425/commits/1c81ce0e4269717d3dedd59283b087429fa3b3d4
- Instead of this, I tried to remove `KinBodyStateSaver` (pointer version). But, it was not trivial and introduced the utility function at this moment.

# Improvement2
- Add missing enum https://github.com/rdiankov/openrave/pull/1425/commits/562d2341d64030a61405fb2545b9055c7fe98578
- Reduce duplicated code for Getgrabbedinfo https://github.com/rdiankov/openrave/pull/1425/commits/33eaf8824ea32d53d74343ed757ab4cf044d9b66
- Remove unused `CheckGrabbedInfo` overloads. In our code base, there is only one place to use it. so, it looks other overloads are not necesssary : https://github.com/rdiankov/openrave/pull/1425/commits/facd9474958d7209d2d70e106edb52dc2ac527bb

# Other bugs not fixed in this PR yet.
- Some of the APIs like `_RestoreKinBody`, `ResetGrabbed`, ...etc are implemented : 1) release all grabbed bodies. 2) restore necesary grabbed bodies one by one.
- However, if there is an exception between 1 and 2, all necessary grabbed bodies are gone. Even if the caller side uses try-catch, it's not recoverable.
- For example, some of the robot grabs cameras from the static setting in the scene file. If such robot grab the target workpiece, and this bug happens, all of the cameras are gone, and looks dangerous.
- We'd better not to remove all of the grabbed bodies and need to improve the fallback logic. To do so, we need to determine what's the expected behavior for such case.


snozawa:
hi @yoshikikanemoto 

This PR tries to restore the states of  grabbed bodies.

But, when I see the past PR, I'm confused now.
https://github.com/rdiankov/openrave/pull/1324/

I thought the responsibility of  `Save_GrabbedBodies` is to revert the changes on `Grabbed` instance.
However, `Grabbed::ComputeListNonCollidingLinks` internally calls `Save_GrabbedBodies`, which means,  the function wants to compute `_listNonCollidingLinksWhenGrabbed`, but also wants to revert it.

Should we revert the internal states of `Grabbed` instances when `Save_GrabbedBodies` are specified?
Or, should we keep them?

snozawa:
> hi @yoshikikanemoto
> 
> This PR tries to restore the states of grabbed bodies.
> 
> But, when I see the past PR, I'm confused now. #1324
> 
> I thought the responsibility of `Save_GrabbedBodies` is to revert the changes on `Grabbed` instance. However, `Grabbed::ComputeListNonCollidingLinks` internally calls `Save_GrabbedBodies`, which means, the function wants to compute `_listNonCollidingLinksWhenGrabbed`, but also wants to revert it.
> 
> Should we revert the internal states of `Grabbed` instances when `Save_GrabbedBodies` are specified? Or, should we keep them?

This PR changes this by avoiding the `Save_GrabbedBodies` in the `ComputeListNonCollidingLinks`.

snozawa:
discussed with @yoshikikanemoto 

Since the PR became big, I'll try splitting this into smaller PRs.

PR #1426, opened 2024-09-16
Fix missing format variable

rschlaikjer:
Missing format variable causes exception on env destroy if there are enough bodies in it


rdiankov:
Thanks

PR #1428, opened 2024-09-18
Fix ResetGrab and RelesaeAllWithLink.

snozawa:
# Summary
- Fix several bugs in Reset and Release grabbed.
- These commits are splitted from the original all-in-one branch: https://github.com/rdiankov/openrave/pull/1425. These commits are tested via all-in-one branch as well.

# Bug fix 1
- In `ReleaseAllWithLink` API, other grabbed bodies are not post processed, while `Release` API postprocesses it by calling `_RemoveGrabbedBodies`. Thus, unused _listNonCollidingLinkWhenGrabbed remains after `ReleaseAllWithLink`.
-  Fixed via : https://github.com/rdiankov/openrave/commit/1fbd14a7115dfca682f99ca953ec1cee8686cb14

# Bug fix 2
- Fix the bugs of ResetGrabbed, introduced in https://github.com/rdiankov/openrave/pull/1421.
- Issue
   - Members of Grabbed is overwritten in ResetGrabbed, such as `tRelative`. However, `pGrabbedSaver` and `pGrabberSaver` are initialized in the constructor and these are not consistent with the newly set `tRelative`.
   - The pointer of `Grabbed` is held by `KinBodyStateSaver`. If `ResetGrabbed` overwrites the contents of `Grabbed` instance, the saved information in the saver gets changed.
- Resolution
   - Revert some of the change in https://github.com/rdiankov/openrave/pull/1421, Thanks for @rschlaikjer to make this commit!
   - In detail, `ResetGrabbed` creates the new `Grabbed` instances for consistency. As a background
        - To resolve the issue 2, `KinBodyStateSaver` might be able to have copy of `Grabbed`. But, it might be more expensive than having the new `Grabbed` in `ResetGrabbed`. Based on such trade-off, this commit creates the new `Grabbed` in `Reset
Grabbed`.
        - To resolve the issue 1, `ResetGrabbed` might be able to update the `pGrabbedSaver` and `pGrabberSaver`. But, it still allocates memories and creating the new Grabbed would be simpler.
  - fixed via https://github.com/rdiankov/openrave/commit/b9a02dfb6d5472ef1d99416df1226f6e9592a329


rschlaikjer (approved):

src/libopenrave/kinbodygrab.cpp
```diff
@@ -825,11 +826,10 @@ std::string KinBody::GrabbedInfo::GetGrabbedInfoHash() const
 void KinBody::ResetGrabbed(const std::vector<KinBody::GrabbedInfoConstPtr>& vGrabbedInfos)
 {
     // First pass: remove all existing grabs that are no longer valid per the new list of grab infos
+    // Keep track of which bodies are still attached, so that we can skip re-attaching them later
+    std::unordered_set<int> existingAttachedGrabBodyIndices;
```
kanbouchou, line 830: What about `vector<uint64_t>` as underlying data structure? Each bit represents attached state. N'th bit represents attached state for environment body index `N`. If somebody has env body index larger than 64, it can go to next element in the vector. This is how `_vLinkEnableStatesMask` is represented, and there are utility functions (`*StateBitMasks`) around it.

I think `vector` would be more performant than `unordered_set` for this case where data (environment body index of bodies) is dense from 0 to highest body index in the environment.

snozawa: Thanks for this comment!

@rschlaikjer 

do you have any idea about it?
I think recent these code started to use the `unordered_set`.
I'm not fully catching up the past discussion about how the `undered_set` is introduced : https://github.com/rdiankov/openrave/pull/1421

rdiankov: @kanbouchou in AGV environments, we might have 10k objects, so it is not practical to allocate that much memory, unordered_set would be faster.

rdiankov:
thanks~

src/libopenrave/kinbodygrab.cpp
kanbouchou: > @kanbouchou in AGV environments, we might have 10k objects, so it is not practical to allocate that much memory, unordered_set would be faster.

I see. When most of the bits would be 0 (number of attached bodies is small compared to number of total bodies), bit mask like representation is not that great.

PR #1429, opened 2024-09-18
Fix misc bugs in grabbed bodies saver

snozawa:
# Summary
- Several bug fixes when we call `KinBodyStateSaver` classes with `Save_GrabbedBodies`.
- This PR is separated from https://github.com/rdiankov/openrave/pull/1425. This PR contents are tested with all-in-one branch : https://github.com/rdiankov/openrave/pull/1425

# Bug fix 1
- Fix typo. Otherwise, restoring toward cloned env does not correctly work. 
- Fixed via : https://github.com/rdiankov/openrave/pull/1429/commits/de1911446d1f4ff826a7e9f432642fce2e574321

# Bug fix 2
- Fix typo, which is newly introduced in https://github.com/rdiankov/openrave/pull/1425/commits/cbd375ccf505af64950798e54a4c3087f22e56e0

# Bug fix 3
- In the previous code, `_listNonCollidingLinksWhenGrabbed` does not restore the other grabbed bodies at all.
- Fixed via https://github.com/rdiankov/openrave/pull/1429/commits/cc2355226702a10608fe2111c898b1c885859f4a and later commits

 # Bug fix 4 and improvement.
 - Restoring code is duplicated between `KinBodyStateSaver` and `KinBodyStateSaverRef`. Some of the past improvement exists in one class, but does not exist in another.
 - Apply the fixes to another class
  - https://github.com/rdiankov/openrave/commit/4c3ae1927af5fcc99c2c92adef277ab3e4182a35 
  - https://github.com/rdiankov/openrave/commit/bc9f2e1006590a8221cceed3845cc5a6658219c9 
  - https://github.com/rdiankov/openrave/commit/d05e69a284762e1d7af4fc9b9775d69d00ec44ae
  - To prevent from future bugs like that, use the common utility code for restoring between these classes.
     - like https://github.com/rdiankov/openrave/pull/1429/commits/318fc86764101b51f8ba6562a827878f0861305d
     - Instead of this, I tried to remove KinBodyStateSaver (pointer version). But, it was not trivial and introduced the utility function at this moment.


src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -18,6 +18,121 @@
...
+                                                        const int linkindex,
+                                                        const KinBody::LinkConstPtr& pLinkRef,
+                                                        const std::vector<KinBody::LinkPtr>& vLinks,
+                                                        const EnvironmentBasePtr& pEnv)
+{
+    if( linkindex < 0 || linkindex >= (int)vLinks.size() ) {
+        RAVELOG_WARN_FORMAT("env=%s, could not restore link '%s' since its index %d is out of range (body num links is %d)", pEnv->GetNameId()%pLinkRef->GetName()%pLinkRef->GetIndex()%vLinks.size());
+    }
+    else {
+        grabbed._listNonCollidingLinksWhenGrabbed.push_back(vLinks.at(linkindex));
```
rschlaikjer: this access is already ranged checked, may as well use `operator[]` instead of `at` here so we don't waste time checking again

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -18,6 +18,121 @@
...
+    }
+}
+
+void KinBody::_RestoreGrabbedBodiesFromSavedData(const KinBody& savedBody,
+                                                 const int options,
+                                                 const std::unordered_map<int, GrabbedPtr>& savedGrabbedBodiesByEnvironmentIndex)
+{
+    {
```
rschlaikjer: Why is the entire function doubly-nested?

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -18,6 +18,121 @@
...
+        // have to release all grabbed first
+        ReleaseAllGrabbed();
+        OPENRAVE_ASSERT_OP(_grabbedBodiesByEnvironmentIndex.size(),==,0);
+        for (const std::unordered_map<int, GrabbedPtr>::value_type& grabPair : savedGrabbedBodiesByEnvironmentIndex) {
+            const GrabbedPtr& pGrabbed = grabPair.second;
+            KinBodyPtr pGrabbedBody = pGrabbed->_pGrabbedBody.lock();
+            if( !!pGrabbedBody ) {
```
rschlaikjer: If we invert this check (`if (!pGrabbedBody) { WARN(...); continue; }`) we can unnest a significant portion of this function

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -18,6 +18,121 @@
...
+
+void KinBody::_RestoreGrabbedBodiesFromSavedData(const KinBody& savedBody,
+                                                 const int options,
+                                                 const std::unordered_map<int, GrabbedPtr>& savedGrabbedBodiesByEnvironmentIndex)
+{
+    {
+        // have to release all grabbed first
+        ReleaseAllGrabbed();
+        OPENRAVE_ASSERT_OP(_grabbedBodiesByEnvironmentIndex.size(),==,0);
+        for (const std::unordered_map<int, GrabbedPtr>::value_type& grabPair : savedGrabbedBodiesByEnvironmentIndex) {
```
rschlaikjer: You use use `MapGrabbedByEnvironmentIndex::value_type` to ensure that if the type changes we don't have to update it everywhere

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -18,6 +18,121 @@
...
+                    KinBodyPtr pNewGrabbedBody = GetEnv()->GetBodyFromEnvironmentBodyIndex(pGrabbedBody->GetEnvironmentBodyIndex());
+                    if( !!pNewGrabbedBody ) {
+                        if( pGrabbedBody->GetKinematicsGeometryHash() != pNewGrabbedBody->GetKinematicsGeometryHash() ) {
+                            RAVELOG_WARN_FORMAT("env=%s, new grabbed body '%s' kinematics-geometry hash is different from original grabbed body '%s' from env=%s", GetEnv()->GetNameId()%pNewGrabbedBody->GetName()%pGrabbedBody->GetName()%savedBody.GetEnv()->GetNameId());
+                        }
+                        else {
+                            // body is supposed to already be set to some "proper" configuration as the newly
+                            // initialized Grabbed objects will save the current state of pbody for later computation of
+                            // _listNonCollidingLinksWhenGrabbed (in case it is not yet computed).
+                            KinBody::LinkPtr pNewGrabbingLink = GetLinks().at(KinBody::LinkPtr(pGrabbingLink)->GetIndex());
```
rschlaikjer: What's the cast for?

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -18,6 +18,121 @@
...
+                    // grabbed bodies could have been removed from env and self collision checker.
+                    CollisionCheckerBasePtr collisionchecker = GetSelfCollisionChecker();
+                    if (!!collisionchecker) {
+                        collisionchecker->InitKinBody(pGrabbedBody);
+                    }
+                }
+                else {
+                    // The body that the state was saved from is from a different environment from savedBody. This case can
+                    // happen when cloning an environment (see Environment::_Clone). Since cloning is supposed to
+                    // preserve environmentBodyIndex of bodies, it is ok do the following.
```
rschlaikjer: Is it possible to clone an env that has state savers that reference body N, remove body N from the original environment, add a new body that then gets reassigned that index, and _then_ have the state saver trigger? What happens in this case? 
Actually, why do the cloned bodies / state savers point to the original environment? Wouldn't it be better if they all pointed to bodies within their own env?

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -87,86 +202,7 @@ void KinBody::KinBodyStateSaver::_RestoreKinBody(boost::shared_ptr<KinBody> pbod
...
-                    } // end if !!pNewGrabbedBody
-                }
-            }
-        }
-
-        // if not calling SetLinkTransformations, then manually call _UpdateGrabbedBodies
-        if( !(_options & Save_LinkTransformation ) ) {
-            pbody->_UpdateGrabbedBodies();
-        }
+        pbody->_RestoreGrabbedBodiesFromSavedData(*_pbody, _options, _grabbedBodiesByEnvironmentIndex);
```
rschlaikjer, line 204: Love to see duplicate code elimination

rschlaikjer:
Do you have new test cases to go along with these changes?

src/libopenrave/kinbodystatesaver.cpp
snozawa: right!

fixed it!

src/libopenrave/kinbodystatesaver.cpp
snozawa: The motivation is same as https://github.com/rdiankov/openrave/pull/1429#discussion_r1767724097.

I originally tried to cleanup the codes, by minimizing the `diff`.
The original code is like this:
```
    if( _options & Save_GrabbedBodies ) {
        xxx
    }
```
And, i just removed this `if` statement, and avoided from other cleaning up to save the diff lines.

However, I now ended up with separating the duplicated code into one function, and thus, we already have huge `diff`.

So, I don't need to stick to the original code, so i'll try to resolve this, and https://github.com/rdiankov/openrave/pull/1429#discussion_r1767724097.

Thanks!

snozawa: fixed via https://github.com/rdiankov/openrave/pull/1429/commits/4970937f95fa69faa5839465e0b3b21be80db0bf

src/libopenrave/kinbodystatesaver.cpp
snozawa: same for the above, fixed via https://github.com/rdiankov/openrave/pull/1429/commits/4970937f95fa69faa5839465e0b3b21be80db0bf.

Thanks!

snozawa:
> Do you have new test cases to go along with these changes?

yes!
I replied in another channel.

src/libopenrave/kinbodystatesaver.cpp
snozawa: Thanks for your comments!

Actually, your concerns are very good points.

> Is it possible to clone an env that has state savers that reference body N, remove body N from the original environment, add a new body that then gets reassigned that index, and then have the state saver trigger? What happens in this case?

It's kind of undefined behavior.

> Actually, why do the cloned bodies / state savers point to the original environment? Wouldn't it be better if they all pointed to bodies within their own env?

Yea, totally agree.


This code is already in `production` branch mostly.
It's purely used from `Environment::_Clone`.
https://github.com/rdiankov/openrave/blob/production/src/libopenrave-core/environment-core.h#L3883-L3884

As for your former question, the `_Clone` function seems carefully written to preserve the `envBodyIndex` of the original env.
So, this use case is barely ok.

But, if the user accidentally call the saver's functionality, it's not safe in general.
And, unfortunately, there seems no way to detect such wrong use case inside of saver.

Thus, in another PR, I deprecate the saver to restore the grabbed bodies from one env to another env. I still keep the use case from `_Clone`:
https://github.com/rdiankov/openrave/pull/1431

src/libopenrave/kinbodystatesaver.cpp
snozawa: I guess this is same question as https://github.com/rdiankov/openrave/pull/1429#discussion_r1767769482?

rschlaikjer: No I mean why is there an explicit `KinBody::LinkPtr()` around pGrabbingLink? Isn't it already a linkptr? Won't this just create another shared_ptr just to delete it again?

rdiankov:
@snozawa @rschlaikjer let me know when this is ready. thanks

src/libopenrave/kinbodystatesaver.cpp
snozawa: Oh, right!

The code was originally there, but indeed, it's unnecessary.

I also find one more place that creates unnecessary shared_ptr.

Fixed via https://github.com/rdiankov/openrave/pull/1429/commits/d53538ed7bf238aab574e9c69fb5dad856c4cc99, thanks!

src/libopenrave/kinbodystatesaver.cpp
snozawa: @rschlaikjer 

Oh, the original code is written by you https://github.com/rdiankov/openrave/commit/9176271a500f12fbb3f1392fd839a329c40f9189.

So, do you want to replace this by `MapGrabbedByEnvironmentIndex::value_type`, and to  slightly re-order the declaration of `MapGrabbedByEnvironmentIndex` so that we can replace the same thing in `KinBodyStateSaver` member?

Even though, can we do that change in another PR, since it's touching the code line to update the saved data contents?
https://github.com/rdiankov/openrave/pull/1430

Thanks!

rschlaikjer: Oh I see, this is outside the range of that using-decl. It was a small comment, feel free to leave it or extract that decl in a different PR.

snozawa: thanks!
I'll try revisiting in the next PR!

src/libopenrave/kinbodystatesaver.cpp
```diff
@@ -18,6 +18,120 @@
 
 namespace OpenRAVE {
 
+/// \brief push link to _listNonCollidingLinksWhenGrabbed of grabbed.
+static void _PushLinkToListNonCollidingLinksWhenGrabbed(Grabbed& grabbed,
```
yoshikikanemoto, line 22: @snozawa comment for readability: I had to check usage of `pLinkRef`. but `pLinkRef->GetIndex()` is guaranteed to be the same as `linkindex` with the current usage. It would be easier to read if you pass only the name of the link to this function.

snozawa: thanks for your comment!

fixed it!

https://github.com/rdiankov/openrave/pull/1429/commits/4b85188527c5b9cbabeaec07349476939b18c6eb

yoshikikanemoto (approved):
looks good to me now

snozawa:
> @snozawa @rschlaikjer let me know when this is ready. thanks

@rdiankov 
Thanks!
looks ready now.

rdiankov:
thanks!

PR #1430, opened 2024-09-18
Save more grabbed data to prevent from losing of the Grabbed's internal states

snozawa:
# Summary
- Fix the bug where the grabbed bodies states are not fully restored in `KinBodyStateSaver`.
- This PR is separated from https://github.com/rdiankov/openrave/pull/1425. This PR contents are tested with all-in-one branch : https://github.com/rdiankov/openrave/pull/1425
- Also, this PR is on top of https://github.com/rdiankov/openrave/pull/1429

# Bug fix 1
- Fix bugs that `KinBodyStateSaver` does not restore some of the Grabbed's members. This might cause self colliding after saver.
- Background
  - When APIs such as `KinBody::Grab`, `KinBody::Release`, ...etc are called, new `Grabbed` instance is created in `KinBody` and the original ptr for the original `Grabbed` instance is swapped.
  - Thus, the original information of `Grabbed` instance is unchanged. Holding the ptr of it as `pGrabbed` is enough to store the saved data for `KinBodyStateSaver`.
- Issue
  - However, some of the grab-related API mutates the internal states of `Grabbed` instance, such as `_listNonCollidingIsValid`, `_listNonCollidingLinksWhenGrabbed`, and `_setGrabberLinkIndicesToIgnore`.
  - This might cause the unexpected deleting of grabbed bodies in the `_listNonCollidingLinksWhenGrabbed`,  e.g. this might cause unexpected self colliding.
- Resolution
  - Introduce the `SavedGrabbedData`. This holds the `GrabbedPtr` same as before. In addition, this holds the copied data for _listNonCollidingIsValid, _listNonCollidingLinksWhenGrabbed, and _setGrabberLinkIndicesToIgnore.
- Fixed via  https://github.com/rdiankov/openrave/commit/7374bdf633faf0b185b9808acdbdc05f20b48bed and https://github.com/rdiankov/openrave/commit/d3d189d63dd66c0d4b08ff7c6820345904201689

# Bug fix 2
- Related fix to Bug fix 1.
- Issue
  - Actually, the meaning of `Save_GrabbedBodies` were not clear in the previous code.
    - KinBodyStateSaver holds Grabbed pointers, and Grabbed uses KinBodyStateSaver with Save_GrabbedBodies. Therefore, the relationship is cyclic.
    - It looks `Save_GrabbedBodies` of KinBodyStateSaver wants to restore any data saved before the saver constructor.
    - However, the `Save_GrabbedBodies` used in `Grabbed::ComputeListNonCollidingLinks` wants to restore some of the data, but does not want to restore others like `_listNonCollidingLinksWhenGrabbed`.
  - Resolving of this issue is also required for Bug 1.
- Resolution
  - Avoid cyclic relationships. Do not use `Save_GrabbedBodies` in `Grabbed::ComputeListNonCollidingLinks`.
  - Instead, call link-link collision checking APIs. This method is actually used in `KinBody::CheckSelfCollision` to achieve the collision checking with the attached bodies.
- Fixed via https://github.com/rdiankov/openrave/commit/adf67a0ab3cc296fb564782a5807875a611e7ee5 and https://github.com/rdiankov/openrave/commit/e96b03aafcbca62eb308293bc47fccefc351ce72


rschlaikjer (approved):
Makes sense I think, do you have test cases to verify the new behaviour?

Also since this is an extension of #1429, maybe setting the merge base to that branch instead of production would be useful to highlight the actual changes

snozawa:
Really appreciate for your comment and checking!

Yes, i do have the test cases and replied in another channel.

> Also since this is an extension of #1429, maybe setting the merge base to that branch instead of production would be useful to highlight the actual changes

If we change the base from `production` to [this](https://github.com/rdiankov/openrave/pull/1429), the `diff` looks fine.
But, once we'll merge this into `production` branch, we need to revert the base to `production`.

Is this correct?
If so, i'm not sure which is better.

snozawa:
I set this as WIP since i might find a regression in this PR.

I replaced link-kinbody collision checking by link-link collision checking.
In my current implementation, I missed to find the links in the attached bodies of the target.

I'm thinking about how to resolve it.

Btw, do we have a use case of `CheckSelfCollision` for nested grabbed bodies?
e.g. robot grabs `A`, and `A` grabs `B`, and then check self collision.

@kanbouchou @yoshikikanemoto

snozawa:
> I set this as WIP since i might find a regression in this PR.
> 
> I replaced link-kinbody collision checking by link-link collision checking. In my current implementation, I missed to find the links in the attached bodies of the target.
> 
> I'm thinking about how to resolve it.
> 
> Btw, do we have a use case of `CheckSelfCollision` for nested grabbed bodies? e.g. robot grabs `A`, and `A` grabs `B`, and then check self collision.
> 
> @kanbouchou @yoshikikanemoto

~~I think `CheckSelfCollision` of nested grabbed bodies are not supported.~~

~~In the following code,~~

~~https://github.com/rdiankov/openrave/blob/production/src/libopenrave/kinbodycollision.cpp#L117~~
~~https://github.com/rdiankov/openrave/blob/production/src/libopenrave/kinbodycollision.cpp#L167~~

~~we only refer the grabbed body's links.~~
~~no checking for attached bodies' links.~~

~~Some time in future, it might be worth to support `CheckSelfCollision` for nested grabbed bodies.~~
~~But, it's not trivial change.~~

~~So far, I think it might be safer to add assert in  `CheckSelfCollision`  or in `ComputeListNonCollidingLinks`, so that we can allow nested grabbed bodies modeling, but once we call `CheckSelfCollision`, ask user to notify that it's not supported yet.~~

I updated my understanding. In the current production branch:

- Modeling of nested grabbed bodies are totally supported. Let's say, `A` grabs `B`, and `B` grabs `C`.
- We slightly support the `CheckSelfCollisoin` in such case.
   - One-layer collision checking is conducted. Checking between `A` and `B` is done. Checking between `B` and `C` is done.
   - However, we don't check self collision between `A` and `C`.
   
 In that sense, this PR does not change this behavior.
 Slightly changes the unexpected checking of collision of nested bodies in `ComputeListNonCollidingLinks`.
 
 In future, we might be able to consider how to support the collision checking between `A` and `C`.

include/openrave/kinbody.h
```diff
@@ -2471,7 +2484,7 @@ class OPENRAVE_API KinBody : public InterfaceBase
         std::vector<std::pair<Vector,Vector> > _vLinkVelocities;
         std::vector<dReal> _vdoflastsetvalues;
         std::vector<dReal> _vMaxVelocities, _vMaxAccelerations, _vMaxJerks, _vDOFWeights, _vDOFLimits[2], _vDOFResolutions;
-        std::unordered_map<int, GrabbedPtr> _grabbedBodiesByEnvironmentIndex;
+        std::unordered_map<int, SavedGrabbedData> _grabbedDataByEnvironmentIndex;
```
rdiankov, line 2487: @snozawa I'm guessing you are introducing `SavedGrabbedData` because `GrabbedPtr`  can change and then we cannot restore? Why not have a clone of GrabbedPtr saved here instead? Why go through all the trouble of creating a new structure `SavedGrabbedData` which still maintains the old Grabbed pointer in `pGrabbed`?

snozawa: thanks for your comment!
@rdiankov 

I thought using `GrabbedPtr` is the original code is intended for performance.
so, I follow the similar pattern.

Here is an example.
In the following example, there is no difference about the copied data in `SavedGrabbedData`.
So, I skip it and only focus on `GrabbedPtr`.

The first example is for this PR and thus mostly in the original `production` branch, e.g. storing the `GrabbedPtr` in `KinBodyStateSaver`.
Screenshot from 2024-09-27 09-36-52

In this scenario, 
1. The robot is originally grabbing `grabbed1`.
2. `KinBodyStateSaver` saves it. Just a pointer operation.
3. The robot releases `grabbed1` and then grabs `grabbed2`. `grabbed2` is `new`-ed.
4. `KinBodyStateSaver` restores the `grabbed1`, and it's pointer exchange.

The second example is to clone the original `Grabbed`.
Screenshot from 2024-09-27 09-43-14

In this scenario,
1. The robot is originally grabbed `grabbed1`.
2. `KinBodyStateSaver` saves it as cloned one.
3. The robot releases `grabbed1` and then grabs `grabbed2`. `grabbed2` is `new`-ed.
4. `KinBodyStateSaver` restores the `grabbed1`, and it's pointer exchange.

The difference between these two is, the second one requires additional clone, e.g. allocating and copying.
The thing here is `Grabbed` class holds internal `KinBodyStateSaver`s and cloning of it is not supported, and even if we'll support it, I guess cloning of the savers are not negligibly small. (btw, this seems necessary for lazy-computation : https://github.com/rdiankov/openrave/pull/1038)

The first example is mostly in the `production` branch, and I think it would be better for the performance, especially for the recent trend about high-count situations : https://github.com/rdiankov/openrave/pull/1421

This PR's proposal is we still need some of the member to copy, but we'd like to minimize such copying.



One of the possible improvement on this is to get the compiler's help.

By using the following way, we can share the same structure between saver and `Grabbed` classes.
And, we can make immutable `Grabbed` member as `const` and clarify the mutable member with non `const`.
(of course, some of the formers are not perfectly constant, since the pointer is `const` but the pointed object needs to call non-const API).


```
struct MutableGrabbedData
{
    std::list<KinBody::LinkPtr> listNonCollidingGrabbedGrabberLinkPairsWhenGrabbed;
    std::set<int> setGrabberLinkIndicesToIgnore;
    bool listNonCollidingIsValid = false;
};

struct SavedGrabbedData
{
    GrabbedPtr pGrabbed;
    MutableGrabbedData mutableGrabbedData;
}

class OPENRAVE_API KinBodyStateSaver
{
    ...
    std::unordered_map<int, SavedGrabbedData> _grabbedDataByEnvironmentIndex;
    ...
}

class OPENRAVE_API Grabbed : public UserData, public boost::enable_shared_from_this<Grabbed>
{
public:
    Grabbed(KinBodyPtr pGrabbedBody, KinBody::LinkPtr pGrabbingLink, const Transform& tRelative, const rapidjson::Document& rGrabbedUserData);
    ...

    // Member Variables
    const KinBodyWeakPtr _pGrabbedBody;
    const KinBody::LinkPtr _pGrabbingLink;
    const Transform _tRelative;
    const rapidjson::Document _rGrabbedUserData;

    MutableGrabbedData _mutableGrabbedData;
private:
    const std::vector<KinBody::LinkPtr> _vAttachedToGrabbingLink;
    const KinBody::KinBodyStateSaverPtr _pGrabberSaver;
    const KinBody::KinBodyStateSaverPtr _pGrabbedSaver;
}; // end class Grabbed
```

rdiankov: i see, thanks for the explanation.

src/libopenrave/kinbodygrab.cpp
```diff
@@ -122,64 +151,78 @@ void Grabbed::ComputeListNonCollidingLinks()
...
+            _RemoveLinkFromListNonCollidingLinksWhenGrabbed<KinBodyPtr>(pOtherGrabbed->_listNonCollidingLinksWhenGrabbed, pGrabbedBody);
+
+            // extract valid pointers
+            KinBodyPtr pOtherGrabbedBody = pOtherGrabbed->_pGrabbedBody.lock();
+            if( !pOtherGrabbedBody ) {
+                RAVELOG_WARN_FORMAT("env=%s, other grabbed body on %s has already been released. So ignoring it.", penv->GetNameId()%pGrabber->GetName());
+                continue;
+            }
+            if( pOtherGrabbedBody->GetEnvironmentBodyIndex() == 0 ) {
+                RAVELOG_WARN_FORMAT("env=%s, other grabbed body on %s has invalid environment body index. Perhaps already removed from the environmnet. So ignoring it.", penv->GetNameId()%pGrabber->GetName());
```
yoshikikanemoto:
```suggestion
                RAVELOG_WARN_FORMAT("env=%s, other grabbed body on %s has invalid environment body index. Perhaps already removed from the environment. So ignoring it.", penv->GetNameId()%pGrabber->GetName());
```

yoshikikanemoto (approved):

rdiankov:
thanks~

PR #1433, opened 2024-09-19
Improve environment loading for bodies with many links

rschlaikjer:
- UpdateOrCreateInfoWithNameCheck performs a sequential scan for each new element; create alternative form that operates on the entire range at once and uses lookup tables into the list
- _InitAndAddLink has the same issue; same remedy
- ComputeInternaInformation has a lot of n^2 operations on the total links, but they can't be removed without some new flag to allow not computing all-link-shortest pairs (data that doesn't seem to be used for our high-link-count bodies anyway). 
  -  We can still slightly optimize some internal loops though to at least do _fewer_ iterations
  - Calculating the shortest path can be skipped for links that are already adjacent

Draft PR while I check for regressions


rschlaikjer:
See associated testopenrave PR 121

src/libopenrave/libopenrave.h
```diff
@@ -481,6 +481,145 @@ void UpdateOrCreateInfo(const rapidjson::Value& value, std::vector<boost::shared
...
+        // Update our jump tables and push this info onto the list
+        const size_t newInfoIndex = vInfosOut.size();
+        infoIdxById.emplace(id, newInfoIndex);
+        infoIdxByName.emplace(pNewInfo->_name, newInfoIndex);
+        vInfosOut.emplace_back(std::move(pNewInfo));
+    }
+
+    // If we deleted any elements, scan through and ensure there are no holes in the output before returning.
+    // Scan backwards through the list, swapping all of the invalid pointers to the end.
+    // Once we're done, truncate the list to drop all the invalid entries.
```
rschlaikjer: I delete things this way to avoid `memmove`, but the tradeoff is that we don't guarantee the total ordering of links with this implementation - is this something that is implicitly relied on somewhere? I think the old impl would have the same issue if you had a delete + add, but wanted to bring it up to be sure.

include/openrave/kinbody.h
```diff
@@ -3675,6 +3675,16 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
 
+    /// \brief initializes the internal link vector with the provided link infos.
+    ///
+    /// Internal method called as part of Init() or InitFromLinkInfos()
+    /// LinkInfoT may be a LinkInfo struct or a pointer to a LinkInfo
+    template <typename LinkInfoT>
+    void _InitWithInitialLinks(const std::vector<LinkInfoT>& linkInfos);
+
+    /// Initialize a valid link pointer from the specified link info
+    void _InitLinkFromInfo(KinBody::LinkPtr& linkPtr, const KinBody::LinkInfo& linkInfo);
```
kanbouchou: Does this have to be a class method? Could this be a free function?

src/libopenrave/kinbody.cpp
```diff
@@ -4964,14 +5003,10 @@ void KinBody::_ComputeInternalInformation()
...
-            if( link0.GetGeometries().size() == 0 ) {
+            if (link0.GetGeometries().size() == 0) {
                 int ind0 = link0.GetIndex();
-                for (const LinkPtr& plink1 : _veclinks) {
-                    const Link& link1 = *plink1;
-                    int ind1 = link1.GetIndex();
-                    if (ind0 != ind1) {
-                        _vAdjacentLinks.at(_GetIndex1d(ind0, ind1)) = 1;
-                    }
+                for (int ind1 = ind0 + 1; ind1 < _veclinks.size(); ind1++) {
```
kanbouchou: I think you still need to check links whose index is smaller than `ind0`. Say, if `_veclinks[1]` has no geometry, `_vAdjacentLinks.at(_GetIndex1d(1,0))` should be set to 1, but this code doesn't do so.

src/libopenrave/libopenrave.h
```diff
@@ -481,6 +481,145 @@ void UpdateOrCreateInfo(const rapidjson::Value& value, std::vector<boost::shared
     vInfos.push_back(pNewInfo);
 }
 
+template <typename T>
```
kanbouchou, line 489: Please provide docstring for this function.

src/libopenrave/libopenrave.h
```diff
@@ -481,6 +481,145 @@ void UpdateOrCreateInfo(const rapidjson::Value& value, std::vector<boost::shared
...
+void UpdateOrCreateInfosWithNameCheck(rapidjson::Value::ConstValueIterator sourceItBegin, rapidjson::Value::ConstValueIterator sourceItEnd, std::vector<boost::shared_ptr<T>>& vInfosOut, const char* pNameInJson, dReal fUnitScale, int options)
+{
+    // If the begin/end iterators are empty, don't bother even building the lookup table
+    if (sourceItBegin == sourceItEnd) {
+        return;
+    }
+
+    // In order to memoize name/id match lookup, create jump tables of name/id to vInfosOut index
+    // Note that we allocate full strings rather than string views because otherwise deserializing into the existing info struct may mutate the view out from under the map
+    std::unordered_map<std::string, size_t> infoIdxByName;
```
kanbouchou, line 499: Would it be faster if we `reserve` by `vInfosOut.size()`?

src/libopenrave/libopenrave.h
kanbouchou: > I think the old impl would have the same issue if you had a delete + add

I think old information (`UpdateOrCreateInfoWithNameCheck`) doesn't change the order of elements in `vInfos`, does it?

> but the tradeoff is that we don't guarantee the total ordering of links with this implementation - is this something that is implicitly relied on somewhere? 

I cannot think of this assumption being made somewhere, but I feel that it is safer to keep the order. What do you think @rdiankov ?

rdiankov:
@kanbouchou @snozawa @rschlaikjer let me know when this is ready. thanks

include/openrave/kinbody.h
rschlaikjer: It accesses protected members of the Link, so it can't be a free function

src/libopenrave/kinbody.cpp
rschlaikjer: Hmm yeah I was thinking this was commutative, but it isn't. Will revert.

src/libopenrave/libopenrave.h
rschlaikjer: > I think old information (UpdateOrCreateInfoWithNameCheck) doesn't change the order of elements in vInfos, does it?

It will if you do a delete then an add, as mentioned - 
```
vInfosOut = {a, b, c, d};
UpdateOrCreateInfosWithNameCheck(vInfosOut, {b, __deleted__}); // vInfosOut = {a, c, d}
UpdateOrCreateInfosWithNameCheck(vInfosOut, {b}); // vInfosOut = {a, c, d, b}
```
does this show up naturally? Not sure, but it's possible.

If we do have to revert it, it's slower for the delete case but it's still on par with the old method, and delete is (probably?) a colder path.

include/openrave/kinbody.h
kanbouchou: Hm, what about `static` method then? The method is not touching `this`?

rschlaikjer: sure

kanbouchou:
> @kanbouchou @snozawa @rschlaikjer let me know when this is ready. thanks

@rdiankov I think it looks good now. Please check following thread just in case.

https://github.com/rdiankov/openrave/pull/1433#discussion_r1770866802

src/libopenrave/kinbody.cpp
```diff
@@ -726,16 +722,64 @@ void KinBody::SetLinkGroupGeometries(const std::string& geomname, const std::vec
     _PostprocessChangedParameters(Prop_LinkGeometryGroup); // have to notify collision checkers that the geometry info they are caching could have changed.
 }
 
+void KinBody::_InitLinkFromInfo(KinBody::LinkPtr& linkPtr, const KinBody::LinkInfo& linkInfo)
+{
+    linkPtr->_vGeometries.clear();
+    linkPtr->_collision.vertices.clear();
+    linkPtr->_collision.indices.clear();
```
lazydroid, line 729: we might know in advance how many items should be there, might "reserve()" instead of adding elements one-by-one with reallocations

src/libopenrave/libopenrave.h
```diff
@@ -481,6 +481,152 @@ void UpdateOrCreateInfo(const rapidjson::Value& value, std::vector<boost::shared
...
+
+    // Track whether we deleted any elements and need to garbage collect the output vector before returning
+    bool didEraseElements = false;
+    for (rapidjson::Value::ConstValueIterator it = sourceItBegin; it != sourceItEnd; it++) {
+        // Bind the next JSON value to consider
+        const rapidjson::Value& value = *it;
+
+        // If the ID is non-empty, attempt to look up a matching entry in vInfosOut
+        std::string id = OpenRAVE::orjson::GetStringJsonValueByKey(value, "id");
+        std::string name = OpenRAVE::orjson::GetStringJsonValueByKey(value, pNameInJson);
```
lazydroid: we could probably "const" these two guys?

src/libopenrave/kinbody.cpp
rschlaikjer: added reserve for _vGeometries, but the collision vectors are sized based on the mesh vertices, which aren't known yet / probably aren't 1:1?

src/libopenrave/libopenrave.h
```diff
@@ -481,6 +481,152 @@ void UpdateOrCreateInfo(const rapidjson::Value& value, std::vector<boost::shared
...
+    if (didEraseElements) {
+        size_t validElementCount = vInfosOut.size();
+        for (ssize_t i = vInfosOut.size() - 1; i >= 0; i--) {
+            // If this element is still valid, leave it where it is
+            if (!!vInfosOut[i]) {
+                continue;
+            }
+
+            // If this element is null, swap it with the last valid element and decrement the valid element count
+            std::swap(vInfosOut[i], vInfosOut[--validElementCount]); // Note prefix decrement to turn the count into an index
```
lazydroid: just curious here, why can't we use a simple assignment and have to `swap()` ?

src/libopenrave/kinbody.cpp
lazydroid: agree about collisions, thank you!

src/libopenrave/libopenrave.h
rschlaikjer: Swapping is faster because it avoids a construct/incref (when assigning the last valid element into this position) and a desctruct/decref (when we truncate the array). Shared pointer refcounting is atomic, so it's a bit expensive, and swapping/moving where possible helps avoid this cost.

lazydroid: sounds good, thanks!

src/libopenrave/kinbody.cpp
```diff
@@ -346,25 +346,21 @@ void KinBody::KinBodyInfo::DeserializeJSON(const rapidjson::Value& value, dReal
...
-        }
+        UpdateOrCreateInfosWithNameCheck(value["links"].Begin(), value["links"].End(), _vLinkInfos, "name", fUnitScale, options);
 
-        // if has conflicting names, should error here
-        for(int ilink0 = 0; ilink0 < (int)_vLinkInfos.size(); ++ilink0 ) {
-            for(int ilink1 = ilink0+1; ilink1 < (int)_vLinkInfos.size(); ++ilink1 ) {
-                if( _vLinkInfos[ilink0]->_name == _vLinkInfos[ilink1]->_name ) {
-                    throw OPENRAVE_EXCEPTION_FORMAT("Body '%s' has info with link[%d] and link[%d] having the same linkname '%s', which is not allowed. link[%d].id='%s', link[%d].id='%s'", _name%ilink0%ilink1%_vLinkInfos[ilink0]->_name%ilink0%_vLinkInfos[ilink0]->_id%ilink1%_vLinkInfos[ilink1]->_id, ORE_Assert);
-                }
+        // Assert that we don't have any duplicate link names
```
snozawa, line 351: @rschlaikjer 

might be minor comment, but does `UpdateOrCreateInfosWithNameCheck` guarantee the uniqueness of the link names?

If so, do we need the following additional checking?

rschlaikjer: It's still possible to have duplicate names - since we match on id preferentially to name, the following would create two entries in `UpdateOrCreateInfosWithNameCheck` (IDs are different) but still have a name collision:
```json
[ {"id": "a", "name": "n"}, {"id": "b", "name": "n"}]
```

snozawa: i see~

rdiankov:
@yoshikikanemoto can you do a final check here? thanks

src/libopenrave/libopenrave.h
```diff
@@ -481,6 +481,152 @@ void UpdateOrCreateInfo(const rapidjson::Value& value, std::vector<boost::shared
...
+                }
+            }
+        }
+
+        // If this element is deleted, we need to remove any matching element from the info list
+        bool isDeleted = OpenRAVE::orjson::GetJsonValueByKey<bool>(value, "__deleted__", false);
+        if (existingInfoIndex != -1) {
+            if (isDeleted) {
+                // Erase the referents to this element from our skip lists, but only if they actually point to this index
+                decltype(infoIdxById)::iterator idIt = infoIdxById.find(vInfosOut[existingInfoIndex]->_id);
```
yoshikikanemoto: this `find` and the checks in the next line can be avoided if you reuse the iterator `existingInfoIdxIt` created in the above block.

src/libopenrave/kinbody.cpp
```diff
@@ -726,16 +722,66 @@ void KinBody::SetLinkGroupGeometries(const std::string& geomname, const std::vec
...
+static const KinBody::LinkInfo& _ResolveLinkInfo(const KinBody::LinkInfoConstPtr& linkInfo) { return *linkInfo; }
+static const KinBody::LinkInfo& _ResolveLinkInfo(const KinBody::LinkInfo& linkInfo) { return linkInfo; }
+template <typename LinkInfoT>
+void KinBody::_InitWithInitialLinks(const std::vector<LinkInfoT>& linkInfos)
+{
+    CHECK_NO_INTERNAL_COMPUTATION;
+    BOOST_ASSERT(_veclinks.empty()); // We assume we are the first call to initialize links
+
+    // Keep track of link names as we go to deduplicate
+    std::unordered_set<std::string> existingLinkNames;
```
yoshikikanemoto, line 756: reserve can be used here too

src/libopenrave/libopenrave.h
yoshikikanemoto: Changing order may not be welcome when one has to look at diff of KinBodyInfo JSON.
This will be order preserving and still linear time. Should be better than the old method. can we do this?
```c++
size_t validElementCount = 0;
for (ssize_t iInfo = 0; iInfo < vInfosOut.size(); ++iInfo) {
    if( !!vInfosOut[iInfo] ) {
        if( validElementCount != iInfo ) {
            std::swap(vInfosOut[validElementCount], vInfosOut[iInfo]);
        }
        ++validElementCount;
    }
}
vInfosOut.resize(validElementCount);
```

src/libopenrave/libopenrave.h
rschlaikjer: These are intentionally different - the first check is using the name/id of the input JSON, and the second uses the name/id of the matched deserialized object. The JSON may only specify _one_ of name/id, but the deserialized object may have the other already loaded. Comment added to explain.

src/libopenrave/kinbody.cpp
rschlaikjer: added

yoshikikanemoto (approved):
LGTM

src/libopenrave/libopenrave.h
```diff
@@ -481,6 +481,167 @@ void UpdateOrCreateInfo(const rapidjson::Value& value, std::vector<boost::shared
...
+        // Bind the next JSON value to consider
+        const rapidjson::Value& value = *it;
+
+        // If the ID is non-empty, attempt to look up a matching entry in vInfosOut
+        const std::string id = OpenRAVE::orjson::GetStringJsonValueByKey(value, "id");
+        const std::string name = OpenRAVE::orjson::GetStringJsonValueByKey(value, pNameInJson);
+        OpenRAVE::orjson::GetCStringJsonValueByKey(value, pNameInJson);
+        size_t existingInfoIndex = -1;
+        if (!id.empty()) {
+            decltype(infoIdxById)::iterator existingInfoIdxIt = infoIdxById.find(id);
```
rdiankov: @rschlaikjer can we not use `decltype` like this? Thanks

rdiankov:
merged, thanks.

PR #1434, opened 2024-09-19
Allow skipping generation of all link pair shortest paths

rschlaikjer:
For static bodies that are just used as part of the environment, but that also have a lot of links, generating and then storing the chain of link<->link shortest path is noticeably expensive - for the test case in question, the bod has ~3700 links, resulting in a noticeable slowdown to environment load (exacerbated if there are multiple environments, in this case 8 copies). This information also requires 50MiB of persistent storage and 50MiB of scratch, so it's a nontrivial committment to RSS as well.

This PR adds a flag to skip this calculation on bodies where it is not needed.


ziyan:
I am not sure about this design of adding a flag in link JSON `skipLinkPairShortestPathCalculation`, or if there is any better way to do it

rschlaikjer:
> I am not sure about this design of adding a flag in link JSON `skipLinkPairShortestPathCalculation`, or if there is any better way to do it

The only other thing that came to mind was making this calculation lazy - that way we just never do it for bodies that never use that data. However I wasn't sure if this delayed latency penalty could cause issues for more traditional robots? It seems unlikely no methods that interact with the kinematics would be called before trying to do a realtime operation, but wasn't sure.

rdiankov:
@rschlaikjer I'm really not a fan of adding a flag. There should be a way to automatically detect "simple" kinbodies and skip a lot of the shortest path logic. Some examples:
1. if every joint has the base link as its attachment.
2. Or if every joint has one of its links as an unmovable link, and all links initially start out as unmovable unless connected to a movable (non-static) joint.
3.

rschlaikjer:
Tried skipping this logic altogether for bodies that seemed simple (e.g. no joints), but some of the tests still call `GetChain` for those simple bodies, which will assert. Instead made it lazy-loaded - bodies that have joints will load the table during `ComputeInternalInformation` as before, bodies that don't but still have callers of `GetChain` will calculate it at first call, and bodies that are decorative (e.g. maps) never generate this information.

include/openrave/kinbody.h
```diff
@@ -3696,7 +3699,16 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
-    std::vector<std::pair<int16_t,int16_t> > _vAllPairsShortestPaths; ///< all-pairs shortest paths through the link hierarchy. The first value describes the parent link index, and the second value is an index into _vecjoints or _vPassiveJoints. If the second value is greater or equal to  _vecjoints.size() then it indexes into _vPassiveJoints.
+    /// All-pairs shortest paths through the link hierarchy.
+    /// The first value describes the parent link index, and the second value is an index into _vecjoints or _vPassiveJoints.
+    /// If the second value is greater or equal to  _vecjoints.size() then it indexes into _vPassiveJoints.
+    /// This value and its associated valid flag are mutable, as they are lazily calculated and so may need to be generated as part of a const method call.
+    mutable std::vector<std::pair<int16_t,int16_t> > _vAllPairsShortestPaths;
+
+    /// Flag to indicate whether the _vAllPairsShortestPaths has been calculated and safely accessed.
+    /// This value is mutable, as the associated data are lazily calculated and so this state may need to change as part of a const method call.
+    mutable bool _vAllPairsShortestPathsValid = false;
```
rdiankov: @rschlaikjer  why not use `_vAllPairsShortestPaths.empty()` for determining if it needs computation instead of introducing a new flag `_vAllPairsShortestPathsValid`?

rschlaikjer: I added the flag for the case where a body has no links - the vector will always be empty even if it's initialized, so there's no reason to spin through the rest of the function

rdiankov: If there's no links, so what if it runs the function every time?

rschlaikjer: fair enough; changed

rdiankov: Great, thanks. This change requires MINOR version update since ABI changes

PR #1439, opened 2024-09-26
Add "override" to allow other modules that include header files from OpenRave to compile

lazydroid:
While working on enabling "-Werror=suggest-override" to strictly follow the Mujin C++ coding standard, a few header files ought to be modified to prevent compiler errors brought up by enabling that warning.

Checked in the pipeline #947098

image

### DIFF
--- a/include/openrave/collisionchecker.h
+++ b/include/openrave/collisionchecker.h
@@ -352,7 +352,7 @@ protected:
     }
 
 private:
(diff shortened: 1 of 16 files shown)


lazydroid:
we have a few open MRs hanging unless we merge this, I think the change set has more or less settled.

rdiankov:
thanks~

PR #1441, opened 2024-09-30
Add new KinBody::ExtraGeometryInfo struture for KinBody::Link::_mapExtraGeometries

zax147:
Add support for serialization and deserialization of the `extraGeometries` field.

The `extraGeometries` in OpenRAVE is saved as `std::map< std::string, std::vector<GeometryInfoPtr> >`  while the schema supposes to be: 

```JSON
         "extraGeometries": [
              {
                  "id": "extraGeometry_id",
                  "name": "extraGeometry_name",
                  "geometries": [
                      {
                          "id": "geom_id",
                          "name": "geom_name",
                      }
                  ]
              }
          ]
```


cielavenir:
just to comment, he is zirui.xie@mujin

src/libopenrave/kinbodylink.cpp
```diff
@@ -337,21 +340,33 @@ void KinBody::LinkInfo::DeserializeJSON(const rapidjson::Value &value, dReal fUn
...
-    //         vgeometries.reserve(it->value.Size() + vgeometries.size());
-    //         size_t iGeometry = 0;
-    //         for(rapidjson::Value::ConstValueIterator im = it->value.Begin(); im != it->value.End(); ++im, ++iGeometry) {
-    //             std::string id = orjson::GetStringJsonValueByKey(*im, "id");
-    //             UpdateOrCreateInfoWithNameCheck(*im, id, vgeometries, "name", fUnitScale);
-    //         }
-    //     }
-    // }
+    if (value.HasMember("extraGeometries") && value["extraGeometries"].IsArray()) {
+        for (rapidjson::Value::ConstValueIterator it_l = value["extraGeometries"].Begin(); it_l != value["extraGeometries"].End(); ++it_l) {
```
woswos: @zax147 Could you use a more relevant and meaningful iterator name other than `it_1`?

src/libopenrave/kinbodylink.cpp
```diff
@@ -337,21 +340,33 @@ void KinBody::LinkInfo::DeserializeJSON(const rapidjson::Value &value, dReal fUn
...
+            }
+            else {
+                continue;
+            }
+
+            if (_mapExtraGeometries.find(name) == _mapExtraGeometries.end()) {
+                _mapExtraGeometries[name] = std::vector<GeometryInfoPtr>();
+            }
+
+            if (it_l->HasMember("geometries")) {
```
woswos: Here we need to deal with merging incoming changes based on the `id` field instead of replacing the data directly.

woswos (changes requested):

src/libopenrave/kinbodylink.cpp
zax147: I committed a new change to address this, and I keep the iterator name as `it` for the consistency with previous fields' iterators.

src/libopenrave/kinbodylink.cpp
zax147: I modified the deserialization to merge incoming changes instead of replacing. Can you review the new commit?

src/libopenrave/kinbodylink.cpp
```diff
@@ -212,24 +212,27 @@ void KinBody::LinkInfo::SerializeJSON(rapidjson::Value &value, rapidjson::Docume
...
-    //     }
-    //     value.AddMember("extraGeometries", extraGeometriesValue, allocator);
-    // }
+    if (_mapExtraGeometries.size() > 0) {
+        rapidjson::Value extraGeometriesValue;
+        extraGeometriesValue.SetArray();
+        FOREACHC(im, _mapExtraGeometries) {
+            rapidjson::Value extraGeometryValue;
+            extraGeometryValue.SetObject();
+            orjson::SetJsonValueByKey(extraGeometryValue, "name", im->first, allocator);
```
woswos: I think you missed handling the `id` field here and in deserialization

snozawa: thanks for offline conversation, I think you can put `id` instead of `name` here.
the `key` of the `std::map` is now represented by `id`.

snozawa: talked in person.
It looks we still need both `id` and `name`.

The user of `id` can handle limited characters.

Meanwhile, we want to use the more characters for the key of the `mapExtraGeometries`.
In the upcoming safety-related geometry handling (c.f. https://github.com/rdiankov/openrave/pull/1402), we want to define the key based on the `KinBody::GetName`, ...etc. Thus, the supported characters should be the same level as `KinBody::GetName` and others.
Therefore, we also need `name` field.

That means, we need to introduce the new data structure for `mapExtraGeometries`...

include/openrave/kinbody.h
```diff
@@ -1331,19 +1359,19 @@ class OPENRAVE_API KinBody : public InterfaceBase
         ///
         /// \param name The name of the geometry group. If name is empty, will initialize the default geometries.
```
justism: can you update the param comments

src/libopenrave/kinbodylink.cpp
```diff
@@ -212,24 +212,28 @@ void KinBody::LinkInfo::SerializeJSON(rapidjson::Value &value, rapidjson::Docume
...
+            rapidjson::Value geometriesValue;
+            geometriesValue.SetArray();
+            FOREACHC(iv, im->second->_vgeometryinfos) {
+                if (!!(*iv)) {
+                    rapidjson::Value geometryValue;
+                    (*iv)->SerializeJSON(geometryValue, allocator, fUnitScale, options);
+                    geometriesValue.PushBack(geometryValue, allocator);
+                }
+            }
+            extraGeometryValue.AddMember("geometries", geometriesValue, allocator);
```
justism: you can just reuse `ExtraGeometryInfo::SerializeJSON`

src/libopenrave/kinbodylink.cpp
```diff
@@ -813,45 +837,48 @@ void KinBody::Link::SetGeometriesFromGroup(const std::string& groupname)
...
     FOREACH(itgeominfo, geometries) {
         if (!(*itgeominfo)) {
             int igeominfo = itgeominfo - geometries.begin();
             throw OPENRAVE_EXCEPTION_FORMAT("GeometryInfo index %d is invalid for body %s", igeominfo % GetParent()->GetName(), ORE_InvalidArguments);
         }
     }
-    std::map<std::string, std::vector<KinBody::GeometryInfoPtr>>::iterator it = _info._mapExtraGeometries.insert(make_pair(groupname, std::vector<KinBody::GeometryInfoPtr>())).first;
-    it->second.resize(geometries.size());
-    std::copy(geometries.begin(), geometries.end(), it->second.begin());
+    KinBody::ExtraGeometryInfoPtr new_extrageom( new KinBody::ExtraGeometryInfo() );
```
justism: please use camel case

rdiankov:
The problem with serializing extra geometry infos is that "self" geometry group gets serialization, which is just a copy of the regular geometry. Furthermore, a lot of other geometries are auto-generated (like padding), so there's really no reason to serialize them as part of the JSON, which then gets into the saved files, and significantly increases the save/load times of the files along with increasing the size on disk.

Perhaps you can begin with talking about the problem you are trying to solve here, and we can think about the right solution.

zax147:
> The problem with serializing extra geometry infos is that "self" geometry group gets serialization, which is just a copy of the regular geometry. Furthermore, a lot of other geometries are auto-generated (like padding), so there's really no reason to serialize them as part of the JSON, which then gets into the saved files, and significantly increases the save/load times of the files along with increasing the size on disk.
> 
> Perhaps you can begin with talking about the problem you are trying to solve here, and we can think about the right solution.

The problem we're solving is that in the latest MCD, we would like to use the  `extraGeometries` (or `geometry groups`)  field in `Webstack`, for example, to store `extraGeometries` in webstack and to selectively show `extraGeometries` on the UI.

We've added the `extraGeometries` field and its related function in webstack, and now trying to match with the changes in Webstack.

rdiankov:
Why does webstack need to handle extra geometries at the moment?

PR #1445, opened 2024-10-03
Add virtual destructors to the classes that have virtual functions

lazydroid:
Classes with virtual functions must have virtual destructors, otherwise these cannot be overriden in the derived classes, leading to memory leaks, pain and suffering. While cleaning up the rest of the code base, there are a few header files from this repository the compiler complained about while building the other modules. Hence this pull request.

No visible regressions

image


yoshikikanemoto:
@lazydroid why not adding the `-Werror=non-virtual-dtor` compiler option and fixing other places too?

lazydroid:
> @lazydroid why not adding the `-Werror=non-virtual-dtor` compiler option and fixing other places too?

These small and benign changes unblock fixing other repositories, adding compiler options and fixing other places will be done in another (much bigger?) MR.

yoshikikanemoto:
this change is ABI breaking. can you increment minor version?

lazydroid:
了解, minor version is incremented. thank you!

yoshikikanemoto (approved):
@rdiankov this PR is ready.

yoshikikanemoto:
merging since this prevents memory leaks from happening with almost zero cost.

lazydroid:
> merging since this prevents memory leaks from happening with almost zero cost.

thank you very much!  :1st_place_medal:

PR #1456, opened 2024-10-25
Minimize null_deleter usage

cielavenir:
null_deleter is against the raison d'etre of shared_ptr. Also it is confusing when we want to make custom factory (instantiater).

Hence it should not be used (unless we need to convert some reference to shared_ptr).

For the full background why I wanted this pull request, please see controllercommon!1108.


cielavenir:
pipeline https://tiny.mujin.co.jp/7nt7z

rdiankov:
The point of using `null_deleter` was to have the object on the stack, not on the heap. We should finding ways to decrease the new allocator from being used. If you want to reduce `null_deleter`, then need to change the API to take in raw pointers...

PR #1459, opened 2024-10-29
Extract isRobot flag from the reference body

justism:
Fixes an issue where `env.ReadRobotURI`  does not load `isRobot` flag from the reference body, eventually causing the loading of the robot body to fail.
```
2024-10-29 10:58:20,503 openrave [WARN] [plugindatabase_virtual.cpp:146 InterfaceBasePtr OpenRAVE::RaveDatabase::Create] env=1 failed to create name GenericRobot, interface kinbody
2024-10-29 10:58:20,503 openrave [WARN] [kinbody.cpp:853 KinBody::InitFromKinBodyInfo] body 'sc2_cameras' interfaceType does not match  != GenericRobot
```


ziyan (approved):
Thanks! I've been looking forward to this for a long time.

Please bump version and changelog.

rdiankov:
thanks~

PR #1460, opened 2024-11-01
WIP: Add print message for non adjacent links

snozawa:
# Summary
- Add print message for non adjacent links. Otherwise, hard to track the self collision issue.
- When `GetNonAdjacentLinks` is called at the first time, print the results. The contents of vector is not small, but i home it's not frequently called and thus, not spamming.
- If `_vInitialLinkTransformations` is modified, print the dof values.


src/libopenrave/kinbody.cpp
```diff
@@ -65,6 +65,21 @@ inline void _ResizeVectorFor2DTable(std::vector<int8_t>& vec, size_t vectorSize)
...
+        return;
+    }
+    std::stringstream ssJoints;
+    for(size_t iDOF = 0; iDOF < vdoflastsetvalues.size(); ++iDOF) {
+        if( iDOF > 0 ) {
+            ssJoints << ",";
+        }
+        ssJoints << vdoflastsetvalues[iDOF];
+    }
+    RAVELOG_INFO_FORMAT("env='%s', body '%s' _vInitialLinkTransformations is updated in %s by dofValues=[%s]", body.GetEnv()->GetNameId()%body.GetName()%context%ssJoints.str());
```
Puttichai: Very minor, but let's print `env=%s` instead of `env='%s'` (no single quotes) so that log messages are consistent with other places.

snozawa: oops, right!

since many of printing like `KinBody`'s `GetName` has single quotes, I misunderstood that we need quotes for env.

thanks~

PR #1464, opened 2024-11-08
Exclude disabled/virtual links from tool speed/accel computation

Puttichai:
# Description

Robot links that are disabled or virtual should not affect planning computation.

Sometimes disabled links may consist of large geometries. These geometries interfere with the tool speed/accel computation such that the computed speed/accel become larger than they could have been.


yoshikikanemoto:
@Puttichai thanks. can you add changelog?

rdiankov:
Thanks!

PR #1465, opened 2024-11-12
clang19 compatibility

lazydroid:
clang19 treats `-Wenum-constexpr-conversion` as an error by default, and gives the following message:
```
/usr/include/boost/mpl/aux_/integral_wrapper.hpp:73:31: error: integer value -1 is outside the valid range of values [0, 3] for the enumeration type 'sign_mixture_enum' [-Wenum-constexpr-conversion]
```

in case of a simple warning adding to the system includes should work with:
```
include_directories(SYSTEM ${OPENRAVE_BOOST_INCLUDE_DIRS})
```
but it does not in this case, hence the need of disabling this particular error directly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,6 +269,9 @@ find_package(Boost REQUIRED COMPONENTS filesystem system thread iostreams)
 set(OPENRAVE_BOOST_INCLUDE_DIRS ${Boost_INCLUDE_DIRS})
 set(OPENRAVE_BOOST_LIB_DIRS ${Boost_LIBRARY_DIRS})
 
+# prevents CLANG from "error: integer value -1 is outside the valid range of values [0, 3] for the enumeration type 'sign_mixture_enum' [-Wenum-constexpr-conversion]"
+include_directories(SYSTEM ${OPENRAVE_BOOST_INCLUDE_DIRS})
+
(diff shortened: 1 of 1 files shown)


lazydroid:
looks good to me...

image

CMakeLists.txt
```diff
@@ -269,6 +272,9 @@ find_package(Boost REQUIRED COMPONENTS filesystem system thread iostreams)
 set(OPENRAVE_BOOST_INCLUDE_DIRS ${Boost_INCLUDE_DIRS})
 set(OPENRAVE_BOOST_LIB_DIRS ${Boost_LIBRARY_DIRS})
 
+# prevents CLANG from "error: integer value -1 is outside the valid range of values [0, 3] for the enumeration type 'sign_mixture_enum' [-Wenum-constexpr-conversion]"
+include_directories(SYSTEM ${OPENRAVE_BOOST_INCLUDE_DIRS})
```
yoshikikanemoto: @lazydroid do you still need this?

lazydroid: @yoshikikanemoto it's not strictly required for suppressing `-Wenum-constexpr-conversion` errors, however, it should help to reduce warning noise coming from Boost and maybe other system includes.

If you think this should be removed, I shall remove it.

lazydroid: @yoshikikanemoto this part is gone.

yoshikikanemoto: ok. thanks

lazydroid:
thank you!

PR #1469, opened 2024-11-19
Add function to set neighstateoptions to use with neighstatefn in jitterer

Puttichai:
# Description

Jitterers use `_neighstatefn` to compute a neighbor state that satisfies user constraints. Depending on the `NeighborStateOptions` given to the function, `_neighstatefn` computes a neighbor differently.

This pull request adds an interface for users to set `NeighborStateOptions` to supply to the jitterer's `_neighstatefn`.


yoshikikanemoto (approved):
LGTM @Puttichai can you update changelog and increment patch version?

PR #1471, opened 2024-11-25
Set OPENRAVE_STD_SCOPED_LOCK and OPENRAVE_STD_STRING_VIEW by config.h

cielavenir:
tl;dr : never use `__cplusplus >` in public headers.

----

Using std::scoped_lock and std::string_view changes ABI. It means that if openrave is compiled using C++17 (or later) the downstream libraries must be compiled using C++17. C++14 or former is vice-versa.

This becomes issues when going back and forth among gcc10-11 / clang15-16.

Hence `__cplusplus >` must not be used in public header and it must be managed by config.h.


include/openrave/environment.h
```diff
@@ -24,19 +24,21 @@
...
 
+#include <openrave/config.h>
+
 #if OPENRAVE_ENVIRONMENT_RECURSIVE_LOCK
-#if __cplusplus >= 201703L
+#if OPENRAVE_STD_SCOPED_LOCK
 #include <mutex>
 using EnvironmentMutex = ::std::recursive_mutex;
-using EnvironmentLock  = ::std::unique_lock<std::recursive_mutex>;
+using EnvironmentLock  = ::std::scoped_lock<std::recursive_mutex>;
```
cielavenir, line 33: this line must be a typo from https://github.com/rdiankov/openrave/pull/1120 , but no one noticed - gcc12 build works as-is.

cielavenir:
pipeline

- master https://tiny.mujin.co.jp/7a9hd
- clang https://tiny.mujin.co.jp/qeq5e

include/openrave/openrave.h
```diff
@@ -54,8 +54,15 @@
 #include <map>
 #include <set>
 #include <string>
+#include <iomanip>
+#include <fstream>
+#include <sstream>
+
+namespace OpenRAVE {
+#include <openrave/config.h>
```
hemangandhi, line 61: Is there a reason this is in the namespace? Feels weird.

cielavenir: for historical reason...? cf https://github.com/rdiankov/openrave/commit/b924fc174b26310cd6fa04e91e6b1b766dcbb3d7#diff-41da4ae7b8b151cbbfccd0e2500d636aba93b78acbe01ffb197a0a23991bca92R120

cielavenir: hopefully other files are not affected, so I fixed the matter.

hemangandhi: Looks like we're actually more consistently `#include`-ing `config.h` outside the namespace now:

```
$ rg -l -tcpp 'openrave/config.h'
include/openrave_c/openrave_c.h
include/openrave/openravejson.h
include/openrave/units.h
include/openrave/openrave.h
include/openrave/openravemsgpack.h
include/openrave/utils.h
include/openrave/openraveexception.h
include/openrave/openravecontainer.h
python/bindings/pyann.cpp
python/bindings/convexdecompositionpy.cpp
python/bindings/docstrings.cpp
python/bindings/include/openravepy/bindings.h

$ rg -l -tcpp --multiline --multiline-dotall 'namespace.+?#include <openrave/config.h>'
include/openrave/openrave.h
```

I don't think this should be a breaking change right now, so may be it's an ok time to make the `#include` consistently outside the `namespace`?

hemangandhi (approved):

cielavenir:
mrcs5 zpl is going to use C++17 and the ABI inconsistency is becoming problematic. Please take a look ASAP.

felixvd:
I agree that this way of checking should be more stable, and I understand that we will need this shortly. I don't see why we wouldn't do this.

felixvd (approved):

rschlaikjer (approved):

yoshikikanemoto:
@cielavenir LGTM. can you bump patch version and update changelog?

cielavenir:
bumped version

Due to `set(CMAKE_CXX_STANDARD 14)`, this can be patch bump.

cielavenir:
Btw I forgot to mention that CMP0067 is to propagate CMAKE_CXX_STANDARD to check_cxx_source_compiles.

PR #1473, opened 2024-11-26
Remove using namespace std from Python

cielavenir:
openrave_int.h is a public header of python binding, so `using namespace std;` should not be used.


cielavenir:
controllercommon etc's python bindings replies on `using namespace std;` already. removing it has too much impact.

cielavenir:
@felixvd @hemangandhi could you fix mujin (controllercommon etc) compilation with this MR?

hemangandhi:
https://tiny.mujin.co.jp/or-1473-internal-test -- the internal changes needed for this.

PR #1475, opened 2024-12-11
Add HasReadable interface

rschlaikjer:
When using the python bindings, we may want to test for the presence of a readable without caring about the contents. Passing just the bool indicator back to python can save some overhead by not having to create new wrapper objects for readable data that is then immediately discarded.


ziyan:
@rschlaikjer Please bump minor version and changelog

rdiankov:
thanks

PR #1476, opened 2024-12-12
Fix C++17 build again

cielavenir:


include/openrave/environment.h
```diff
@@ -30,7 +30,7 @@ namespace OpenRAVE {
 #if OPENRAVE_STD_SCOPED_LOCK
 #include <mutex>
 using EnvironmentMutex = ::std::recursive_mutex;
-using EnvironmentLock  = ::std::scoped_lock<std::recursive_mutex>;
+using EnvironmentLock  = ::std::unique_lock<std::recursive_mutex>;
```
cielavenir: sorry about this, commented https://github.com/rdiankov/openrave/pull/1120/files#r1882322482

cielavenir:
currently my workflow is not blocked, so we could postpone this MR until Debian 13 release....

include/openrave/environment.h
cielavenir: I don't know what is the expected behavior - I would even remove boost path here.

cielavenir: I mean, I do understand that OPENRAVE_STD_SCOPED_LOCK does not make sense compared to std::unique_lock.

rdiankov: Let's make sure there are no regressions here with changing to unique_lock. Thanks

cielavenir:
by making C++17 default https://github.com/rdiankov/openrave/commit/28ab56fc58ae9381dbf7680fe7145c70f624e8ac

- clang-19 pipeline (whose default is -std=gnu++17) https://tiny.mujin.co.jp/i3omp
- github action https://github.com/rdiankov/openrave/actions/runs/12306878959/job/34349436949

however @yoshikikanemoto suggested to remove the ifdef

https://github.com/rdiankov/openrave/pull/1477

PR #1479, opened 2024-12-18
Follow-Up on Connected Body Changes

Puttichai:
# Description

- @cielavenir discovered a problem in the newly introduced `_UpdateConnectedBodyInfo` function. I added a check to skip resolved joint names that are empty strings.
- Some other tests are failing due to unrelated reasons. In the function `_UpdateFromInfo`, there are codes that check and compare two transforms (of links, geometries, etc.) to see if the original body should be updated according to the given infos. The current check, which is done via `CompareTransform` does not take into account the fact that `quat` and `-quat` represents the same rotation. This results in the update function returning `UFIR_RequireReinitialize` instead of `UFIR_NoChange`.
  
  The function that compares two transforms (for the purpose of checking for updates) was originally implemented similar to `TransformDistanceFast`. The function implementation (as well as the function name itself) was changed in https://github.com/rdiankov/openrave/pull/955. The later implementation directly compare values one by one.


src/libopenrave/robotconnectedbody.cpp
```diff
@@ -585,6 +585,10 @@ void RobotBase::ConnectedBody::_UpdateConnectedBodyInfo()
         if( !!pattachedrobot ) {
             for( const std::pair<std::string, RobotBase::JointPtr>& pair: _vResolvedJointNames ) {
                 const std::string& resolvedJointName = pair.first;
+                if( resolvedJointName.empty() ) {
+                    // Can be empty after being deinitialized
+                    continue;
+                }
```
cielavenir, line 591: thank you~

cielavenir (approved):

rdiankov:
thanks

PR #1481, opened 2024-12-21
Fix wrong ComputeInverseDynamics for prismatic joint

snozawa:
Fix wrong ComputeInverseDynamics for prismatic joint.
Unnecessary conversion was multiplied.


rdiankov:
Thanks

PR #1482, opened 2024-12-24
Fix initializing quadratic trajectory in smoother

Puttichai:
### Background

When a parabolic trajectory segment is checked for constraint violation via `CheckPathAllConstraints` (in the function `SegmentFeasible2` in parabolicsmoother), the resulting trajectory segment returned from the check function may be slightly different due to modifications from the check function. Parabolicsmoother handles this slight modification by 

1. computing the acceleration value that is the most consistent with the position and velocity boundary conditions ([here](https://github.com/rdiankov/openrave/blob/production/plugins/rplanners/parabolicsmoother2.cpp#L1209)) and
2. if the computed acceleration value might violate the acceleration limit, the smoother does further adjustment by forcing the ramp's acceleration value to be the acceleration limit.

If the ramp is still consistent with the new acceleration value, then the ramp is accepted as valid.

Even though acceleration values of all ramps are computed and validated by the smoother, they are not getting written to the output trajectory.

### Issue

When a trajectory output from parabolicsmoother is fed back as an input to parabolicsmoother (as part of different computation process, for example), during initialization of ramps, the smoother only does step 1 above (computing the acceleration values most consistent with the boundary conditions) but without the step 2 (adjusting acceleration value in case of limit violation).

This discrepancy between when the ramps were first produced by the smoother and when the ramps are being initialized anew from an OpenRAVE trajectory instance may lead to an error later on.


rdiankov:
@cielavenir did you test for regressions? thanks

cielavenir:
this branch has been tested at latest_test pipeline, and we ran some cycles as morning test

cielavenir:
I ran some cycles at the real cell and I don't think there are issues

cc @ntohge 

----

@Puttichai Is this still draft?

Puttichai:
> I ran some cycles at the real cell and I don't think there are issues
> 
> cc @ntohge
> 
> @Puttichai Is this still draft?

I have been running simulations with these changes and have not noticed any issues so far. So I'm changing the status to ready now. Thanks for helping check it @cielavenir

rdiankov:
thanks~

PR #1483, opened 2024-12-25
Remove back-emf code for torque limit calculation and improving torque limit calculation codes.

snozawa:
- Update for torque-limit calculation codes.
  1. Rmove back emf code
     - There was back emf (electromotive force) code in the production branch (https://github.com/rdiankov/openrave/blob/master/src/libopenrave/kinbodyjoint.cpp#L1995-L2002). However,
        - Several values are hardcoded like `0.9`.
        - The back emf computation is not considered in the all conditions.
        - theoretical formulation is not clear.
     - Since our upcoming robot with dynamics parameter does not need back emf, for now, simply remove the back emf code.
     - If we'll need to consider the back emf effect, we'll revisit the theoretical background and formulation. and then, add the corresponding parameters and revive the code.
  2. Reuse the same utility function for torque calculation to reduce the code duplication. also, do slight code optimization.
- Add test cases in the downstream repos. This PR does not change the behavior except for `1. Remove back emf code`.


cielavenir:
I agree as https://github.com/rdiankov/openrave/pull/554 is left befind for 6 years.

rdiankov:
thanks~

PR #1484, opened 2025-01-07
Skip grab saver for unjointed bodies

rschlaikjer:
- When grabbing a body with a body that has no joints, we can skip saving the current state of the body since all links are effectively rigid, and so would be excluded from the results of `ComputeListNonCollidingLinks` (all links are rigidly attached to the grabbing link)
- It doesn't look like we need to save velocities at all for the  `ComputeListNonCollidingLinks` savers (has no impact on collision?)
- Convert a couple of shared pointer copies to shared pointer references in `ComputeInternalInformation`


rdiankov:
@snozawa Can you approve this? Thanks

src/libopenrave/kinbodygrab.cpp
```diff
@@ -118,17 +118,29 @@ Grabbed::Grabbed(KinBodyPtr pGrabbedBody, KinBody::LinkPtr pGrabbingLink)
...
+    if (pGrabber->GetJoints().size() + pGrabber->GetPassiveJoints().size() == 0) {
+        // This body is fully rigid: all links are excluded from the non colliding link set, so just flag that the list is valid (empty) and return.
+        _listNonCollidingIsValid = true;
+        return;
+    }
+
+    // if this body is not rigid, then we need to save the current state of the grabber/grabbee so that we can roll back to it when computing the non-colliding links
     _listNonCollidingIsValid = false;
     const bool bDisableRestoreOnDestructor = true; // This is very important! These saver are used only in ComputeListNonCollidingLinks and we don't want to restore on destructor.
+    static const int saverOptions = KinBody::Save_LinkTransformation | KinBody::Save_LinkEnable | KinBody::Save_JointLimits | KinBody::Save_ConnectedBodies;
```
snozawa: thanks @rschlaikjer !

Apart from `Save_LinkVelocities`, why you added `Save_ConnectedBodies`?

snozawa: Also, `Save_LinkVelocities` are removed from constructor, but still in `ComputeListNonCollidingLinks`.

Is this just in case?

kanbouchou: @rdiankov cc @rschlaikjer While checking this code, I noticed that `KinBody::ComputeInverseDynamics` doesn't seem to care about grabbed. Is it intentional that `ComputeInverseDynamics` doesn't consider grabbed?

src/libopenrave/kinbodygrab.cpp
```diff
@@ -118,17 +118,29 @@ Grabbed::Grabbed(KinBodyPtr pGrabbedBody, KinBody::LinkPtr pGrabbingLink)
...
     _CreateSaverForGrabbedAndGrabber(_pGrabbedSaver,
                                      pGrabbedBody,
-                                     KinBody::Save_LinkTransformation|KinBody::Save_LinkEnable|KinBody::Save_JointLimits,
+                                     saverOptions,
```
kanbouchou, line 139: If someone tries to set velocity of a currently grabbed kinbody, should we reject it because its velocity is determined by the grabbing link? If such guarantee exists that body's velocity doesn't get modified while being grabbed, it is fine  to not save at the time of grabbing and restore at the time of releasing. @rdiankov @rschlaikjer

rdiankov: Should be okay if velocity is not saved..

src/libopenrave/kinbodygrab.cpp
rschlaikjer: Hmm, not sure where that came from - might have been an intermediate change. Since it gets added in `_CreateSaverForGrabbedAndGrabber` for robots shouldn't be needed as you say, removed.

For velocities, I think you're right - if we don't save/restore it using the grab saver, it should be fine to not save/restore it in ComputeListNonColliding.

rschlaikjer: Internal tests re-run and pass using controllercommon MR 1185 pipeline 1072255

rdiankov:
thanks~

PR #1485, opened 2025-01-16
Add combined SetTransformAndVelocity method

rschlaikjer:
When setting the transform/velocity of a body that is grabbing other bodies, _UpdateGrabbedBodies will call SetTransform / SetVelocity on all grabbed bodies, _each_ of which will then recursively call _UpdateGrabbedBodies on all of their grabbed bodies, etc. 

Since more calls are generated at each level, for heavily nested grabs, the exponential increase in duplicated calls becomes untenable. 

Allowing for a single update position + velocity call that only calls update grabbed bodies _once_ removes this expontential time problem.


rdiankov:
does it pass test_openrave?

rdiankov:
will need MINOR version update

rschlaikjer:
> will need MINOR version update

This should already be pushed 

https://github.com/rdiankov/openrave/pull/1485/commits/032bec1a94b04094067ef1589947875b9b7e23b2

rdiankov:
thanks!

PR #1486, opened 2025-01-17
Check validity of mesh indices

snozawa:
# Summary
- Add checking of mesh indices.

# Issue
- When mesh is set from convexdecomposed geometry, segfault happens from fcl.
- The cache file of the convexdecomposed geometry contains invalid indices (index out of range and  negative index)
- Although the root cause of the invalid cache file is not figured out yet since it's not reproduced, it's better to have a protection against such invalid geometry data.

# Resolution
- Throw exception if `indices` is invalid when `GeometryInfo` is made, or `_mapExtraGeometry` is set.
- Add the similar checking when `GeometryInfo` is created from convexdecompositionmodel


kanbouchou:
Looks good to me. @rdiankov please review and merge, this is relatively small and local changes.

rdiankov:
thanks~

PR #1488, opened 2025-01-29
Use PyUnicode_GET_LENGTH

cielavenir:
PyUnicode_GET_SIZE was removed in Python 3.13.


cielavenir:
autotester: https://tiny.mujin.co.jp/3g9qj

rdiankov:
thanks~

PR #1490, opened 2025-01-30
Fix using wrong quadratic solution for timestep

Puttichai:
### Background

In each iteration of the `Check` function, we solve for the next time step to check for constraints by solving a quadratic equation. The next time step to do constraint checking is the time step such that the joint `nLargestStepIndex` moves for an amount of its resolution.

Relevant codes are as follows
```c++
            dReal fMinNextTimeStep = prevtimestep; // when computing the new timestep, have to make sure it is greater than this value
            if( bComputeNewStep ) {
                if( prevtimestep >= fLargestInflectionTime ) {
                    fBestNewStep = fStep-fLargestStepDelta;
                }
                else {
                    // could be straddling the inflection so have to compensate. note that this will skip checking collision at the inflection
                    if( (fLargestStepDelta > 0 && fStep+fLargestStepDelta > fLargestInflection) || (fLargestStepDelta < 0 && fStep+fLargestStepDelta < fLargestInflection) ) {
                        fBestNewStep = fLargestInflection - (fStep+fLargestStepDelta - fLargestInflection);
                        fMinNextTimeStep = fLargestInflectionTime-1e-7; // in order to force to choose a time after the inflection
                    }
                    else {
                        fBestNewStep = fStep+fLargestStepDelta;
                    }
                }
            }
```

When the current time step is near the inflection time and we expect to the next time step to be after the inflection time, `fMinNextTimeStep` should be set to `fLargestInflectionTime`. The value `fMinNextTimeStep` will be used later on to choose an appropriate solution of the following quadratic equation
```c++
int numroots = mathextra::solvequad(fLargestStepAccel*0.5, fLargestStepInitialVelocity, -fNewStep, timesteproots[0], timesteproots[1]);
```

### The Issue

`fMinNextTimeStep` should stay the same for the same `istep`. However, the following situation happens

- Iteration `j`
  - `bComputeNewStep` is `true` so we proceed to compute `fBestNewStep` as usual.
  - It happens that `fMinNextTimeStep` is also modified to be the inflection time `fLargestInflectionTime`.
  - The next time step for this iteration is computed to be `t_next > fLargestInflectionTime`.
  - At the end, it happens that `istep` does not get incremented because the reached time step is `t_reached < fLargestInflectionTime`. `bComputeNewStep` is now `false` because we should be using the same `fBestNewStep` in the following iteration.
- Iteration `j + 1`
  - `istep` remains the same as in iteration `j`. But `fMinNextTimeStep` gets reset to `prevtimestep` (which is now `t_reached`).
  - The next time step for this iteration is computed to be `t_next < fLargestInflectionTime` (due to the wrongly set `fMinNextTimeStep`.
  - At the end of the iteration, `istep` gets incremented while not having reached `fBestNewStep` yet. This messes up the counting of `istep`.
- In the end, `istep` can reach `numSteps` while the function does not yet reach the end point of the ramp.

This causes an error in parabolic smoother (function `SegmentFeasible2`) when it checks if the returned configurations from the `Check` function actually reached the desired values.

### The Fix

Make sure that `fMinNextTimeStep` gets computed correctly even when `bComputeNewStep = false`.

### Changelog

```
version-bump-type: patch   # (can be "minor", "patch" or "none")
version-file: CMakeLists.txt
version-file: docs/source/changelog.rst

Version 0.16x.x
===============

- In `Check` function, correctly compute `fMinNextTimeStep` when `bComputeNewStep = false`.
```


rdiankov:
@snozawa always scary to touch this function.. please review . thanks

snozawa:
> @snozawa always scary to touch this function.. please review . thanks

sure, i'll check!
(it might take time since the change is not trivial though)

rdiankov:
@Puttichai can we add more regression tests for this specific function? it's pretty easy to accidentally break it..

rdiankov:
@snozawa have you reviewed this pull request? can we get more regression tests for this function?  thanks

Puttichai:
> @Puttichai can we add more regression tests for this specific function? it's pretty easy to accidentally break it..

I have already added a test case for this specific scenario that is fixed in this pull request. I will try to add more tests on general scenarios.

rdiankov:
@kanbouchou @snozawa are you okay with merging this feature?

cielavenir:
It is unfortunate but I would need to revert changelog update if Shunichi/Kei's review does not come this week.

cc @ntohge

cielavenir:
Sorry for sudden notice but I decided to REVERT CHANGELOG as my batch-merger raises error.

@felixvd @hemangandhi better system to update-changelog for github repos is highly welcomed. I'd like it for the "next iteration"

cielavenir:
@ntohge If this incident affects my evaluation (especially teamwork part), I need to accept.

cielavenir:
now my batch merger worked, though it sacrificed honor :cry:

rdiankov:
@snozawa this merge request has been pending for 4 months now, need your input here. thanks

snozawa:
> @snozawa this merge request has been pending for 4 months now, need your input here. thanks

oh sorry for being late!
yea, if you want to prioritize this, i'll take a look soon!

snozawa:
@Puttichai 

btw, not directly related to this PR.
but i found potential issue in `dqscale` computation.

```
int numroots = mathextra::solvequad(_vtempaccelconfig[idof]*0.5, dq0[idof],q0[idof] - _vtempconfig[idof] - vConfigResolution[idof], timesteproots[0], timesteproots[1]);
```

here, `vConfigResolution[idof]` is subtracted.

but, if we think about the condition to satisfy, e.g. `RaveFabs(dQ[idof]) > vConfigResolution[idof]*1.01`, the sign of `dQ` might be positive or negative.
if so, the offset we should take into account might be `vConfigResolution` or `-vConfigResolution`.

snozawa:
@Puttichai 

overall, it looks the change in this PR is reasonable for me.

I'm not sure how we can thoroughly test this, especially because this `Check` function involves non-straightforward factor such as `_neighstatefn`.
Let's continue discussing how  to test this in internal page.

Puttichai:
Thanks for checking.

> overall, it looks the change in this PR is reasonable for me.

Then I think we can merge this fix along with the test I have added in the internal repo to protect this case. We can address any other potential issues separately and independently from this issue.

rdiankov:
Thanks guys~ let's be on the lookout for regressions

PR #1492, opened 2025-03-12
AddGripperInfo with removeduplicate=true should replace old GripperInfo with new

Puttichai:
# Description

From the current implementation in `production`, `Robot::AddGripperInfo` with `removeduplicate=true` actually adds the new `gripperInfo` to the robot when `gripperInfo` with the same name already exists.

This pull request corrects the implementation so that if a `gripperInfo` with the same name already exists, the new `gripperInfo` will replace the old one.


yoshikikanemoto:
thanks @Puttichai

PR #1494, opened 2025-03-25
add _grippername to GrabbedInfo

y423610m:
add _grippername to GrabbedInfo and Grabbed


include/openrave/kinbody.h
```diff
@@ -3931,6 +3932,7 @@ class OPENRAVE_API Grabbed : public UserData, public boost::enable_shared_from_t
     // Member Variables
     KinBodyWeakPtr _pGrabbedBody; ///< the body being grabbed
     KinBody::LinkPtr _pGrabbingLink; ///< the link used for grabbing _pGrabbedBody. Its transform (as well as the transforms of other links rigidly attached to _pGrabbingLink) relative to the grabbed body remains constant until the grabbed body is released.
+    std::string _grippername;
```
kanbouchou: Please add docstring comment to new members and new parameters.

python/bindings/openravepy_kinbody.cpp
```diff
@@ -4848,21 +4854,23 @@ class GrabbedInfo_pickle_suite
...
     static py::tuple getstate(const PyKinBody::PyGrabbedInfo& r)
     {
-        return py::make_tuple(r._grabbedname, r._robotlinkname, r._trelative, r._setIgnoreRobotLinkNames);
+        return py::make_tuple(r._grabbedname, r._robotlinkname, r._grippername, r._trelative, r._setIgnoreRobotLinkNames);
     }
     static void setstate(PyKinBody::PyGrabbedInfo& r, py::tuple state) {
 #ifdef USE_PYBIND11_PYTHON_BINDINGS
         r._grabbedname = extract<std::string>(state[0]);
         r._robotlinkname = extract<std::string>(state[1]);
+        r._grippername = extract<std::string>(state[2]);
```
kanbouchou: Isn't it better to avoid using existing index (2)? What happens if you open a scene saved with old version of openrave which doesn't have gripper name? Don't you end up trying to parse relative transform (previously stored at index 2) as gripper name?

y423610m: @kanbouchou fixed to assign `_grippername` to state[4].

include/openrave/kinbody.h
y423610m: added

kanbouchou:
@y423610m thanks, looks good to me. Please prepare change log and version bump.

@rdiankov please review.

y423610m:
updated changelog and bumped version

CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ set( CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS TRUE )
 # Define here the needed parameters
 set (OPENRAVE_VERSION_MAJOR 0)
 set (OPENRAVE_VERSION_MINOR 163)
```
rdiankov, line 7: @y423610m needs minor version update since ABI changes, not patch

kanbouchou: @rdiankov cc @y423610m handled in https://github.com/rdiankov/openrave/pull/1501/commits/00b62948bc2f43bd2209f7d8589347da33e26e0c

kanbouchou:
@rdiankov please check https://github.com/rdiankov/openrave/pull/1501 instead of this.

PR #1497, opened 2025-04-03
Do not set 4th element of diffuseColor to 1

Puttichai:
### Description

In helper functions for initializing KinBody, we currently set `_vDiffuseColor` to `Vector(x, y, z, 1)`. The last (4-th) element of the diffuse color being non-zero leads to the serialized value of the diffuse color being a vector of length 4 instead of 3.

```python
import json
box = RaveCreateKinBody(env, '')
box.InitFromBoxes([[0, 0, 0, 1, 1, 1]])
box.SetName('box')
env.Add(box)
data = json.loads(env.WriteToMemory('json'))
print( data['bodies'][0]['geometries'][0]['diffuseColor'] )

>>> [1.0, 0.5, 0.5, 1.0]
```

The serialized environment info cannot be validated with the schema due to failure caused by `diffuseColor` having an inappropriate size.


cielavenir:
Could you add a testcase?

Also, is SerializeJson() affected?

Puttichai:
@cielavenir 

> Could you add a testcase?

Will do shortly.

> Also, is SerializeJson() affected?

We wouldn't see any differences for codes running in python unless it specifically uses `env.WriteToMemory`.

In python, if we do
```python
env.ExtractInfo().SerializeJSON()
```
we'll actually see `diffuseColor` being a 3-vector. This is because `PyGeometryInfo` is implemented like this
```python
_vDiffuseColor = toPyVector3(info._vDiffuseColor);
```

cielavenir:
@Puttichai understood thank you for explanation!

yoshikikanemoto:
confirmed that we don't/shouldn't have any usages of 4th element of diffuseColor and ambientColor. merging.

PR #1496, opened 2025-04-02
Run CI using ubuntu24

cielavenir:
Ubuntu 20 runner will not be available after Apr 15: https://github.com/actions/runner-images/issues/11101

Also, as a part of the Debian 13 migration, we would like to use Ubuntu24.

PR #1500, opened 2025-04-03
Avoid N^2 loops in UpdateFromInfo

rschlaikjer:
Previously, searching for bodies with the same name/id would do a linear search on the bodies in the environment.
Replace with hash lookups to avoid slow iteration for high-body-count environments.

Test cases to confirm behaviour matches existing production have been added in testopenrave MR 141.


rschlaikjer:
While writing tests, noticed a potentially confusing behaviour in current production - 
https://github.com/rdiankov/openrave/pull/1500/files#diff-c5c3a9bad867a7032c72acc1708c24dde1866ad3d1c43ae5fca006461105f850R3238

If we have an update that contains two kin body infos with the same name, we will create two bodies (one renamed to be unique) but then fail this assert that expects both bodies to match the names that were present in the info. Should this assert be removed?

rdiankov:
@cielavenir has this gone through the openrave testing pipeline?

src/libopenrave-core/environment-core.h
```diff
@@ -2986,181 +2986,170 @@ class Environment : public EnvironmentBase
...
-                    if (updateMode == UFIM_OnlySpecifiedBodiesExact) {
-                        usedBodyIndexSet.emplace(nMatchingIndex);
-                    }
+                // If we matched by ID instead of name, it's possible there exists another body in the env with the same name as our body info.
+                // This would cause a conflict if we try and update our current (id matched body) to have the same name.
+                // Since the other body with the same name might get processed again later (by id?), temporarily rename it so that we can continue.
+                if (!!pExistingBody && existingBodyIndex != existingBodyIndexSameName && existingBodyIndexSameName >= 0) {
+                    const KinBodyPtr& pExistingBodySameName = vBodies[existingBodyIndexSameName];
+                    RAVELOG_DEBUG_FORMAT("env=%s, have to clear body name '%s' id=%s for loading body with id=%s", GetNameId() % pExistingBodySameName->GetName() % pExistingBodySameName->GetId() % pExistingBody->GetId());
+                    pExistingBodySameName->SetName(_GetUniqueName(pExistingBodySameName->GetName() + "_tempRenamedDueToConflict_"));
```
rdiankov: @rschlaikjer  Changing name of an existing body invalidates the `string_view` you are managing here, which could corrupt your state.

```c++
        std::unordered_set<string_view> bodyIdsToUpdate; // string_views are over info._vBodyInfos, guaranteed lifetime for the rest of this function
        std::unordered_set<string_view> bodyNamesToUpdate;
```

rdiankov: There are also these variables that use string_view which might also be affected:
```c++
                const std::unordered_map<string_view, int>::iterator itExistingBodyIndexById = existingBodyIndicesById.find(kinBodyInfo._id);
                const std::unordered_map<string_view, int>::iterator itExistingBodyIndexByName = existingBodyIndicesByName.find(kinBodyInfo._name);
```

rschlaikjer: Those first views are over the body _info_ record name strings (`info._vBodyInfos`, per the comment), so they don't need to worry about the data getting changed (info is immutable for this function).

As for the second one, good catch - we handle removing the reused body, but not any fallout from renaming any _non_ matched colliding bodies.

cielavenir:
@rdiankov now openrave ci needs https://github.com/rdiankov/openrave/pull/1496

cielavenir:
seems this one is ok

src/libopenrave-core/environment-core.h
```diff
@@ -2986,181 +2986,179 @@ class Environment : public EnvironmentBase
...
             RobotBase::RobotBaseInfoConstPtr pRobotBaseInfo = OPENRAVE_DYNAMIC_POINTER_CAST<const RobotBase::RobotBaseInfo>(pKinBodyInfo);
-            KinBodyPtr pMatchExistingBody; // matches to pKinBodyInfo
-            int bodyIndex = -1; // index to vBodies to use. -1 if not used
-            {
-                // find existing body in the env
-                std::vector<KinBodyPtr>::iterator itExistingSameId = vBodies.end();
-                std::vector<KinBodyPtr>::iterator itExistingSameName = vBodies.end();
-                std::vector<KinBodyPtr>::iterator itExistingSameIdName = vBodies.end();
-
-                if( updateMode == UFIM_OnlySpecifiedBodiesExact ) {
```
woswos, line 3040: Is `UFIM_OnlySpecifiedBodiesExact` correctly handled in the new version? I think there is a behavior change in the API after these changes.

src/libopenrave-core/environment-core.h
```diff
@@ -3305,73 +3297,64 @@ class Environment : public EnvironmentBase
...
-            // remove extra bodies at the end of vBodies
-            if( vBodies.size() > info._vBodyInfos.size() ) {
-                ExclusiveLock lock009(_mutexInterfaces);
-                for (std::vector<KinBodyPtr>::iterator itBody = vBodies.begin() + info._vBodyInfos.size(); itBody != vBodies.end(); ) {
-                    KinBodyPtr pBody = *itBody;
-                    if (!pBody) {
-                        ++itBody;
-                        continue;
-                    }
-                    RAVELOG_VERBOSE_FORMAT("remove extra body env=%s, id=%s, name=%s", GetNameId()%pBody->_id%pBody->_name);
```
woswos, line 3341: Can we keep the logs?

src/libopenrave-core/environment-core.h
rschlaikjer: Can you elaborate? The `UFIM_OnlySpecifiedBodiesExact` case should be handled down on line 3301. See testopenrave PR 141 for the relevant tests.

src/libopenrave-core/environment-core.h
ziyan: Added back.

src/libopenrave-core/environment-core.h
ziyan: I think the old logic here was to ignrore duplicated matching in case of `UFIM_OnlySpecifiedBodiesExact`, i.e. if the caller passed in a list of kinbody info that has duplicates. I think that's automatically an error, which is guarded by exceptions added in the beginning of the function.

ziyan:
LGTM

ziyan (approved):

rdiankov:
@rschlaikjer  `vBodies` was not necessary and was getting copied. Furthermore, it was introducing two sets of BodyIndices, one for vBodies and one for _vecbodies, which diverge as update happens and makes the code harder to understand.
By getting rid of `vBodies` we save an expensive copy of all the bodies in the scene and can remove code that managed vBodies.
Please review the changes, thanks.

rschlaikjer:
Changes look good, thanks. Updated a couple comments to reflect the new behaviour.

rdiankov:
Thanks~

PR #1502, opened 2025-04-09
Avoid N^2 search in DeserializeJSONWithMapping

rschlaikjer:
Similar to #1500, don't perform double-for-loop search over all body infos. Create a lookup table and just use that. 

Additional tests to exercise this function have been added to the same testopenrave PR as #1500 (MR 141).


rdiankov:
@cielavenir has this gone through the openrave testing pipeline?

src/libopenrave/environment.cpp
```diff
@@ -183,124 +244,87 @@ void EnvironmentBase::EnvironmentBaseInfo::DeserializeJSONWithMapping(const rapi
...
-                pKinBodyInfo->DeserializeJSON(rKinBodyInfo, fUnitScale, options);
-                pKinBodyInfo->_id = id;
+                _DeserializeKinBodyInfo<KinBody::KinBodyInfo>(
+                    rKinBodyInfo, fUnitScale, options, isDeleted, existingBodyIndex, pId,
+                    /* mutable */ _vBodyInfos, /* mutable */ existingBodyInfoIndicesById, /* mutable */ existingBodyInfoIndicesByName);
             }
         }
+
+        // Now that we're done processing, filter _vBodyInfos to remove any holes left from deleting bodies
+        _vBodyInfos.erase(std::remove_if(_vBodyInfos.begin(), _vBodyInfos.end(), [](const KinBody::KinBodyInfoPtr& p) { return !p; }), _vBodyInfos.end());
```
rdiankov: Can you re-write this in an optimized way without using lambda functions and std::remove_if?

rschlaikjer: Is the objection `remove_if` or the lambda? `remove_if` is already O(N), there isn't much reason to re-write it. Would you accept a struct based predicate instead? I have to say that the aversion to lambdas remains difficult to understand. 
```diff
         // Now that we're done processing, filter _vBodyInfos to remove any holes left from deleting bodies
-        _vBodyInfos.erase(std::remove_if(_vBodyInfos.begin(), _vBodyInfos.end(), [](const KinBody::KinBodyInfoPtr& p) { return !p; }), _vBodyInfos.end());
+        struct IsValidPointer {
+            bool operator()(const KinBody::KinBodyInfoPtr& p) { return !!p; }
+        };
+        _vBodyInfos.erase(std::remove_if(_vBodyInfos.begin(), _vBodyInfos.end(), IsValidPointer{}), _vBodyInfos.end());
```

rschlaikjer:
I notice that you changed the boxing of id from std::string to const char* - what is your opinion on using a view instead? Using a raw const char* avoids a copy compared to std::string, but is slower for many operations (since there's no cached length). String views can take the known length from the rapidjson value to get the speed benefit without copying the underlying data.

rdiankov:
If you can make it work with string_view, then go for it

src/libopenrave/environment.cpp
rdiankov: Because usually there's a way to write clearer code without lambdas..
```c++
_vBodyInfos.erase(std::remove(_vBodyInfos.begin(), _vBodyInfos.end(), KinBody::KinBodyInfoPtr()), _vBodyInfos.end());
```

rdiankov:
thanks~ let's see how much this one will get us in trouble

cielavenir:
@rschlaikjer could you fix this? I saw you removed `robot.serialize`

```
2025-04-29T23:32:38.2945848Z ======================================================================
2025-04-29T23:32:38.2945971Z ERROR: test_kinematics.TestKinematics.test_hashes
2025-04-29T23:32:38.2946089Z ----------------------------------------------------------------------
2025-04-29T23:32:38.2946236Z Traceback (most recent call last):
2025-04-29T23:32:38.2951112Z   File "/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/nose/case.py", line 171, in runTest
2025-04-29T23:32:38.2951234Z     self.test(*self.arg)
2025-04-29T23:32:38.2951309Z     ~~~~~~~~~^^^^^^^^^^^
2025-04-29T23:32:38.2951625Z   File "/home/runner/work/openrave/openrave/test/test_kinematics.py", line 716, in test_hashes
2025-04-29T23:32:38.2951764Z     s = robot.serialize(SerializationOptions.Kinematics)
2025-04-29T23:32:38.2951839Z         ^^^^^^^^^^^^^^^
2025-04-29T23:32:38.2952112Z AttributeError: 'openravepy._openravepy_.openravepy_int.Robot' object has no attribute 'serialize'
2025-04-29T23:32:38.2952120Z 
2025-04-29T23:32:38.2952210Z ----------------------------------------
2025-04-29T23:32:38.2952290Z Ran 152 tests in 1341.325s
```

rschlaikjer:
> @rschlaikjer could you fix this? I saw you removed `robot.serialize`

Fix: https://github.com/rdiankov/openrave/pull/1505

PR #1503, opened 2025-04-11
Add warning when self-collision checker does not exist

Puttichai:
# Description

Generally, the environment-collision checker and the robot self-collision checker might be using different geometry groups for collision checking. Using the environment-collision checker as a substitute for the missing self-collision checker may not give desired results.

At least, when the self-collision checker is missing, prints some warning.

### DIFF
--- a/src/libopenrave/kinbodycollision.cpp
+++ b/src/libopenrave/kinbodycollision.cpp
@@ -53,6 +53,7 @@ bool KinBody::CheckSelfCollision(CollisionReportPtr report, CollisionCheckerBase
                 // no checker set
                 return false;
             }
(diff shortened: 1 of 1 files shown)


yoshikikanemoto:
I am afraid this might cause spamming log message on particular usages.

src/libopenrave/kinbodycollision.cpp
```diff
@@ -53,6 +53,7 @@ bool KinBody::CheckSelfCollision(CollisionReportPtr report, CollisionCheckerBase
                 // no checker set
                 return false;
             }
+            RAVELOG_WARN_FORMAT("env=%s, self-collision checker is not set for '%s' so using env collision checker instead", GetEnv()->GetNameId() % GetName());
```
rdiankov, line 56: this could spam quite a lot, perhaps put the log in another place

Puttichai:
I'm closing this pull request. A check has already been added at the caller side.

PR #1504, opened 2025-04-14
Optimizations for environment syncing

rschlaikjer:
- When calculating kinbody geometry hashes, formatting all of the constituent data as a string _then_ hashing it spends a lot of time doing float formatting. Instead, just digest all the raw data without the string pass.
  - If this was in fact the only thing using `serialize` (lowercase), as the comment implies, maybe it can be removed?
- When removing bodies from the environment, we iterate the entire list of _other_ bodies to check if they had custom collision checkers. If you remove a lot of bodies from an env with a lot of bodies in it, this means a lot of traversals of this list. Since a body can only exist in a different body's collision checker if it was grabbed at some point, add a bit to track whether a body was ever grabbed, and only perform this search for those bodies.
- When writing tests, noticed that InitGeometries takes a mutable reference, when it doesn't seem to need to? Made const.


include/openrave/kinbody.h
```diff
@@ -853,6 +855,7 @@ class OPENRAVE_API KinBody : public InterfaceBase
         }
 
         void serialize(std::ostream& o, int options) const;
+        void digest(HashContext& hash, int options) const;
```
rdiankov: can we have a better function name like `DigestHash`?

src/libopenrave/kinbodygrab.cpp
```diff
@@ -350,6 +350,9 @@ bool KinBody::Grab(KinBodyPtr pGrabbedBody, LinkPtr pGrabbingLink, const std::se
     OPENRAVE_ASSERT_FORMAT(pGrabbingLink->GetParent().get() == this, "env=%s, pGrabbingLink name='%s' for grabbing '%s' is not part of body '%s'", GetEnv()->GetNameId()%pGrabbingLink->GetName()%pGrabbedBody->GetName()%GetName(), ORE_InvalidArguments);
     OPENRAVE_ASSERT_FORMAT(pGrabbedBody.get() != this, "env=%s, body '%s' cannot grab itself", GetEnv()->GetNameId()%pGrabbedBody->GetName(), ORE_InvalidArguments);
 
+    // Flag that the body we're grabbing has been grabbed and may be tracked by more collision checkers than just the env checker
+    pGrabbedBody->_wasEverGrabbed = true;
```
rdiankov, line 354: @snozawa @kanbouchou can you take a look at this usage of  `_wasEverGrabbed`? Thanks

rdiankov:
Let's also try to remove `serialize`, I don't think we use it for anything else. Thanks

PR #1506, opened 2025-04-30
Fix hashing problems with grab info

rschlaikjer:
Not entirely sure how/whether the `_rGrabbedUserData` serialization actually worked before... checked before my changes and it was still only serializing if the field was null.


rdiankov:
Let's do a better job of testing features and getting people to do reviews.

PR #1507, opened 2025-05-01
Allow fetching only bodies that have a certain readable

rschlaikjer:
Performing this filtering in C++ is much better than copying _all_ bodies to python callers and then having _them_ do the filtering


rdiankov:
Perhaps, similar to the IterateBodies pull request you have, there could be many types of filters, so perhaps user can pass in a `std::function<bool(KinBody&body)>` to decide whether they want to get that body.

rschlaikjer:
> Perhaps, similar to the IterateBodies pull request you have, there could be many types of filters, so perhaps user can pass in a `std::function<bool(KinBody&body)>` to decide whether they want to get that body.

This is probably useful as well, I might add this in addition to the current filter instead of replacing it just because this one is a little better for pybind callers (no invoking function pointers back into python land that then have to cross back into c++)

rdiankov:
GetBodiesWityReadarInterface can be just defined for the python bindings. There is no reason to define them also in environment.h

rschlaikjer:
> GetBodiesWityReadarInterface can be just defined for the python bindings. There is no reason to define them also in environment.h

Sure, updated

yoshikikanemoto (approved):
LGTM

ziyan (approved):

src/libopenrave-core/environment-core.h
```diff
@@ -1655,11 +1655,30 @@ class Environment : public EnvironmentBase
         }
     }
 
+    void GetBodiesMatchingFilter(std::vector<KinBodyPtr>& bodies, const std::function<bool(const KinBody&)>& filterFunction, uint64_t timeout = 0) const override
+    {
+        TimedSharedLock _lockMutexInterfaces(_mutexInterfaces, timeout);
```
ziyan: Why prefix `_` for local var?

rdiankov:
thanks~

PR #1508, opened 2025-05-02
Provide load context to cache documents across calls

rschlaikjer:


ziyan (approved):
LGTM.

rdiankov:
@rschlaikjer  It's not good to expose details of cache in header files, it will force users to recompile everytime the implementation changes.
Instead, it's better for `EnvironmentLoadContext` to be a virtual base class, and the internal implementation to somewhere in libopenrave-core.
I did that by creating `EnvironmentLoadContextJSON`. Also to decrease amount of shared pointer dereferencing or creating load contexts on the fly, the RaveParseXXX functions take a reference to `EnvironmentLoadContextJSON` now.

ziyan:
LGTM.

PR #1509, opened 2025-05-06
Recursive mutex with gil check

eisoku9618:
Add a custom mutex that can predict a potential deadlock between Python GIL and environment mutex. I would like to use this mutex on testing phase of our codebase and detect potential deadlocks at runtime during testing.

Also release GIL on `pyRaveDestroy()` so that this new custom mutex does not raise an exception on `pyRaveDestroy()` as `pyRaveDestroy()` internally locks the env mutexes.

Here is a sample code that causes the deadlock. This custom mutex can raise an exception in this case instead of doing deadlock.

```python
import openravepy
import threading
import time

def Thread1(env):
    print('thread 1: entering with env')
    with env:
        print('thread 1: releasing GIL with env locked')
        time.sleep(2)
        print('thread 1: calling CheckCollision with env locked')
        env.CheckCollision(env.GetBodies()[0])

def Thread2(env):
    time.sleep(1)
    print('thread 2: GIL is acquired. Calling CheckCollision without locking env. Note that CheckCollision locks env internally. -> deadlock')
    env.CheckCollision(env.GetBodies()[0]) # env is locked inside CheckCollision

def main():
    env = openravepy.Environment(0)
    box = openravepy.RaveCreateKinBody(env, '')
    box.SetName('test')
    env.AddKinBody(box)

    t0 = threading.Thread(target=Thread1, args=[env])
    t1 = threading.Thread(target=Thread2, args=[env])
    t0.start()
    t1.start()
    t0.join()
    t1.join()

if __name__ == '__main__':
    main()
```


eisoku9618:
Note to self: Releasing GIL on env destroy does not make `Environment::Destroy()` thread safe because this function locks `_mutexInit` first then env mutex.  I attached a sample code to show that.

```python
import openravepy
import threading
import time

def Thread1(env):
    print('thread 1: entering with env')
    with env:
        print('thread 1: releasing GIL with env locked')
        time.sleep(2)
        print('thread 1: calling Destroy with env locked')
        env.Destroy()

def Thread2(env):
    time.sleep(1)
    print('thread 2: GIL is acquired. Calling Destroy without locking env. Note that Destroy locks env internally. -> deadlock')
    env.Destroy() # _mutexInit is locked first then env is locked in Destroy

def main():
    env = openravepy.Environment(0)

    t0 = threading.Thread(target=Thread1, args=[env])
    t1 = threading.Thread(target=Thread2, args=[env])
    t0.start()
    t1.start()
    t0.join()
    t1.join()

if __name__ == '__main__':
    main()
```

eisoku9618:
@yoshikikanemoto Could you check this MR? This MR is composed of two changes.

1. introduce a new env lock that can predict a potential deadlock and allow users to switch to it with a CMake option.
2. release GIL on `pyRaveDestroy()` for the new env lock not to raise an exception on `pyRaveDestroy()`.

pipelineid=745377

The default behavior change by merging this PR is only `point 2` (`pyRaveDestroy()`) because `point 1` (new mutex) is disabled by default.

CMakeLists.txt
```diff
@@ -71,7 +72,11 @@ set(PACKAGE_VERSION "0" CACHE STRING "the package-specific version used for uplo
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/modules-cmake")
 set(CPACK_DEBIAN_PACKAGE_NAME openrave${OPENRAVE_VERSION_MAJOR}.${OPENRAVE_VERSION_MINOR})
 
-if (OPT_ENVIRONMENT_RECURSIVE_LOCK)
+if (OPT_ENVIRONMENT_RECURSIVE_LOCK_WITH_GIL_CHECK)
+  message(STATUS "using a custom mutex that can detect a potential deadlock between GIL and environment mutex.")
+  set(OPENRAVE_ENVIRONMENT_RECURSIVE_LOCK 2)
```
yoshikikanemoto: `2` is bad at readability. Is there good way to replace a number with a named something?

src/libopenrave-core/CMakeLists.txt
```diff
@@ -77,16 +77,32 @@ endif()
 link_directories(${OPENRAVE_LINK_DIRS})
 
 add_library(libopenrave-core SHARED ${openrave_core_SOURCES})
-target_include_directories(libopenrave-core PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+if (OPT_ENVIRONMENT_RECURSIVE_LOCK_WITH_GIL_CHECK)
```
yoshikikanemoto, line 81: why not
```
target_include_directories(libopenrave-core PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
if(OPT_ENVIRONMENT_RECURSIVE_LOCK_WITH_GIL_CHECK)
    target_include_directories(libopenrave-core PUBLIC ${Python3_INCLUDE_DIRS})
endif()
```

src/libopenrave/libopenrave.cpp
```diff
@@ -15,6 +15,70 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "libopenrave.h"
+#include <pthread.h>
```
yoshikikanemoto: why outside `#if OPENRAVE_ENVIRONMENT_RECURSIVE_LOCK == 2`?

CMakeLists.txt
eisoku9618: @yoshikikanemoto I introduced `OPENRAVE_ENVIRONMENT_RECURSIVE_LOCK_WITH_GIL_CHECK` and changed to use that together with `OPT_ENVIRONMENT_RECURSIVE_LOCK`.

eisoku9618:
@yoshikikanemoto Thank you for reviewing. I just confirmed that the tests pass with the latest changes based on your review comments. Could you merge this PR?

cielavenir:
@yoshikikanemoto @rdiankov We decided to setup a dedicated autotester schedule to check the mutex. We do prohibit to use that build on the controller. Still, we would like to have the mutex option on master (production branch). cc @ntohge

eisoku9618:
@yoshikikanemoto This PR is not merged yet. Could you take a look at this?

yoshikikanemoto:
LTGM.  thanks. note that this change does not make behavioral changes with default build options (except releasing GIL on destroying env from python binding)

eisoku9618:
cc @wkentaro

PR #1512, opened 2025-05-15
improve CheckCollision(ray, link, report) in ODE plugin even when CO_RayAnyHit option is not set

MiraiHattori:
This PR fixes those issues below:
- ray collision checking is supported in ODE plugin, but the return value can be wrong depending on the CollisionOption set.
- there were no python bindings for `CheckCollision(ray, link)` or `CheckCollision(ray, link, report)` signature
- the order of `CheckCollision` function overloads is bad, `CheckCollision(object, object)` should be defined after `CheckCollision(PyKinBodyPtr, object)` or `CheckCollision(ray, link)` for example

I could not find the existing usage of `CheckCollision(ray, link, report=)` in our codebase, and I'm planning to use it for a new implementation.


yoshikikanemoto:
LGTM. please bump patch version and add changelog.

yoshikikanemoto:
thanks. in addition to changes in the description this PR changes to expose `CheckCollision(PyKinBodyPtr pbody, object bodyexcluded, object linkexcluded, PyCollisionReportPtr pyreport)` and `CheckCollision(PyKinBodyPtr pbody, object bodyexcluded, object linkexcluded)` that were previously hidden. i also double-checked they are correct.

PR #1515, opened 2025-05-25
Do not use purely random value since it causes flakiness

cielavenir:
Porting testopenrave!120 to openrave CI.


cielavenir:
@snozawa please check

rdiankov:
Not sure I agree with this one. If there is a random value that causes the test to fail, then we need to investigate.

PR #1517, opened 2025-06-09
Fix CheckCollision pybind definition order

MiraiHattori:
After the recent change in the link below about CheckCollision definition order, `CheckCollision(py::object, py::object, py::object)` unexpectedly came earlier than `CheckCollision(py::object, py::object, PyCollisionReportPtr)`.

https://github.com/rdiankov/openrave/pull/1512/files#diff-f650196f54d2c274a196b6859776476696754a99811a9388333cd1ef8e04b0e5R1068

To avoid further mistake as much as possible, in this PR I
- fixed the `.def` order of those three-args function
- fixed the order of the function pointers definitions to match the `.def` order


rdiankov:
@MiraiHattori to avoid further mistake, can you create a test in testopenrave to call all these different flavors of check collision?

MiraiHattori:
@rdiankov sure. I added such test in testopenrave!149 and updated controllercommon!1295 with it. I'm waiting for the pipeline to finish (edited: succeeded)

kanbouchou:
@rdiankov This MR looks good to me, there is newly added test (MR 149) for this which fails with production and passes with this branch. Please check and merge.

eisoku9618:
@kanbouchou cc @rdiankov It would be great if this PR could be merged this week.

kanbouchou:
> @kanbouchou cc @rdiankov It would be great if this PR could be merged this week.

@eisoku9618 I've already approved. @rdiankov please check and merge~

rdiankov:
thanks

Issue #1057, opened 2022-02-17
Exactly same robot may return different KinematicsGeometryHash but it should return the same hash


eisoku9618:
When we compute the hash, we convert floating point value to string with `std::fixed` and `std::setprecision(4)`, so the hash can be affected by a super small difference like this.

```c++
#include <iostream>
#include <iomanip>

int main () {
    std::ostringstream ss;
    ss << std::fixed << std::setprecision(4);
    double a = 0.001249999999;
    double b = 0.001250;
    ss << a << " " << b;
    std::cout << ss.str() << std::endl;
}
```

-> `0.0012 0.0013`

And this kind of small difference can happen based on the robot kinematics configuration like the following code.

```
env.Remove(robot)
env.Add(robot)
robot.GetKinematicsGeometryHash() # return hash A
env.Remove(robot)
env.Add(robot)
robot.GetKinematicsGeometryHash() # can return hash B, but robot is not changed at all so should return hash A
```

It is because 
1. `KinBody::_ComputeInternalInformation()` called in `RobotBase::_ComputeInternalInformation` does 
 like this `SetDOFValues(robot.GetDOFValues())` operation, and it changes each link's transform slightly from each original transform.
2.  In `_ComputeConnectedBodiesInformation` called just before `KinBody::_ComputeInternalInformation()`, `_tLeftNoOffset`, which is used for kinematic hash computation, is already computed based on the original transform.
3. So next time we call `RobotBase::_ComputeInternalInformation` even when we don't move robot dof value nor the base link transform, the kinematics hash for the connected body can be changed.

cc @kanbouchou @Puttichai @yoshikikanemoto @strmio @kanunikov-denys @rdiankov

cielavenir:
Thank you, it is even possible to be related to https://redmine.mujin.co.jp/issues/5268 . At that time I tried to introduce a workaround https://github.com/rdiankov/openrave/commit/b1c74335398b45425c7dd149a089ac6da6cb0047 , but if it can be fixed without this workaround it is great.

rdiankov:
I think the simplest way to solve the problem is to round to the nearest 4th or 5th decimal digit (controlled by `SERIALIZATION_PRECISION`). That should be the most robust since the numbers we input into these things usually don't have that many significant digits.

our current implementation is:
```
template<typename T>
inline T SerializationValue(T f)
{
[...]

PR #1518, opened 2025-06-11
revive the more consistent hash computation trick

eisoku9618:
At 63fc2a70595fd605203763fd12ab35fefb94ecda (Jan 8 04:13:05 2010), @rdiankov added a trick to make hash computation more consistent by truncating floating values around 0.

However, by the transition from `serialize` to `DigestHash` in https://github.com/rdiankov/openrave/pull/1504, the trick was dropped.

I don't think it is intentional, so I would like to reveive it in this PR. Without this trick, the probability to hit the hash computation issue around connected bodies explained in https://github.com/rdiankov/openrave/issues/1057 gets increased. And actually I encountered the issue again in another project.

I also added a unit test in our internal repository PR 1296 to ensure that just Remove/Add does not change the hash.


rschlaikjer:
Is there a reason to add this other codepath instead of changing then serialization precision? It seems like it is accomplishing the same thing?

```c++
int main(int argc, char**argv)
{
    const float value = std::stof(argv[1]);
    for (int precision = 1; precision < 6; precision++) {
        double roundingScale = pow(10, precision);
        const float result = round(value * roundingScale) / roundingScale;
        printf("precision %d: %.10f (%08x)\n", precision, result, *reinterpret_cast<const uint32_t*>(&result));
    }
}
```

```
$ g++ foo.cpp && for VALUE in 1 0.1 0.01 0.001 0.0001 0.00001; do echo $VALUE; ./a.out $VALUE; echo; done
1
precision 1: 1.0000000000 (3f800000)
precision 2: 1.0000000000 (3f800000)
precision 3: 1.0000000000 (3f800000)
precision 4: 1.0000000000 (3f800000)
precision 5: 1.0000000000 (3f800000)

0.1
precision 1: 0.1000000015 (3dcccccd)
precision 2: 0.1000000015 (3dcccccd)
precision 3: 0.1000000015 (3dcccccd)
precision 4: 0.1000000015 (3dcccccd)
precision 5: 0.1000000015 (3dcccccd)

0.01
precision 1: 0.0000000000 (00000000)
precision 2: 0.0099999998 (3c23d70a)
precision 3: 0.0099999998 (3c23d70a)
precision 4: 0.0099999998 (3c23d70a)
precision 5: 0.0099999998 (3c23d70a)

0.001
precision 1: 0.0000000000 (00000000)
precision 2: 0.0000000000 (00000000)
precision 3: 0.0010000000 (3a83126f)
precision 4: 0.0010000000 (3a83126f)
precision 5: 0.0010000000 (3a83126f)

0.0001
precision 1: 0.0000000000 (00000000)
precision 2: 0.0000000000 (00000000)
precision 3: 0.0000000000 (00000000)
precision 4: 0.0001000000 (38d1b717)
precision 5: 0.0001000000 (38d1b717)

0.00001
precision 1: 0.0000000000 (00000000)
precision 2: 0.0000000000 (00000000)
precision 3: 0.0000000000 (00000000)
precision 4: 0.0000000000 (00000000)
precision 5: 0.0000100000 (3727c5ac)

```

eisoku9618:
@rschlaikjer The difference that made the hash different was something like `-0.00001` vs `0`.

```
$ ./a.out -0.00001
precision 1: -0.0000000000 (80000000)
precision 2: -0.0000000000 (80000000)
precision 3: -0.0000000000 (80000000)
precision 4: -0.0000000000 (80000000)
precision 5: -0.0000100000 (b727c5ac)

$ ./a.out 0
precision 1: 0.0000000000 (00000000)
precision 2: 0.0000000000 (00000000)
precision 3: 0.0000000000 (00000000)
precision 4: 0.0000000000 (00000000)
precision 5: 0.0000000000 (00000000)
```

rschlaikjer:
Ugh, I forgot about negative zero... I love IEE754. Maybe the better fix is to handle -0 explicitly; that way everything is still keyed off the same `SERIALIZATION_PRECISION` macro instead of having a second set of magic numbers

```c++
FloatT result = round(value * _roundingScale) / _roundingScale;
if (result == -0) {
    result = 0;
}
```

eisoku9618:
@rschlaikjer That's better, updated!

rdiankov:
Clearly we could not track the regression for this, so can you guys make a test for this? Thanks

eisoku9618:
@rdiankov Of course, I already added it in PR 1296 in our internal test repository.

yoshikikanemoto:
confirmed with the regression test.

PR #1520, opened 2025-07-03
feat: use scenefile#bodyid to download single body from scene that has multiple bodies via http

yoshikikanemoto:
@shossjer


yoshikikanemoto:
- autotester : no problem found
- unit test added to testopenrave
- real cell test : worked as expected

rdiankov:
thanks~

PR #1521, opened 2025-07-10
Support vector<bool> when loading and saving json

CapObvios:
- openrave MR: https://github.com/rdiankov/openrave/pull/1521
- controllerclientcpp MR: https://github.com/mujin/controllerclientcpp/pull/189


include/openrave/openravejson.h
```diff
@@ -348,6 +348,20 @@ inline void LoadJsonValue(const rapidjson::Value& v, bool& t) {
     }
 }
 
+inline void LoadJsonValue(const rapidjson::Value& v, std::vector<bool>::reference t)
```
rdiankov, line 351: what is `std::vector<bool>::reference`? is that `std::vector<bool>&`?

cielavenir: According to https://en.cppreference.com/w/cpp/container/vector_bool.html , vector\<bool\>& is not applicable

PR #1522, opened 2025-07-25
Add format_as for geometry type

cielavenir:
@rschlaikjer 

This adds OpenRAVE::format_as(GeometryType) so that fmt9+ works.

Already tested in trixie autotester schedule.

### DIFF
--- a/include/openrave/kinbody.h
+++ b/include/openrave/kinbody.h
@@ -66,6 +66,10 @@ enum GeometryType : uint8_t
 
 OPENRAVE_API const char* GetGeometryTypeString(GeometryType geometryType);
 
+OPENRAVE_API inline std::string format_as(GeometryType geomtype){
+    return GetGeometryTypeString(geomtype);
+}
+
 enum DynamicsConstraintsType : int8_t
 {
     DC_Unknown = -1, ///< constraints type is not set.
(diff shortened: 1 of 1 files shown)


include/openrave/kinbody.h
```diff
@@ -66,6 +66,10 @@ enum GeometryType : uint8_t
 
 OPENRAVE_API const char* GetGeometryTypeString(GeometryType geometryType);
 
+OPENRAVE_API inline std::string format_as(GeometryType geomtype){
```
rdiankov: err... returning std::string is not good, it allocates memory.

cielavenir: @rdiankov changed to `const char*`. This has been tested on trixie pipeline

cielavenir: @rdiankov could you check again?

cielavenir:
@pico737 @rolandjitsu can we get this forward?

yoshikikanemoto:
openrave is not using `fmt`. so i am afraid doing this will not be sustainable (will keep adding similar changes and usage of our time would be unproductive).

cielavenir:
I moved this to mujinopenravecommon.h.

Closing.

Might reopen if feasible though.

PR #1524, opened 2025-08-01
Store all files with LF endings, per git best practices

rschlaikjer:
This PR converts DOS line ending files to UNIX line endings.

It is best viewed using the `?w=1` option to reveal that it makes no non-whitespace changes:
https://github.com/rdiankov/openrave/pull/1524/files?w=1

This standardizes line endings within the repo, and aligns to git best practices (checked in code is LF, windows users can use `autocrlf`)


rdiankov:
thanks~

PR #1528, opened 2025-08-28
Fix extraction of single body with transform in referenced body

rschlaikjer:
If a base environment has a body with a transform, and a second environment has a body referencing the base body that _doesn't_ have its own transform, and the body is loaded via ReadKinBodyUri, then the loaded kinbody will have an (incorrect) identity transform, instead of the (correct) transform from the referenced body.

This seems to be specific to the ExtractOne case in the json reader:
- The kinbody is created with InitFromKinBodyInfo, which doesn't apply the _transform field of the info. This is in contrast to Update*, which _does_ read the transform and apply it
- The _Extract codepath has an extra call to a _ExtractTransform helper, which default constructs a transform then applies it to the body even if there _wasn't_ a transform in the referencing body info.


rschlaikjer:
Internal discussion: https://tiny.mujin.co.jp/w2zdp

rdiankov:
I'm okay with this change if it passes all of our tests. Thanks!

yoshikikanemoto:
the tests passed. thanks!

PR #1529, opened 2025-09-01
Fix order of operations error

rschlaikjer:
Fixes overlooked test failure from https://github.com/rdiankov/openrave/pull/1528 

SetTransform needs to occur _after_ _baseLinkInBodyTransform is calculated, since otherwise that transform is affected by the body's transform in space.


yoshikikanemoto:
sorry for overlooking. thanks

rdiankov:
@rschlaikjer Can you add tests for this condition to catch it easier in the future?

rschlaikjer:
> @rschlaikjer Can you add tests for this condition to catch it easier in the future?

This was actually caught by existing tests, the tests were just still running when the PR was merged.

PR #1530, opened 2025-09-11
Guard list of plugins with mutex

shossjer:


cielavenir:
So not only adding but every iterating vPlugins needs mutex. LGTM.

yoshikikanemoto:
internal test pipeline was successful.

yoshikikanemoto:
thanks!

PR #1531, opened 2025-09-15
Add friction per kinbody's geometry, Enable friction by default in ODE plugin, Split ERP/CFM  of joint and contact in ODE plugin

yoshikikanemoto:
This MR is to enhance ODE plugin for wider usages.
contents of this MR:

- Introduce friction to kinbody geometry.
- Add support for using per-object frictions in ODE plugin.
- Enable friction model by default in ODE plugin.
- Split ERP and CFM to to the ones for joint constraints and the ones for contact constraints (`erp`/`cfm` and `contactErp`/`contactCfm` respectively)


plugins/oderave/odephysics.h
```diff
@@ -627,17 +671,34 @@ The possible properties that can be set are: ";
             }
         }
 
+        // get frictions before processing collisions.
+        float combinedFriction;
+        if (_usePerSurfaceFrictions) {
+            // we assume that a friction is a property of a link and hence every geom in a link has the same friction value.
```
felixvd:
```suggestion
            // we assume that friction is a property of the link and hence every geom in a link has the same friction value.
```
nit. But also, if we assume the link has uniform friction, why not store it with the link?

fedorrrd: On ODE side friction is defined for a surface, which is linked to geoms, not links (see official [documentation](https://ode.org/wiki/index.php/Manual#Contact_points:~:text=struct%20dContact%20%7B%0A%20%20%20%20%20dSurfaceParameters%20surface%3B%0A%20%20%20%20%20dContactGeom%20geom%3B%0A%20%20%20%20%20dVector3%20fdir1%3B%0A%7D%3B)). Hence, I believe it's better to store the information about friction on the geom level to comply with ODE notation, but use the assumption that a link is made of the same material, so that the friction for all its geoms is the same - this simplifies the code for extracting the friction from geoms.

yoshikikanemoto:
@fedorrrd do all necessary places have additional code to support  `_friction`? looks `GeometryInfo::Compare` and `Geometry::UpdateFromInfo` do not have it

plugins/oderave/odephysics.h
yoshikikanemoto: @fedorrrd 

> On ODE side friction is defined for a surface, which is linked to geoms, not links (see official documentation).

Really? I thought ODE can have different friction coefficient for each contact constraint. I think storing friction on geometry is a right choice. Just  the ode plugin in OpenRAVE does not associate ODE's geom to OpenRAVE's geom (instead, associated to OpenRAVE's link).

I think comment should be reworded to clarify this restriction.

yoshikikanemoto:
@fedorrrd can we initialize `_friction` with a negative value (`-1.0` for example) and use `_globalfriction` if `_friction` is negative? so that we can remove `_usePerSurfaceFrictions`?

plugins/oderave/odephysics.h
fedorrrd: @yoshikikanemoto @felixvd does this look better?

```
// Friction is a property of a contact surface in ODE which is linked to ODE geoms.
// Hence _friction is stored on the geom level in openRAVE.
// We assume that OpenRAVE link has a uniform material, therefore every geom in a link has the same _friction value.
// Hence we can just take the friction of the first geom GetGeometry(0).
```

yoshikikanemoto: `We assume that OpenRAVE link has a uniform material` This is misleading. To me, it is just a limitation of the current implementation. 

maybe we should just put a TODO.
```
TODO: Being able to get corresponding OpenRAVE's geometry from ODE's dGeomID so that we can use friction coefficient from the actual geometries that have the collision.
```

yoshikikanemoto:
@fedorrrd can we remove `_usePerSurfaceFrictions`?

yoshikikanemoto:
test pipeline is all green and manual test is done for the relevant features that are not covered by automated test.

felixvd:
:partying_face: Thanks @yoshikikanemoto @fedorrrd

PR #1532, opened 2025-09-24
Fix potential bad memory access at polyroots function

MiraiHattori:
image

This PR fixes the attached compiler warning. Potentially if the input of those functions are per-joint linearly changing trajectories, then there may be segmentation fault.

## Details
- when the template parameter `D`=1, then this `roots[1]` is out of bound.
https://github.com/rdiankov/openrave/blob/7fdb0b5b8c42df46e3e16057027f963c9e01d90b/include/openrave/mathextra.h#L1233
- caller with `D`=1. if the system accidentally hits it, segmentation fault will happen
https://github.com/rdiankov/openrave/blob/7fdb0b5b8c42df46e3e16057027f963c9e01d90b/include/openrave/mathextra.h#L1467-L1470


include/openrave/mathextra.h
```diff
@@ -1466,7 +1468,9 @@ inline bool computequinticnextdiscretizedstep(const T* coeffs, const T step, con
         polyroots<T, 2>(&tempcoeffs[3], rawroots, numroots);
     }
     else if( tempcoeffs[4] != 0 ) {
-        polyroots<T, 1>(&tempcoeffs[4], rawroots, numroots);
+        // polyroots<T, 1>(&tempcoeffs[4], rawroots, numroots);  // out-of-bound memory access
+        numroots = 1;
+        rawroots[0] = -tempcoeffs[4]/tempcoeffs[3];
```
Puttichai: @MiraiHattori Please fix this. It should be `-tempcoeffs[5]/tempcoeffs[4]`.

MiraiHattori: that's true, fixed.

MiraiHattori:
I added `-Werror=array-bounds` compiler options -> later reverted

MiraiHattori:
bumped "minor" version (instead of patch version) / added changelog.

MiraiHattori:
> I added `-Werror=array-bounds` compiler options

compilation failed, taking a look.

cielavenir:
Could be https://qiita.com/cielavenir/items/3e52823a8193b9fb6989

MiraiHattori:
> Could be https://qiita.com/cielavenir/items/3e52823a8193b9fb6989

With an older boost version, compiler did not catch the false positive out of bounds.

I think it's okay without `-Werror=array-bounds` in this PR, so reverted.

MiraiHattori:
@yoshikikanemoto could you review this MR?

yoshikikanemoto:
@MiraiHattori why is minor version bump? looks like there are no ABI breaking changes? do you have unit tests that produce out-of-bounds access?

MiraiHattori:
> @MiraiHattori why is minor version bump? looks like there are no ABI breaking changes? do you have unit tests that produce out-of-bounds access?

I reverted to patch version bump since there was no interface changes.

I don't have a unit test to reproduce the issue so far, let me make one.

PR #1533, opened 2025-09-30
Set --Wl,--no-undefined

cielavenir:
cc @rolandjitsu (Why he does not have the access to this repo yet?)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,6 +190,8 @@ endif()
 if( CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX OR COMPILER_IS_CLANG)
   add_definitions("-fno-strict-aliasing -Wall -Werror=shadow")
   add_definitions("-Werror=return-type")
+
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -813,13 +815,13 @@ if (OPT_PYTHON3)
   find_package(pybind11 2.9.2) # 2.9.2 is the last version of pybind11 to support Python 2.
 
   if (pybind11_FOUND)
-    target_link_libraries(openravepy3_meta INTERFACE rt Python3::Module Python3::NumPy pybind11::module libopenrave libopenrave-core)
+    target_link_libraries(openravepy3_meta INTERFACE rt Python3::Python Python3::NumPy pybind11::module libopenrave libopenrave-core)
```
cielavenir, line 818: When you see controllercommon CMakeLists, you can see LiBoon had difficulty here.

The thing is, to apply -Wl,--no-undefined, we need to use Python3::Python.

Python3::Module causes:

1. On Windows, link to python3.dll
2. On Unix, allow undefined

Of course, if we want to use stable ABI (for public binary wheel), we need to use `Python3::SABIModule` and give up -Wl,--no-undefined, but anyway stable ABI is not supported by pybind11 (as for 3.0).

PR #1537, opened 2025-10-30
Prevent user from generating NaN view matrix

rschlaikjer:
If a NaN value ends up in OSG, it will repeatedly spam the following message to stderr, completely swamping the logs of the parent process:
```
CullVisitor::apply(Geode&) detected NaN,
    depth=-nan, center=(0 0 0),
    matrix={
	nan nan nan nan 
	nan nan nan nan 
	nan nan nan nan 
	nan nan nan nan 
}
```

This seems to be caused by updating the view on scenes with nothing in them, which results in a +-FLT_MAX bounding box, which causes overflows later. Add a check for NaN matrices before we write back to OSG, and abandon the view update if it would cause a problem

Sample output w/ patch:
``` 
2025-10-30 11:15:26,080 openrave [WARN] [osgviewerwidget.cpp:1277 _MoveCameraPointOfView] scene has bounds [340282346638528859811704183484516925440.000000, 340282346638528859811704183484516925440.000000, 340282346638528859811704183484516925440.000000] to [-340282346638528859811704183484516925440.000000, -340282346638528859811704183484516925440.000000, -340282346638528859811704183484516925440.000000] for a size of 884079469945044887409358651045961531392.000000, camera distance 1067178323278472196836424400885707177984.000000, resulting in new camera position of [0.000000, 0.000000, inf] which produces an invalid view matrix; refusing to set new view matrix
```

### DIFF
--- a/plugins/qtosgrave/osgviewerwidget.cpp
+++ b/plugins/qtosgrave/osgviewerwidget.cpp
@@ -1263,6 +1263,20 @@ void QOSGViewerWidget::_MoveCameraPointOfView(const osg::Vec3d& axis)
     osg::Matrixd newViewMatrix;
     osg::Vec3d newCameraPos = boundingBox.center() + axis * cameraDistance;
     newViewMatrix.makeLookAt(newCameraPos, newCameraPos - axis, worldUpVector);
+
+    // If the scene was _empty_, then our bounding box will end up being +-FLT_MAX in all directions.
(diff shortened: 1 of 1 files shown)


cielavenir (approved):

yoshikikanemoto (approved):

yoshikikanemoto:
thanks

PR #1541, opened 2025-11-19
Require assimp 5.0

cielavenir:
to avoid the collision with vendored assimp.


cielavenir:
I decided to build assimp:uninstall in jhbuild tree instead.

PR #1542, opened 2025-11-25
reduce preallocation

yoshikikanemoto:
to reduce memory footprint,

- delay allocation of buffer region for JSON loading from the constructor to timing right before usage. saves 4MB per env
- delay allocation of `vVisitedDiscretization` in `parabolicsmoother2` and reduce amount of pre-allocation 16MB -> 1MB as typically `nEndTimeDiscretization` is less than 1000.


Puttichai:
I think you might want to update this check in `_Shortcut` function too.
```c++
                    if( nEndTimeDiscretization <= 0x1000 ) {
                        vVisitedDiscretization.resize(nEndTimeDiscretization*nEndTimeDiscretization,0);
                    }
```

yoshikikanemoto:
> I think you might want to update this check in `_Shortcut` function too.
> 
> ```c++
>                     if( nEndTimeDiscretization <= 0x1000 ) {
>                         vVisitedDiscretization.resize(nEndTimeDiscretization*nEndTimeDiscretization,0);
>                     }
> ```

I would like to keep it for now as changing this may affect computation speed a lot when nEndTimeDiscretization is large.

Puttichai (approved):
LGTM

PR #1544, opened 2025-12-27
py::extract before releasing gil

cielavenir:
py::extract after releasing gil is illegal and it must be done before releasing gil.

@yoshikikanemoto @rolandjitsu @pico737

### DIFF
--- a/python/bindings/openravepy_planningutils.cpp
+++ b/python/bindings/openravepy_planningutils.cpp
@@ -93,11 +93,12 @@ object pyGetReverseTrajectory(PyTrajectoryBasePtr pytraj)
 void pyVerifyTrajectory(object pyparameters, PyTrajectoryBasePtr pytraj, dReal samplingstep, bool releasegil=true)
 {
     PlannerBase::PlannerParametersConstPtr parameters = openravepy::GetPlannerParametersConst(pyparameters);
+    TrajectoryBasePtr traj = openravepy::GetTrajectory(pytraj);
     openravepy::PythonThreadSaverPtr statesaver;
     if( releasegil ) {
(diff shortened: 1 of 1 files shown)


python/bindings/openravepy_planningutils.cpp
```diff
@@ -92,11 +92,12 @@ object pyGetReverseTrajectory(PyTrajectoryBasePtr pytraj)
 
 void pyVerifyTrajectory(object pyparameters, PyTrajectoryBasePtr pytraj, dReal samplingstep, bool releasegil=true)
 {
+    PlannerBase::PlannerParametersConstPtr parameters = openravepy::GetPlannerParametersConst(pyparameters);
     openravepy::PythonThreadSaverPtr statesaver;
     if( releasegil ) {
         statesaver.reset(new openravepy::PythonThreadSaver());
     }
-    OpenRAVE::planningutils::VerifyTrajectory(openravepy::GetPlannerParametersConst(pyparameters), openravepy::GetTrajectory(pytraj),samplingstep);
+    OpenRAVE::planningutils::VerifyTrajectory(parameters, openravepy::GetTrajectory(pytraj),samplingstep);
```
rdiankov: @cielavenir don't you also have the same problem with `openravepy::GetTrajectory`?

rdiankov:
Thanks

python/bindings/openravepy_planningutils.cpp
cielavenir: the background is, currently GetTrajectory just returns member variable and does not call py::extract, so it is not required but I accept your suggestion to avoid further surprise after changing GetTrajectory implementation.

cc @yoshikikanemoto @rolandjitsu

PR #1546, opened 2026-02-02
WIP: Do deep copy of geomety info

kanbouchou:


rdiankov:
This will slow things down, so not sure I can agree to this...
The caller should know that the geometry ptr will be used as is, so why can't this be solved on the caller side?

kanbouchou:
> This will slow things down, so not sure I can agree to this... The caller should know that the geometry ptr will be used as is, so why can't this be solved on the caller side?

@rdiankov yes, alternatively we can ask the caller side to make deep copy. But what do you think about `AddGeometry`'s `addToGroups` option being true? I don't come up with a use case that someone wants to share same shared pointer among different extra geometry groups and `_vgeometryinfos`.

PR #1549, opened 2026-02-24
change to minor version bump

yoshikikanemoto:
https://github.com/rdiankov/openrave/pull/1548 breaks ABI so needed to be minor version bump


cielavenir:
@yoshikikanemoto I uploaded the abitest result to https://tiny.mujin.co.jp/vdfmw

PR #1550, opened 2026-02-27
Add uri to KinBodyInfo::SerializeJSON output

eisoku9618:
When sending `KinBody` to another process, we serialize it on the sender side as `KinBodyInfo` and deserialize on the receiver side. However, `KinBodyInfo::SerializeJSON` supports `_referenceUri` but not `_uri` intentionally, so we have been working around this by copying `_uri` into `_referenceUri` before serialization in application code. The problem is that this overwrites the original `_referenceUri`, so the original value is lost. To avoid that, I would like `KinBodyInfo::SerializeJSON` to support `_uri` as well.

Here is an example of the current application code for the sender and receiver sides.

```c++
// sender side
void SerializeBodyInfo(const KinBody& body, rapidjson::Value& rKinBodyInfo)
{
    KinBody::KinBodyInfo kinfo;
    body.ExtractInfo(kinfo, EIO_Everything);

    // this swap is the current workaround, and I would like to remove this line.
    kinfo._uri.swap(kinfo._referenceUri);

    kinfo.SerializeJSON(rKinBodyInfo);
}

// receiver side
void ApplyEnvBodiesCacheState(const rapidjson::Value& rKinBodyInfo)
{
    KinBodyPtr pbody;

    // this is the current workaround, and I would like to change "referenceUri" to "uri"
    std::string referenceUri = mujinjson::GetStringJsonValueByKey(rKinBodyInfo, "referenceUri");
    bool bIsDummy = referenceUri.find("__dummyuri__") != std::string::npos;
    if( !bIsDummy ) {
        pbody = env.ReadKinBodyURI(KinBodyPtr(), referenceUri);
    }

    if( !pbody ) {
        KinBody::KinBodyInfo kinfo;
        kinfo.DeserializeJSON(rKinBodyInfo);
        // currently this pbody does not have the original "referenceUri",
        // and I would like to change to have it.
        pbody->InitFromKinBodyInfo(kinfo);
    }
}
```


yoshikikanemoto (approved):
Makes sense.

@rdiankov @ziyan 

Do you see any problems? Further description about motivation can be found https://tiny.mujin.co.jp/lz998
Our tests are all green with this change + other changes in internal repos.

yoshikikanemoto:
saving uri into serialized data requires a lot of design changes and user flow changes when dealing with scene rename etc so closing

PR #1552, opened 2026-04-09
Fix GetReadableInterface binding

cielavenir:


cielavenir:
since py::typing::Optional is expanded to py::object, there should not be ABI change though

cielavenir:
pipeline: https://tiny.mujin.co.jp/vg52m

PR #1553, opened 2026-04-20
Fix GetName() for OSG viewr

ReinaKousaka:
For `qtosg` viewer, its implementation of `GetName()` is now missing. This MR fixes it and makes this behavior almost identical as `qtcoin` viewer 

#### `production` branch
```

In [1]: env.GetViewer().GetName()
---------------------------------------------------------------------------
_OpenRAVEException                        Traceback (most recent call last)
Cell In[1], line 1
----> 1 env.GetViewer().GetName()

_OpenRAVEException: ('openrave (NotImplemented): [virtual const std::string& OpenRAVE::ViewerBase::GetName() const:186] not implemented', 'NotImplemented')
```


rdiankov:
@cielavenir can you take a look?

cielavenir:
ok, tested env.GetViewer().GetName() and env.GetViewer().SetName()

thank you for the patch~~

cielavenir (approved):

ReinaKousaka:
@cielavenir @rdiankov A little more background here: I tried using videoRecorder to save a batch of planning trajectory animations executed by a script, but neither the OSG nor the Coin viewer supports it. And I feel like fixing the videoRecord codebase would require large-scale changes, not just a few lines of modification.

cielavenir:
@ReinaKousaka it is known that offscreen rendering is required, which I tested years ago but not on Debian 13 yet.

I guess using Xvfb and `ffmpeg -f x11grab` combo is better.

But still, title screen should be changeable (this pull request).

cielavenir:
(test_kinematics.TestKinematics.test_bodybasic had flakiness)

PR #1554, opened 2026-05-21
fix: make jsonreader to download the top scene

gcccai:
```cpp
    JSONReader(const AttributesList& atts, EnvironmentBasePtr penv, const std::string& defaultSuffix, EnvironmentLoadContextJSON& loadContext)
```
When the attributes has `"mustresolveenvuri" = "true"`, the JSONReader seems to download only the referencedURIs but not the main scene itself. This MR tries to fix this problem.


woswos (approved):

rdiankov:
Have you tested for regressions on our internal side?

gcccai:
> Have you tested for regressions on our internal side?

let me run the pipeline again since it is 1 month old

gcccai:
> Have you tested for regressions on our internal side?

the pipeline did pass, see robotbridge MR 3403

woswos:
@rdiankov @yoshikikanemoto Do you have further comments on this pull request?

PR #1556, opened 2026-06-05
Allow registering default json reader

rschlaikjer:
In addition to registering readers for specific readables, allow registering a default fallback reader in the event that no specific reader exists.

This allows callers to implement their own preferred fallback behaviour, instead of relying on the default json readable (overwrite) behaviour.

Additional context: https://tiny.mujin.co.jp/c15b1


yoshikikanemoto:
LGTM. @rschlaikjer can you bump minor version and add changelog entry?

rschlaikjer:
> LGTM. @rschlaikjer can you bump minor version and add changelog entry?

Added

yoshikikanemoto (approved):

PR #1557, opened 2026-06-09
perf(kinbody): store link transform directly in SetDOFValues loop

Puttichai:
`KinBody::SetDOFValues` stores each child link's transform with `childlink->SetTransform(t)` inside the forward-kinematics loop. `Link::SetTransform` locks the parent weak_ptr (`KinBodyPtr(_parent)`, an atomic refcount bump) and increments `_nUpdateStampId` once per link. The stamp is already bumped once per call by `_PostprocessChangedParameters`, so the per-link work is redundant.

This writes the transform directly (`childlink->_info._t = t`), removing one weak_ptr lock + stamp bump per link.

- Bit-for-bit identical results
- `SetDOFValues` ~2–30% faster, scaling with link count.

Unit test and report: https://tiny.mujin.co.jp/2ckuc


rdiankov:
@Puttichai does `_nUpdateStampId`  still get updated and where?

Puttichai:
> @Puttichai does `_nUpdateStampId` still get updated and where?

@rdiankov Yes. `_nUpdateStampId` gets updated in the end of `SetDOFValues` ([`kinbody.cpp:2405`](https://github.com/rdiankov/openrave/blob/optimize_setdofvalues_bookkeeping_2_20260609/src/libopenrave/kinbody.cpp#L2405) and [`kinbody.cpp:2652`](https://github.com/rdiankov/openrave/blob/optimize_setdofvalues_bookkeeping_2_20260609/src/libopenrave/kinbody.cpp#L2652)) via `KinBody::_PostprocessChangedParameters`.

In my understanding, exactly how many times `_nUpdateStampId` is bumped in one `SetDOFValues` operation does not matter as long as it is bumped.

rdiankov:
got it, thanks~

PR #1558, opened 2026-06-11
Use an unordered / case sensitive map for readable interfaces

rschlaikjer:
This is a 3-6x performance improvement, and accesses to this map tend to happen in hot loops.


rdiankov:
I'm okay with this, thanks! Once our internal pipeline passes, feel free to merge.

rschlaikjer:
There was one instance of a case typo, after fixing that everything seems to pass: https://tiny.mujin.co.jp/c920i

yoshikikanemoto:
thanks!

PR #1559, opened 2026-06-11
Allow mapping functions over bodies in the env

rschlaikjer:
Allow mapping a function over the set of bodies in the env directly, instead of hackily through `GetBodiesMatchingFilter` + const cast

Context - https://tiny.mujin.co.jp/dby29


yoshikikanemoto:
LGTM. thanks

PR #1560, opened 2026-06-11
Keep cache of bodies with given readable interfaces

rschlaikjer:
There are a lot of places in our codebase where we want to find all bodies with a given readable - MRCS and mapfcommon are always searching for robots / payloads, any path through robot bridges that hits NotifyEnvironmentChanged has to search the env for location readables, etc. 

All of these call sites use linear scan over the entire env + GetReadable. This
- Is linear on the size of the env, even if the number of bodies that will match is constant and small
- Creates a huge number of wasted calls to GetReadableInterface, which costs a lock / find / unlock on every body.

Optimize this by maintaining a cache, updated when readables change, of readable ID -> bodies that currently have that readable ID. All of the call sites above go from being O(N) to O(1), and the additional cost under normal operation is fairly low - readable presence / absence on a body changes infrequently.

Tests: https://tiny.mujin.co.jp/z7viy


yoshikikanemoto:
@rdiankov do you agree with this direction to get bodies that have a particular readable interface with O(1)?

include/openrave/environment.h
```diff
@@ -602,6 +602,17 @@ class OPENRAVE_API EnvironmentBase : public boost::enable_shared_from_this<Envir
     /// \throw openrave_exception with ORE_Timeout error code
     virtual void GetBodiesMatchingFilter(std::vector<KinBodyPtr>& bodies, const std::function<bool(const KinBody&)>& filterFunction, uint64_t timeout = 0) const = 0;
 
+    /// \brief Get all bodies added to the environment that currently have a non-null readable interface with the given id. <b>[multi-thread safe]</b>
+    ///
+    /// Uses an internal cache that is kept in sync as bodies are added, removed, and have their readable interfaces modified.
+    /// Therefore this does not scale with the total number of bodies in the environment, unlike scanning every body with GetBodies.
+    /// A separate **interface mutex** is locked for reading the bodies.
+    /// \param[out] bodies filled with the matching bodies, sorted by environment body index in ascending order
```
yoshikikanemoto: implementation just iterates through `unordered_set` so not sorted?

include/openrave/kinbody.h
```diff
@@ -3654,6 +3654,19 @@ class OPENRAVE_API KinBody : public InterfaceBase
...
     }
 
+    /// \name Readable interface overrides
+    ///
+    /// These override the ReadablesContainer methods so that, while this body is added to an environment,  any change to which readable interfaces are non-null is reported to the environment.
+    /// This allows the environment to maintain a cache of which bodies have which readable IDs, to allow faster lookup.
+    /// The base readable container behaviour is unchanged.
+    //@{
+    ReadablePtr SetReadableInterface(const std::string& id, ReadablePtr readable) override;
+    void SetReadableInterfaces(const READERSMAP& mapReadables, bool bClearAllExisting) override;
```
yoshikikanemoto, line 3669: doc of `ReadablesContainer::SetReadableInterfaces` says the function is multi-thread-safe but overridden one is not thread safe anymore. it changes the contract. aren't there any problems with the existing code base? doc should also be updated?

src/libopenrave-core/environment-core.h
```diff
@@ -4566,6 +4667,11 @@ class Environment : public EnvironmentBase
 
     std::set<int> _environmentIndexRecyclePool; ///< body indices which can be reused later, because kin bodies who had these id's previously are already removed from the environment. This is to prevent env id's from growing without bound when kin bodies are removed and added repeatedly. protected by _mutexInterfaces
 
+    /// Map of readable interface ID -> set of body indices that are currently tagged with that readable interface.
+    /// Used by GetBodiesWithReadableInterface for O(1) lookup instead of O(N) on bodies in the environment.
+    /// Protected by _mutexInterfaces
+    std::unordered_map<std::string, std::unordered_set<int>> kinBodyEnvironmentIdByReadableInterfaceId;
```
yoshikikanemoto: variable name is not consistent with existing members in this class (`_`  at front)

yoshikikanemoto:
`xmlreaders-core.cpp` modifies `__mapReadableInterfaces` directly. this might cause problems.

include/openrave/kinbody.h
rschlaikjer: Hmm, good catch - the readable interface update itself is safe (body `_mutexInterface`), and the cache update itself is safe (env `_mutexInterfaces`), but there's no strong ordering between ReadableA + EnvA and ReadableB + EnvB that would prevent 
- Set ReadableA
- Set ReadableB
- Update env cache ReadableB
- Update env cache ReadableA

resulting in stale cache. I don't want to lock env `_mutexInterfaces` more than necessary, so maybe we set up the critical sections such that instead of generating the diff in the body critical section, we detect _change_ in the body critical section and then do the actual diff in the env section. That way,

- Set ReadableA -> flags change
- Set ReadableB -> flags change
- Update env cache ReadableB -> reads body, sees readableB, updates cache
- Update env cache ReadableA -> reads body, sees readableB still, correct results

and we can still do optimizations like not locking the env mutex if the body didn't change and not taking both locks at the same time.

rschlaikjer:
> `xmlreaders-core.cpp` modifies `__mapReadableInterfaces` directly. this might cause problems.

Hmm, it seems that this file completely ignores the lock too... is this function expected to be called on bodies that are already added to the env? If not, we can just add an assert here that they aren't added. Modifying the readables while added with no locking seems unsafe to begin with...

include/openrave/kinbody.h
rschlaikjer: Hmm, I guess this fixes this ordering:
```
Thread 1: Set ReadableA -> Body has readable
Thread 2: Del ReadableA -> Body readable deleted
Thread 2: Update env cache ReadableA -> Deleted
Thread 1: Update env cache ReadableA -> Added (bad!)
```

becomes:
```
Thread 1: Set ReadableA -> Body has readable
Thread 2: Del ReadableA -> Body readable deleted
Thread 2: Update env cache ReadableA -> Re-reads body -> Deleted
Thread 1: Update env cache ReadableA -> Re-reads body -> Still deleted (good)
```

but it doesn't fix third party sync:

```
Thread 1: Call set ReadableA, locks body interfaces, adds readable, releases body interfaces
Thread 2: Queries `ReadableA` by index, returns no match for this body (eventually consistent!)
Thread 1: Locks env, adds body to index
```

Strong consistency means we need to lock the env mutex for readable **write** paths; my hunch is that this probably isn't terrible. The intended locking model is also a bit vague though... if you **aren't holding the env lock**, which operations are supposed to have ordering guarantees?

yoshikikanemoto:
> > `xmlreaders-core.cpp` modifies `__mapReadableInterfaces` directly. this might cause problems.
> 
> Hmm, it seems that this file completely ignores the lock too... is this function expected to be called on bodies that are already added to the env? If not, we can just add an assert here that they aren't added. Modifying the readables while added with no locking seems unsafe to begin with...

> already added to the env?

I don't think so. But I'm not sure if it is strongly guaranteed.
Perhaps, you can just call `SetReadableInterface()` instead for either case assuming that the env lock is properly handled outside.

src/libopenrave/kinbody.cpp
```diff
@@ -934,6 +935,123 @@ void KinBody::SetId(const std::string& newid)
...
+{
+    // If we aren't added to the env, do nothing. We don't exist.
+    const int envBodyIndex = GetEnvironmentBodyIndex();
+    if (envBodyIndex <= 0) {
+        return;
+    }
+
+    // Work out which readables were added/removed
+    std::vector<std::string> affectedIds;
+    std::set_difference(idsAfter.begin(), idsAfter.end(), idsBefore.begin(), idsBefore.end(), std::back_inserter(affectedIds));
```
yoshikikanemoto: `std::set_difference` does not notice removal.

yoshikikanemoto: `set_symmetric_difference` may fit here.

src/libopenrave-core/environment-core.h
```diff
@@ -3506,7 +3533,85 @@ class Environment : public EnvironmentBase
...
+        // Lock the env first, then the body readables
+        ExclusiveLock lock(_mutexInterfaces);
+        OpenRAVE::KinBodyPtr pBody = _vecbodies.at(envBodyIndex);
+        boost::unique_lock<boost::shared_mutex> lockReadables(pBody->GetReadableInterfaceMutex());
+
+        // Re-check the set of readables on the body, removing it from any invalid indexes and adding it to any valid indexes
+        const ReadablesContainer::READERSMAP& bodyReadables = pBody->GetReadableInterfaces();
+        for (const std::string& readableId : updatedReadableInterfaceIds) {
+            // Readable doesn't exist, drop from the index
+            if (bodyReadables.find(readableId) == bodyReadables.end()) {
```
yoshikikanemoto: `_RegisterAddedBodyReadableInterfaces` excludes null. shouldn;t this function also skip null?

src/libopenrave/kinbody.cpp
rschlaikjer: Good catch, fixed

src/libopenrave-core/environment-core.h
rschlaikjer: Added to be safe

include/openrave/kinbody.h
yoshikikanemoto: the third party sync scenario does not look terrible to me either.

src/libopenrave-core/environment-core.h
```diff
@@ -3506,7 +3533,105 @@ class Environment : public EnvironmentBase
...
+                    }
+                }
+            }
+            return;
+        }
+
+        // Lock the env first, then the body readables
+        ExclusiveLock lock(_mutexInterfaces);
+        OpenRAVE::KinBodyPtr pBody = _vecbodies.at(envBodyIndex);
+        boost::unique_lock<boost::shared_mutex> lockReadables(pBody->GetReadableInterfaceMutex());
```
yoshikikanemoto: looking at other places, pBody might be null.

yoshikikanemoto:
other places look good to me

yoshikikanemoto:
> @rdiankov do you agree with this direction to get bodies that have a particular readable interface with O(1)?

@rdiankov ping~

src/libopenrave-core/environment-core.h
rschlaikjer: Under normal use should be impossible (this function is only called when the body is added to the env, so by definition it is in `_vecbodies`), but I guess someone could theoretically call `NotifyKinBodyReadableInterfacesChanged` directly with bad data? I can add a guard just to be safe.

rdiankov:
Guys, I like problem you have identified and I think it is great to try to optimize our system for this.
From what I understand, the issue we're trying to solve is:

- Allow a user to iterate over all the bodies that hold a specific readable interface without going through all the bodies in the env.

But managing a separate map `_kinBodyEnvironmentIdByReadableInterfaceId` for every readable interface that now has to be updated everytime there is a change via `NotifyKinBodyReadableInterfacesChanged` is a lot of overhead, and that function looks heavy. So we might be killing a lot of the potential gains. And this is essentially punishing every system out there, even if they don't use `GetBodiesWithReadableInterface`.

Here's my observations:

1. Most, if not all of the time, that a specific readable interface is set on a KinBody, it is not removed. Ie: a mobile robot never turns into a pickable target, etc.

2. Ultimately this feature is about reducing the number of bodies to iterate when looking for a specific ReadableInterface. We don't want to miss a body when doing this iteration, but it is okay to iterate through a bigger list that might not have that ReadableInterface.

3. We have many many readable interfaces, but very few of them will be used for `GetBodiesWithReadableInterface`


If we use these observations, then we can optimize the computation and reduce locking even further:

1. `_kinBodyEnvironmentIdByReadableInterfaceId` only stores IDs that have been queried with `GetBodiesWithReadableInterface`.
2. Change definition of `_kinBodyEnvironmentIdByReadableInterfaceId` to hold any kinbody that has ever held a readable interface of the particular id. It is okay if it is not currently holding it.
3. Rename `NotifyKinBodyReadableInterfacesChanged` to `NotifyKinBodyReadableInterfacesAdded` and only trigger when a new unique readable interface id is added.
4. `NotifyKinBodyReadableInterfacesAdded` then only has to iterate through all the keys in `_kinBodyEnvironmentIdByReadableInterfaceId` (which should be very few), and check if the new notified body holds that interface and then add it to `_kinBodyEnvironmentIdByReadableInterfaceId`. If `_kinBodyEnvironmentIdByReadableInterfaceId` is empty or current readable interfaces are empty, completely skip locking any mutex in the environment.


When a user first calls `GetBodiesWithReadableInterface`, `_kinBodyEnvironmentIdByReadableInterfaceId` will be empty and `_kinBodyEnvironmentIdByReadableInterfaceId` will get initialized with its first key then and then it will go through all the existing bodies once. Afterwards, `_kinBodyEnvironmentIdByReadableInterfaceId` will be tracked in `NotifyKinBodyReadableInterfacesAdded`.
This has the added benefit of not slowing down the system if `GetBodiesWithReadableInterface` is never called.

rschlaikjer:
Lazily enabling the feature seems fine to me, but being imprecise with removals means that `GetBodiesWithReadableInterface` now has to doublecheck the readable presence. If the set of bodies with the readable is _also_ large, this undoes the performance benefit to some extent. 

How about
- If nobody ever calls `GetBodiesWithReadableInterface`, we track nothing
- Once `GetBodiesWithReadableInterface` is called, additions and deletions are tracked as before

As you say, _removal_ of interfaces is rare. So, hooking the rare delete path to optimize the hot query path seems to me like it makes a lot of sense.

> tracking overhead

when profiling, the overhead of maintaining this dict was undetectable.

rdiankov:
Returning bodies without that readable is **not** as big of a problem as you are suggesting.
From all of the use cases in our system, most likely the caller will be also calling GetReadableInterface to get a pointer to that readable interface so they can work with the data, and they can check for a valid pointer before doing their task. So there's no computation time lost there.
If you see a different use case, can you point where that is in the system?

Tracking everything precisely is exactly what we should be avoiding because that stresses the rest of the system for no reason.

Thanks

rdiankov:
Once there's a real use case of returning the exact bodies, then let's consider the path of tracking removal of interfaces. Until then, no need for it.

rschlaikjer:
The issue with not tracking deletions is that any time a body index gets 'dirtied' it is irrevocably so. Consider an environment where `MobileRobotPayload` bodies and non-payload bodies are repeatedly created and destroyed - any index that ever held a payload is now in a state of maybe-payload, and you eventually trend back to linear search. It is important for a long lived system to _maintain_ performance. As mentioned before, the measured impact of tracking deletions was negligible because, as we both agree, it doesn't even happen that often. Loosening the contract creates more work in the hot path (query case) for a dubious improvement in the cold path (delete).

As mentioned before, if you want a way to just opt out of this feature entirely, I think that's fine - either implicit on first call or just a `SetReadableIndexingEnabled` function on the env - but the 'weak' version does not seem like an improvement. If you have a specific use case you are worried about we can look into it.

rdiankov:
I agree we need to maintain performance, but not sure your example is valid.

Your example assumes that `_kinBodyEnvironmentIdByReadableInterfaceId` won't get cleaned up when a body is removed from the environment and therefore its environmentBodyIndex would be preserved in this set, which is not what I was suggesting.

Yes, when a kinbody is removed from the environment, `_kinBodyEnvironmentIdByReadableInterfaceId` should also be cleaned up. The environment mutex is already locked, so we don't lose anything by cleaning it up.

So I don't see the extreme case you are talking about here where multiple creations and deletions would somehow make GetBodiesWithReadableInterface return all the bodies, with most of them not holding the desired readable interface.

Locking a mutex every time a a readable interface is deleted is wasteful, and not necessary.
Given that this library is used everywhere, let's be very careful before adding expensive callbacks like this.

Thanks

rschlaikjer:
I see, I interpreted your comment as ignoring _all_ removals. I think that works, though still think it makes for an uglier API - `GetBodiesThatProbablyHaveReadableInterface` is a bit more confusing to consumers, and again, I think the difference in performance is not material. If this is required to get this merged though, then sure.

rdiankov:
Great, thanks.
It depends on how great the implementation is.
Please clean this PR up with the assumptions above and try to reduce locking as much as possible, then I can review and check if that is the tightest we can get it. If even then, it looks like we have to inevitably lock mutexes, then perhaps doing `_kinBodyEnvironmentIdByReadableInterfaceId` modification might not be as expensive as I'm expecting right now.
But if we have a chance to reduce even one mutex lock, we should go for it. Right now, a lot of the slowness is coming from mutex locks, which we need to figure out how to pass around contexts or critical sections to avoid this.

rschlaikjer:
Implementation updated to act more like a bloom filter. For fun, I ran some benchmarks (full output attached) that involved some bodies that act like mobile robots, payloads, and bodies with transient readables. 

High level results are basically
- If nobody is reading any bodies, the lossy version has a slight edge over the accurate version (1.995us vs 2.128us for add body, similar delta for remove body, identical for update readable, 0.2uS difference for add/remove readable). This is probably just the opt-in-on-use check, which I didn't backport.
- The lossy version is much **worse** if people use `UpdateReadableInterfaces`, because it doesn't short circuit if the set of IDs was the same like the accurate set-diff version does.
- With one reader thread:
  - All readable op timings are about equal except remove readable, which is 6.2uS on the accurate version and 0.4uS on the lossy version. Makes sense.
  - Query times are equal, makes sense, these paths are the same
  - The false-positive rate for the current configuration is around 17% for the bodies with transient readables
- With two reader threads, the `Update` penalty gets much larger
  - We actually see an **inversion** here on mutate. The accurate version is faster by 20% across the board (except for remove readable, which is still zero for the lossy version).
 
Another thing I noticed when checking the benchmark profiles is that the query path is dominated by the shared pointer overhead of actually creating/destroying pointers to the changed bodies. For non-zero false positive rates, this is honestly a real cost - shared pointers are slow, and every false positive churns more of them. I think this is part of why things are slower overall for the lossy version with more readers - each one spends more time holding the lock copying body pointers it doesn't actually need.

[readable_benchmark_lossy.txt](https://github.com/user-attachments/files/29364170/readable_benchmark_lossy.txt)
[readable_benchmark_accurate.txt](https://github.com/user-attachments/files/29364171/readable_benchmark_accurate.txt)

rdiankov:
Thanks for the report.
There was a lot of unnecessary locks and local memory allocation in your implementation, so I cleaned it up.

Unfortunately I could not verify timings with your script since you only shared the report and not the source code of "readable_query_benchmark.B", so please double check.

Thanks,

include/openrave/environment.h
```diff
@@ -602,6 +602,20 @@ class OPENRAVE_API EnvironmentBase : public boost::enable_shared_from_this<Envir
     /// \throw openrave_exception with ORE_Timeout error code
     virtual void GetBodiesMatchingFilter(std::vector<KinBodyPtr>& bodies, const std::function<bool(const KinBody&)>& filterFunction, uint64_t timeout = 0) const = 0;
 
+    /** \brief Get all bodies added to the environment that **might** currently have a non-null readable interface with the given id. <b>[multi-thread safe]</b>
+
```
rschlaikjer: Stray whitespace?

src/libopenrave-core/environment-core.h
```diff
@@ -1680,6 +1681,58 @@ class Environment : public EnvironmentBase
...
+            }
+        }
+
+        // If this is the first time this id has ever been requested, we need to build its cache entry
+        // We need **exclusive** cache access for this rebuild, since we're mutating.
+        std::unique_lock<boost::shared_mutex> cacheLock(_mutexReadableInterfaceCache);
+
+        // Double-check: another thread may have seeded this id during the window between our previous lock
+        std::unordered_set<int>& matchedBodyIndices = _kinBodyEnvironmentIdByReadableInterfaceId[id];
+        matchedBodyIndices.clear(); // clear just in case there are race conditions
```
rschlaikjer: > in case there are race conditions

By race condition, you mean another thread seeding this data? The current comment reads like this is defending against some sort of unlocked race condition, which is potentially confusing.

src/libopenrave-core/environment-core.h
```diff
@@ -1680,6 +1681,58 @@ class Environment : public EnvironmentBase
...
+
+        // Scan all bodies in the env for this readable and add their index to the cache
+        // Since all these bodies are in vecbodies, they _must_ have valid environment body indices.
+        for (const KinBodyPtr& pbody : _vecbodies) {
```
rschlaikjer, line 1731: Only happens on the cold path, so minor, but you could still reserve `vBodiesWithReadableInterface` here

src/libopenrave-core/environment-core.h
```diff
@@ -3506,7 +3559,80 @@ class Environment : public EnvironmentBase
         return true;
     }
 
+    void NotifyKinBodyReadableInterfacesAdded(std::unique_lock<boost::shared_mutex>& lockReadableInterface, const KinBody& body, const std::vector<const char*>& vAddedIds) override
+    {
+        if( vAddedIds.empty() || _kinBodyEnvironmentIdByReadableInterfaceId.empty()) { // empty() should be thread safe
```
rschlaikjer: @rdiankov concurrent `emplace` and `empty` on an unordered set is undefined behaviour. 

Additionally, this introduces a false negative datapath: if `Notify` happens concurrently with a rebuild in `GetBodiesWithReadableInterface`, `Notify` can see an empty set even though the rebuild is currently building it, resulting in a missed notification.

src/libopenrave/kinbody.cpp
```diff
@@ -934,6 +934,56 @@ void KinBody::SetId(const std::string& newid)
...
 
+ReadablePtr KinBody::SetReadableInterface(const std::string& id, const ReadablePtr& readable)
+{
+    // Set the new readable
+    std::unique_lock<boost::shared_mutex> lockReadableInterface(GetReadableInterfaceMutex());
+    ReadablePtr pPrevious = ReadablesContainer::_SetReadableInterface(id, readable);
+
+    if( !pPrevious ) {
+        // there was no previous interface, so notify
+        if (GetEnvironmentBodyIndex() >= 0) {
```
rschlaikjer: This should be `> 0`, 0 is not a valid body index for added bodies. This introduces cache corruption if someone edits readables on a body that isn't added to the env.

src/libopenrave/kinbody.cpp
```diff
@@ -934,6 +934,56 @@ void KinBody::SetId(const std::string& newid)
...
+    }
+    return pPrevious;
+}
+
+void KinBody::SetReadableInterfaces(const READERSMAP& mapReadables, bool bClearAllExisting)
+{
+    std::unique_lock<boost::shared_mutex> lockReadableInterface(GetReadableInterfaceMutex());
+    ReadablesContainer::_SetReadableInterfaces(mapReadables, bClearAllExisting);
+
+    if (GetEnvironmentBodyIndex() >= 0) {
```
rschlaikjer: Should be `> 0`

src/libopenrave/kinbody.cpp
```diff
@@ -934,6 +934,56 @@ void KinBody::SetId(const std::string& newid)
...
+            GetEnv()->NotifyKinBodyReadableInterfacesAdded(lockReadableInterface, *this, vAddedIds);
+        }
+    }
+}
+
+bool KinBody::UpdateReadableInterfaces(const std::map<std::string, ReadablePtr>& newReadableInterfaces)
+{
+    std::unique_lock<boost::shared_mutex> lockReadableInterface(GetReadableInterfaceMutex());
+    std::vector<const char*> vAddedIds;
+    bool bNeedEnvNotify = GetEnvironmentBodyIndex() >= 0; // If we aren't added to the env, do nothing. We don't exist. The env re-checks this, but this avoids the virtual call.
```
rschlaikjer: Should be `> 0`

src/libopenrave-core/environment-core.h
```diff
@@ -3506,7 +3559,80 @@ class Environment : public EnvironmentBase
...
+            // nothing to do
+            return;
+        }
+
+        // Take only the cache lock; the body readable mutex is taken inside the helper (lock order: cache -> body readable).
+        std::unique_lock<boost::shared_mutex> cacheLock(_mutexReadableInterfaceCache);
```
rschlaikjer: @rdiankov this rearranging creates a deadlock by inverting the lock order that was explicitly called out in this comment - you now lock body readable mutex -> _mutexReadableInterfaceCache, but `GetBodiesWithReadableInterface` does the opposite.

src/libopenrave/kinbody.cpp
```diff
@@ -934,6 +934,56 @@ void KinBody::SetId(const std::string& newid)
     }
 }
 
+ReadablePtr KinBody::SetReadableInterface(const std::string& id, const ReadablePtr& readable)
+{
+    // Set the new readable
+    std::unique_lock<boost::shared_mutex> lockReadableInterface(GetReadableInterfaceMutex());
+    ReadablePtr pPrevious = ReadablesContainer::_SetReadableInterface(id, readable);
+
+    if( !pPrevious ) {
```
rschlaikjer: Edge case, but the previous code only notified if `!pPrevious && readable` - this will technically track a body as having a readable even if someone did `SetReadableInterface(id, nullptr)`

rschlaikjer (commented):
@rdiankov these new changes add two major bugs and a handful of minor ones. Please either address the comments or revert to the previous version.

rschlaikjer:
> Unfortunately I could not verify timings with your script since you only shared the report and not the source code of "readable_query_benchmark.B", so please double check.

Feel free to experiment, though obviously you'll want new baseline numbers for your machine

[CMakeLists.txt](https://github.com/user-attachments/files/29446759/CMakeLists.txt)
[readable_query_benchmark.cpp](https://github.com/user-attachments/files/29446760/readable_query_benchmark.cpp)

src/libopenrave-core/environment-core.h
rdiankov: reserve with how many bodies? I'm assuming _vecbodies.size() is much bigger than what vBodiesWithReadableInterface will be

src/libopenrave-core/environment-core.h
rdiankov: As for undefined behavior, `NotifyKinBodyReadableInterfacesAdded` is **not** thread safe including a lot of other KinBody functions. This is just part of how the library works.
If a user is modifying a KinBody, the environment GetMutex() has to be locked, which means we won't get into the thread safety issue you are thinking about.

The only function that is thread-safe is `GetBodiesWithReadableInterface`, and that already does the correct locking with _mutexReadableInterfaceCache


As for `empty` corrupting memory, although not relevant in this context, it does look like it is declared const

https://en.cppreference.com/cpp/container/unordered_set/empty


so how can that corrupt memory? I understand the AI answer is that it corrupts, but I'm not convinced. Might  be worth creating a counter example for future reference.

rdiankov:
Running your tool on the latest version and previous version shows latest one is faster by 30%+

[previousreport.txt](https://github.com/user-attachments/files/29471104/previousreport.txt)

[latestreport2.txt](https://github.com/user-attachments/files/29471357/latestreport2.txt)

src/libopenrave-core/environment-core.h
rdiankov: between letting go of the `shared_lock` and re-locking `unique_lock`, another thread could have come in and updating something

src/libopenrave-core/environment-core.h
rschlaikjer: Fair point, better to not over-reserve.

src/libopenrave/kinbody.cpp
```diff
@@ -934,6 +938,59 @@ void KinBody::SetId(const std::string& newid)
...
+    {
+        std::unique_lock<boost::shared_mutex> lockReadableInterface(GetReadableInterfaceMutex());
+        ReadablesContainer::_SetReadableInterfaces(mapReadables, bClearAllExisting);
+
+        if (GetEnvironmentBodyIndex() > 0) {
+            // This bulk update may have added one or more interfaces.
+            // We do not compute the exact diff here to avoid extra work - let the environment short-circuit if nothing is being tracked.
+            vAddedIds.reserve(GetReadableInterfaces().size());
+            for(const READERSMAP::value_type& info : GetReadableInterfaces()) {
+                vAddedIds.push_back(info.first.c_str());
```
rschlaikjer: I think this introduces a memory safety issue - you're keeping a pointer to the key in the readersmap **past** the lock scope. Another thread could either add and cause a rehash invalidation, or just remove the same ID. Either way, now you have a dangling pointer. This is fine in the specific ID set case (pointer refers to stack frame), but not here.

src/libopenrave/kinbody.cpp
```diff
@@ -934,6 +938,59 @@ void KinBody::SetId(const std::string& newid)
...
+}
+
+bool KinBody::UpdateReadableInterfaces(const std::map<std::string, ReadablePtr>& newReadableInterfaces)
+{
+    std::vector<const char*> vAddedIds;
+    bool bChanged;
+    {
+        std::unique_lock<boost::shared_mutex> lockReadableInterface(GetReadableInterfaceMutex());
+        bool bNeedEnvNotify = GetEnvironmentBodyIndex() > 0; // If we aren't added to the env, do nothing. We don't exist. The env re-checks this, but this avoids the virtual call.
+        bChanged = ReadablesContainer::_UpdateReadableInterfaces(newReadableInterfaces, bNeedEnvNotify ? &vAddedIds : nullptr);
```
rschlaikjer: Same data safety issue here - vAddedIds contains pointers to readable map keys, but you release the readable interface mutex before dereferencing.

src/libopenrave-core/environment-core.h
```diff
@@ -3506,7 +3566,78 @@ class Environment : public EnvironmentBase
...
+    {
+        if( vAddedIds.empty() || _kinBodyEnvironmentIdByReadableInterfaceId.empty() || environmentBodyIndex <= 0 ) { // empty() should be thread safe
+            // nothing to do
+            return;
+        }
+
+        // Take only the cache lock; the body readable mutex is taken inside the helper (lock order: cache -> body readable).
+        std::unique_lock<boost::shared_mutex> cacheLock(_mutexReadableInterfaceCache);
+        for(const char* pAddedId : vAddedIds) {
+            std::unordered_map<std::string, std::unordered_set<int> >::iterator itEnvironmentIds = _kinBodyEnvironmentIdByReadableInterfaceId.find(pAddedId);
```
rschlaikjer, line 3582: This `find` call generates a std::string temporary from your `char*` anyway, so I'm not sure this is actually saving you any allocations - storing std::strings directly would alloc the same amount and avoid the current UAF issues.

src/libopenrave-core/environment-core.h
rschlaikjer: > NotifyKinBodyReadableInterfacesAdded is not thread safe

This seems like it would poison the contract of ReadablesContainer::SetReadableinterface:
```c++
/// \brief Set a new readable interface and return the previously set interface if it exists. 
/// <b>[multi-thread safe]</b>
```

It is difficult to discern from the openrave documentation exactly what is supposed to be allowed with what locks. If the user is supposed to hold the env mutex for **all** readable modifications then yes, a lot of the locking from the original PR can be removed. Is this the case? Should it be called out that readable modification must happen with the env mutex held?

> Empty 

Being declared const means nothing for safety here - if `empty()` walked the internal bucket structure, then calling it during a rehash could easily cause invalid dereferences. Methods declared `const` are also entirely free to mutate any `mutable` contents (e.g, an implementation that lazily caches the actual size when someone calls `size` or `empty`). 

In this exact use case, with the current libstdc++, this is **probably** safe. Empty only checks `_M_element_count`, and I don't see a way for this to falsely read 0. But this is adding explicitly undefined behaviour to the codebase - it strikes me as a dangerous tradeoff to make. This was avoided in the initial PR with the lock-free atomic bool check, which I believe is **free** on x86 since it supports TSO anyway.

src/libopenrave-core/environment-core.h
rschlaikjer: Right, that's what my comment above is about - I'm just saying that the wording of this comment could be improved so the intent is clearer to people that read it later.

rschlaikjer (commented):
@rdiankov I think there's now a different memory safety issue. Whether it's real or not depends on the exact lock guarantees around readables, see the reply on the empty UB thread.

src/libopenrave/kinbody.cpp
rdiankov: great catch, i fixed it by extending the lock. Now that NotifyKinBodyReadableInterfacesAdded doesn't lock the readable mutex, this should be okay.

src/libopenrave/kinbody.cpp
rdiankov: great catch, i fixed it by extending the lock. Now that NotifyKinBodyReadableInterfacesAdded doesn't lock the readable mutex, this should be okay.

rdiankov:
@rschlaikjer  in environment.h, there's a comment on the effect of the mutex, which is that the user is guaranteed Environment will not change if the mutex is locked.

image

So the comment on SetReadableInterface that it is "<b>[multi-thread safe]</b>" is not correct, and I removed it.
It wasn't there for ClearReadableInterfaces, UpdateReadableInterfaces, SetReadableInterfaces.

I also added explicitly that NotifyKinBodyReadableInterfacesAdded is **not** multi-thread safe.
So this should alleviate any concerns about `empty()` being called or informing the user of when they can call these functions.

One of my biggest pet peeves with the current architecture is that the env lock is a recursive lock. Unfortunately this was introduced in the first year of openrave, and 20 years later, have not been able to get rid of it XD
Perhaps a deeper discussion for another day.

src/libopenrave-core/environment-core.h
rdiankov: Really? I thought it would use `compare` which takes a `const char*`..

https://en.cppreference.com/cpp/string/basic_string/compare

given that this is a small optimization, perhaps you can look into this more after this PR is merged.

rschlaikjer: The `find` function on unordered map takes a `key_type`, which is `std::string`, and passing a `char*` to a function that takes `std::string` implicitly converts it.

Godbolt: https://godbolt.org/z/jnEqez4Ge

You can see there's a sneaky `call std::__cxx11::basic_string<...>::_M_create` in there.

To work around this I think you'd need transparent hashing, which is cpp20

rschlaikjer:
@rdiankov the current changes seem good, thanks for the updates. I've pushed some minor comment changes to make clear the locking assumptions around empty() and verified that the new tests still pass. If all looks good to you, feel free to merge / approve and I'll merge.

src/libopenrave-core/environment-core.h
rdiankov: i verified, and you are right, it calls the std::string constructor. just one more thing to optimize down the line.

rdiankov:
@rschlaikjer  great, thanks for all the work and comments. this looks acceptable now~

PR #1561, opened 2026-06-18
[fclrave] Refresh active-link mask when tracked FCLKinBodyInfo is swapped

Puttichai:
# What's wrong

When active-DOF collision checking is on, the FCL backend keeps a cached "active link mask", which is the set of robot links that go into the environment broadphase. It only rebuilds that mask when the robot's active DOFs change, which it detects by comparing a counter (`nActiveDOFUpdateStamp`).

The problem is that each geometry group has its own `FCLKinBodyInfo` with its own counter. A geometry-group switch (`SetBodyGeometryGroup`) makes `GetInfo` return a different info object, and the two counters move in lockstep. So the new info's counter can land on the exact value the cache already had.

When that happens, the guard reads "no active-DOF change" and skips rebuilding the active link mask. The mask is left stale.

# Proposed fix

Also rebuild the mask when the tracked info object itself changed.

# Related link
- issue page: https://tiny.mujin.co.jp/6mr81
- unit test: https://tiny.mujin.co.jp/o38tk


yoshikikanemoto (approved):

kanbouchou (approved):

yoshikikanemoto:
thanks!

PR #1564, opened 2026-07-07
perf(kinbody): skip redundant GetDOFValues recompute in SetDOFValues

Puttichai:
# Description

`KinBody::SetDOFValues` opened with `GetDOFValues(_vTempJoints)`, which reconstructs every joint angle from the current link transforms (a `quatMultiply` chain + `atan2` per DOF — the inverse of the forward kinematics it is about to redo). In the common full-vector path every fetched value is then immediately overwritten by the input, so the recompute is pure waste.

This fills `_vTempJoints` from the input directly, and only calls `GetDOFValues` when an input is `NaN` (the "keep the current value" sentinel). The `dofindices` and NaN paths are unchanged.

- Bit-for-bit identical results (111 / 111 scenes)
- `SetDOFValues` ~36% faster on average (median 38%, up to 53%)

Unit test: https://tiny.mujin.co.jp/7f4na


src/libopenrave/kinbody.cpp
```diff
@@ -2274,19 +2274,40 @@ void KinBody::SetDOFValues(const dReal* pJointValues, int dof, uint32_t checklim
...
             }
         }
     }
     else {
-        for(size_t i = 0; i < _vTempJoints.size(); ++i) {
-            if( !std::isnan(pJointValues[i]) ) {
+        // In the common full-vector path, the inputs are copied verbatim into _vTempJoints, so the
+        // current values fetched by GetDOFValues() would be fully overwritten. GetDOFValues() is expensive
+        // (it recomputes every joint angle from the link transforms - quatMultiply + atan2 per dof), so only
+        // call it when an input is NaN, which is the sentinel meaning "keep the current value for that dof".
```
rdiankov: comment too verbose, not sure what a sentinel is.
please use simple language.
thanks

src/libopenrave/kinbody.cpp
```diff
@@ -2274,19 +2274,40 @@ void KinBody::SetDOFValues(const dReal* pJointValues, int dof, uint32_t checklim
...
+        const int ndof = GetDOF();
```
rdiankov: `GetDOF()` is already called above when setting `expecteddof`

src/libopenrave/kinbody.cpp
```diff
@@ -2274,19 +2274,40 @@ void KinBody::SetDOFValues(const dReal* pJointValues, int dof, uint32_t checklim
...
+        if( bHasNaN ) {
+            GetDOFValues(_vTempJoints);
+            for(int i = 0; i < ndof; ++i) {
+                if( !std::isnan(pJointValues[i]) ) {
+                    _vTempJoints[i] = pJointValues[i];
+                }
+            }
+        }
+        else {
+            for(int i = 0; i < ndof; ++i) {
```
rdiankov: If there is no NAN, was is `_vTempJoints` even initialized then? Just use the original pJointValues

src/libopenrave/kinbody.cpp
Puttichai: Thank you for the comment. I updated this part to be more concise.

src/libopenrave/kinbody.cpp
Puttichai: I switched to using `expecteddof`.

src/libopenrave/kinbody.cpp
Puttichai: I see. I updated this part so we avoid unnecessary setting values of `_vTempJoints` in no-NaN path.

src/libopenrave/kinbody.cpp
```diff
@@ -2271,27 +2271,44 @@ void KinBody::SetDOFValues(const dReal* pJointValues, int dof, uint32_t checklim
...
+        pJointValues = &_vTempJoints[0];
     }
     else {
-        for(size_t i = 0; i < _vTempJoints.size(); ++i) {
-            if( !std::isnan(pJointValues[i]) ) {
-                _vTempJoints[i] = pJointValues[i];
+        // When setting values to all joints, the input pJointValues already holds every value so use it directly and
+        // skip the expensive GetDOFValues. Only when pJointValues contains NaN (meaning "keep the current value") do we
+        // fetch the current values and fill in the rest.
+        _vTempJoints.resize(expecteddof);
```
rdiankov: perhaps only resize if bHasNAN?

src/libopenrave/kinbody.cpp
```diff
@@ -2271,27 +2271,44 @@ void KinBody::SetDOFValues(const dReal* pJointValues, int dof, uint32_t checklim
...
     }
-    int expecteddof = dofindices.size() > 0 ? (int)dofindices.size() : GetDOF();
-    OPENRAVE_ASSERT_OP_FORMAT((int)dof,>=,expecteddof, "env=%s, body '%s' not enough values %d<%d", GetEnv()->GetNameId()%GetName()%dof%GetDOF(),ORE_InvalidArguments);
+    const int expecteddof = dofindices.size() > 0 ? (int)dofindices.size() : GetDOF();
+    OPENRAVE_ASSERT_OP_FORMAT((int)dof, >=, expecteddof,
+                              "env=%s, body '%s' not enough values %d<%d", GetEnv()->GetNameId() % GetName() % dof % expecteddof,
+                              ORE_InvalidArguments);
 
-    GetDOFValues(_vTempJoints);
     if( dofindices.size() > 0 ) {
```
rdiankov, line 2279: @Puttichai I wonder if there could be another fast path where dofindices.size() == expecteddof, and dofinidces contains all the indices of the robot (even if they are in a different order).

src/libopenrave/kinbody.cpp
Puttichai: I thought about it too but `_vTempJoints` may also be used under `if( checklimits != CLA_Nothing )` below. I added a comment here to clarify.

src/libopenrave/kinbody.cpp
Puttichai: I see. Thank you for the suggestion. I made a change to skip `GetDOFValues` when `dofindices` already covers all dofs. Also updated test cases.

src/libopenrave/kinbody.cpp
```diff
@@ -2271,27 +2271,71 @@ void KinBody::SetDOFValues(const dReal* pJointValues, int dof, uint32_t checklim
...
-        // user only set a certain number of indices, so have to fill the temporary array with the full set of values first
-        // and then overwrite with the user set values
-        for(size_t i = 0; i < dofindices.size(); ++i) {
-            if( !std::isnan(pJointValues[i]) ) {
+        // When dofindices covers every dof of the body and the input has no NaN, every element of _vTempJoints gets
+        // overwritten, so can skip the expensive GetDOFValues. To check the coverage, pre-fill _vTempJoints with NaN
+        // before writing the input values into it. If no NaN remains, all dofs were covered.
+        bool bNeedCurrentValues = ((int)dofindices.size() != GetDOF());
+        if( !bNeedCurrentValues ) {
+            for(int i = 0; i < expecteddof; ++i) {
```
rdiankov, line 2285: @Puttichai we should also becareful that dofindices doesn't have repeated indices. at the very least we have to assert if you don't think this is a valid input

Puttichai: @rdiankov With repeated indices, `_vTempJoints` will have some values being `NaN` (around [here](https://github.com/rdiankov/openrave/pull/1564/changes?diff=unified#diff-b56c9a013ed0ae0f9583bff93f4aac11a43dd3f7c623352dd055a2575e0f6058R2299)). Then we'll set `bNeedCurrentValues = true` and fall back to the old path that does `GetDOFValues` and then overwrite. The dof whose index is repated get the last value given for it.

In terms of the behavior in this case, it is the same as it was originally, where repeated indices were accepted without assertion.

PR #1565, opened 2026-07-10
Make GetRobots O(robots) not O(bodies)

rschlaikjer:
Maintain a separate tracking list of bodies so that GetRobots doesn't have to linear scan the entire env.
Bodies cannot transition between types while added, and robot addition/removal is rare, so this has minimal costs. 

The bottleneck that prompted this is described here: https://tiny.mujin.co.jp/xi72y


src/libopenrave-core/environment-core.h
```diff
@@ -4204,6 +4213,24 @@ class Environment : public EnvironmentBase
...
 
+        // Whether a body is a robot cannot change while it is in the environment, so track robots in a dedicated cache.
+        // This allows GetRobots to be O(robots) instead of O(bodies), which is important for larger scenes.
+        if (pbody->IsRobot()) {
+            RobotBasePtr probot = RaveInterfaceCast<RobotBase>(pbody);
+            if (!!probot) {
+                // Insert preserving ascending env body index order; recycled indices can land anywhere in the vector
+                const std::vector<RobotBasePtr>::iterator itInsert = std::lower_bound(
+                    _vecrobots.begin(), _vecrobots.end(), envBodyIndex,
+                    [](const RobotBasePtr& pCachedRobot, int index) {
```
rdiankov: can you avoid lambda functions? thanks

src/libopenrave-core/environment-core.h
```diff
@@ -1761,13 +1763,7 @@ class Environment : public EnvironmentBase
...
             throw OPENRAVE_EXCEPTION_FORMAT(_("timeout of %f s failed"), (1e-6 * static_cast<double>(timeout)), ORE_Timeout);
         }
-        robots.clear();
-        for (const KinBodyPtr& pbody : _vecbodies) {
-            if (!pbody || !pbody->IsRobot()) {
-                continue;
-            }
-            robots.push_back(RaveInterfaceCast<RobotBase>(pbody));
-        }
+        robots.assign(_vecrobots.begin(), _vecrobots.end());
```
rdiankov: just curious, why not use `operator=`?

rdiankov:
@rschlaikjer For another optimization, what do you think about having `string_map` be an `unordered_map` instead? Or have you done experiments and determined `map` is faster?

```
using string_map = std::map<std::string, Value, std::less<> >;
```

src/libopenrave-core/environment-core.h
rdiankov: it doesn't looks like the property that `_vecrobots` has escending environment body indices is used anywhere in the code, except here during adding. Perhaps best to just `push_back` for now.

rschlaikjer: Is there a reason for avoiding lambda functions? As compiler primitives they're more optimizable than bound / free functions when not downcast (and equivalent when they are), have clearer capture semantics than `std::bind` + `std::ref`, and keep the implementation of helper functions local to where they're used, which makes the code more readable.

rschlaikjer: > ordering

I don't think it's a published guarantee, but it seemed like the sort of thing consumers might accidentally be depending on... I can remove it and see if anything breaks.

src/libopenrave-core/environment-core.h
rschlaikjer: No real reason - `assign` would keep working if `_vecrobots` got converted to a non-vector container type, but they're equivalent here.

rschlaikjer:
> @rschlaikjer For another optimization, what do you think about having `string_map` be an `unordered_map` instead? Or have you done experiments and determined `map` is faster?
> 
> ```
> using string_map = std::map<std::string, Value, std::less<> >;
> ```

The last time I looked into it I'm pretty sure my takeaway was that `unordered_map` is strictly better than `map` (it has slightly higher memory usage due to the bucket allocation, but that's about it), so unless ordering is important then yeah, changing those to `unordered_map` would probably improve performance.

rdiankov:
@rschlaikjer final changes look good! once our internal pipeline is validated, feel free to merge. thanks

rschlaikjer:
Ran a few pipelines and it seems that everything is a flake except for one internal test that relies on `GetRobots()[0]` being a specific robot. Will patch that then merge.

PR #1569, opened 2026-07-28
Fix builds under Boost 1.90.0 (Ubuntu 26.04)

PeterBowman:
- Boost.System no longer exists as a standalone CMake package since 1.89 (https://www.boost.org/releases/1.89.0/):
  > - The stub compiled library has been removed; System has been header-only since release 1.69.
  > - This may affect `CMakeLists.txt` files containing `find_package(Boost COMPONENTS system ...)`. The easiest fix is to just remove `system` from the list of components as it's no longer required. If compatibility with Boost releases earlier than 1.69 is to be preserved, one can use `find_package(Boost COMPONENTS ... OPTIONAL_COMPONENTS system)`. 
- `boost::filesystem::is_complete` was replaced by `is_absolute` at least since 1.69, i.e. 2018 (https://www.boost.org/doc/libs/1_90_0/libs/filesystem/doc/v3.html)
- `boost::filesystem::path::normalize` was long deprecated and ceased to exist; the replacement code was already there, I have just activated the enclosing #if-branch (https://stackoverflow.com/a/1750710)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,7 +265,7 @@ else()
   set(OPENRAVE_LIBRARY_SUFFIX "${OPENRAVE_SOVERSION}" CACHE STRING "Suffix to append to library names")
 endif()
 
-find_package(Boost REQUIRED COMPONENTS filesystem system thread iostreams)
+find_package(Boost REQUIRED COMPONENTS filesystem thread iostreams)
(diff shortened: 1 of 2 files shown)


PeterBowman:
Hello, sorry for the bump. Would it be feasible to have this merged into master?

PR #1570, opened 2026-08-03
Allow skipping _UpdateGrabbedBodies during serialization

rschlaikjer:
Add a new `ExtractInfoOptions` flag that skips updating the grabbed bodies as part of extract info. 

When e.g. serializing dynamic body info, bodies are transformed to the origin first to get link transforms in the body's own reference frame. However, for bodies with grabbed objects, this pays the price of also moving _every grabbed body_ to the origin and back, which is extremely costly for bodies that grab hundreds or potentially thousands of objects.

If `EIO_SkipGrabbedBodyUpdates` is set, completely skip this move/unmove path for the grabbed bodies. This has no effect for dynamic state serialization, where the grabbed bodies' transforms are never read.


rschlaikjer:
Tests + passing flag https://tiny.mujin.co.jp/nnoip

yoshikikanemoto:
@rschlaikjer I think we don't need the new option and we can always skip updating grabbed bodies.

rschlaikjer:
> @rschlaikjer I think we don't need the new option and we can always skip updating grabbed bodies.

I did wonder if we could just always do this, but I think there is one case where it _would_ be a behaviour change: if there is a `Prop_LinkTransforms` callback, then it would observe the body at the origin but its grabbed bodies at their original location. If we don't care about this use case I can remove the flag.

yoshikikanemoto:
> > @rschlaikjer I think we don't need the new option and we can always skip updating grabbed bodies.
> 
> I did wonder if we could just always do this, but I think there is one case where it _would_ be a behaviour change: if there is a `Prop_LinkTransforms` callback, then it would observe the body at the origin but its grabbed bodies at their original location. If we don't care about this use case I can remove the flag.

I think the callback should not be called from `ExtractInfo` in the first place since transform and DOF values are going to be reverted anyways. what about adding a RAII object to temporarily remove the callback from kinbdoy?

rschlaikjer:
> what about adding a RAII object to temporarily remove the callback from kinbdoy?

Updated

yoshikikanemoto (approved):

rdiankov:
Guys, ExtractInfo was never meant to be called in a critical loop, that's why its implementation is not that fast and lazily uses savers and transforming the links to the origin.
If you really care about the performance of the system, I recommend you figure out an implementation that does not need to call `SetTransform` and `SetDOFValues`

rschlaikjer:
> If you really care about the performance of the system, I recommend you figure out an implementation that does not need to call SetTransform and SetDOFValues

Happy to look into it, but do you mind if we merge this as a temporary measure to at least recover CPU today?

The reason this is in the hot path is that it's how we serialize the dynamic state - I think I remember asking before about why we extract -> serialize and was told it was an intentional design decision. Allowing _direct_ serialization without temporaries seems like the biggest performance win - do we want to reevaluate the decision not to do that, or just focus on optimizing extraction?

rschlaikjer:
Merging this as a step in the right direction.

PR #1573, opened 2026-09-14
Prevent stale buffer memory leak

rschlaikjer:
To avoid memory allocation churn, the published bodies BodyState objects retain any previously allocated capacity on Reset.

However, the ordering of environment body to BodyState is not fixed: as bodies are added and removed, which body maps to which index can change. If some bodies are orders of magnitude larger on some axes (e.g, storage systems that grab hundreds or thousands of cases), every slot they land in gets a permanently enlarged buffer. Instead of being an optimization, this instead _de_optimizes by bloating the resident set and swapping memory out to disk. 

Pin the mapping of body state to actual body, and reset the allocated capacity when the tracked body changes so that we do not accumulate wasted capacity.


rschlaikjer:
Pipeline - https://tiny.mujin.co.jp/u0xqz

yoshikikanemoto:
LGTM. can you bump patch version and add changelog?
@jschaare can you also review?

yoshikikanemoto (approved):

jschaare (approved):
Also looks good to me

